PR #3, opened 2025-12-07
Update README.md

effreek:
Updating the Electronics readme.


effreek:
In addition to the updated readme the folder structure needs created so we can start to drop the KiCad stuff into GitHub. @spencerhhubert

spencerhhubert (approved):

Issue #92, opened 2026-03-25
Ribbon vs JST connections


effreek:
Keep both options on the main board but de-pop the JST versions on the main board. This will reflect the move to the ribbon cable for the interface layer.

PR #162, opened 2026-05-31
hive: fix parts catalog sync re-fetching the same page forever

spencerhhubert:
## Problem

The prod Hive (`hive.basically.website`) runs off `sorthive`, and its Rebrickable **parts catalog sync has never completed** — it's wedged in an infinite re-fetch loop. Live status when diagnosed:

- `sync_type: parts`, `running: true`, stuck at **`2000 / 62549 parts (3%)`**
- **`pages_fetched: 8626`** (should be ~63 pages at `page_size=1000`)
- `last_synced_at.parts: null` — parts has never once finished
- `error: null` — not erroring, just looping; backend log counter was at `[rebrickable] request #92,000+`
- Running ~2h, continuously hammering the Rebrickable API

## Root cause

`_syncPartsPage` derives the page number from `len(parts_data.parts)` — the **in-memory** cache. That cache is only refreshed by `reloadPartsData()` *after the whole sync finishes*, so during the loop it never grows. `page_num` stays pinned, the same page is fetched forever, the DB row count never advances past it, `done` is never reached, and Rebrickable eventually 429s.

## Fix

Derive `page_num` from the **persisted DB row count** (`SELECT COUNT(*) FROM parts`), which grows as each page is upserted, so pagination advances to completion. This unwedges the current stuck state on next run (the DB-count page calc skips past the already-cached pages) without any manual DB surgery.

This mirrors the fix already on `main` (#148), but as a minimal two-line change rather than pulling in the rest of that catalog-sync refactor — intended to get the running `sorthive` deployment back on its feet.

## Notes

- Minimal, surgical change; keeps `sorthive`'s existing `requests.get`/`throttle_fn` structure.
- Not yet deployed to prod — no containers were touched. Deploy is a separate step (rebuild `hive-backend` image off this branch + restart).
- Do not merge until reviewed.

### DIFF
--- a/software/hive/README.md
+++ b/software/hive/README.md
@@ -76,6 +76,37 @@ The production stack expects:
 
 On startup, the backend automatically runs Alembic migrations and bootstraps an admin user from `ADMIN_EMAIL` / `ADMIN_PASSWORD` if those values are set.
 
+### Persisted data
+
+The backend bind-mounts two host directories under `software/hive/data/` so they
+survive container rebuilds/recreates:
+
+- `data/uploads` — uploaded sample images
+- `data/profile_builder` — the Rebrickable parts catalog SQLite db
+  (`parts.db`: parts, categories, colors). **Without this mount the catalog
+  lives only in the container layer and is wiped on every recreate, forcing a
+  full multi-hour re-sync from Rebrickable on each deploy.**
+
+`REBRICKABLE_API_KEY` must be set in `.env.prod` for the catalog auto-sync to run.
+
+To seed `parts.db` (e.g. from a machine that already has a populated catalog)
+instead of syncing from scratch, copy a clean snapshot into the host dir before
+starting the stack — the container runs as uid `100`, so the dir and file must
+be writable by it:
+
+```bash
+# on the source machine: make a consistent copy (works even while the db is open)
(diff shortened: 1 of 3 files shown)


spencerhhubert:
Added a second commit (`2dcb2961`) and **deployed both to prod** (`hive.basically.website`, which runs `sorthive`):

**Commit 2 — persist the parts catalog db:** `parts.db` lived only in the container layer (only `data/uploads` was mounted), so every recreate wiped it and forced a full re-sync — which, combined with the pagination bug, is why it never finished. Now bind-mounts `./data/profile_builder:/app/data/profile_builder` and passes `REBRICKABLE_API_KEY` through to the backend (was a prod-local uncommitted compose edit; now committed). README documents the volume + how to seed `parts.db`.

**Deployed + verified on prod:**
- Backend image rebuilt with the pagination fix; backend-only recreate (frontend/openclaw/traefik/postgres untouched).
- Seeded `parts.db` from a populated local catalog (~62.5k parts, `schema_version=7`) via `sqlite3 .backup` → scp → `chown 100:101` (container runs as uid 100). No parts-db migrations ran (sorthive maxes at migration 006 ≤ 7).
- Catalog now reads complete (`last_sync.parts` set); rebrickable request rate is **0/15s** (no more loop). API `/api/health` → 200, container healthy. WAL sidecars appear on the host mount → persistence confirmed.

Still **not merged** — open for review.

PR #175, opened 2026-06-05
June 2 updates: connectivity diagnostics, CORS escape hatch, chute no-homing, local capture storage cap

spencerhhubert:
Bundles the work on `spencer/june-2-updates-01` that's ahead of `main`.

## Connectivity / server diagnostics
- **Log CORS/WebSocket origin rejections** so blocked origins are diagnosable instead of silently 403ing.
- **Log bind host + effective allowlist at startup** for quick sanity-checking what the server will accept.
- **Request/probe/websocket diagnostics** endpoints + frontend wiring (`AppHeader`, `BackendConnectionGuard`, `ws.ts`, `backend.ts`) to surface connection state.
- **`SORTER_API_ALLOW_ANY_ORIGIN` escape hatch** for permissive-origin situations during bring-up.
- **Allow IP-only (no port) requester** origins.

## Hardware
- **Operate the chute after a no-homing init** so the chute is usable without a homing pass.

## Local capture storage cap
The standalone sample collector wrote every capture with no retention bound, which filled the Pi's disk to 100% and crash-looped the backend (SQLite `disk I/O error` on WAL init).

- Configurable **local-storage cap (default 1 GiB)** enforced at the single write chokepoint (`_archiveSample`): once usage exceeds the cap, **delete oldest sample files first** down to a 90% low-water mark (hysteresis avoids walking the tree on every save). Fully-drained session dirs are removed; `manifest.json` preserved. Cap persists in the classification-training config.
- `/api/system/sample-capture` now reports `storage_cap_mb` / `storage_used_mb` and accepts `storage_cap_mb` (setting it enforces immediately).
- New **"Local storage cap"** GB input in the existing capture menu (`SampleCaptureSection.svelte`), showing current usage.

Verified live on `sorter-dark-brown-axle-0ffbef`: setting the 1 GB cap trimmed the training dir 3.5 GB → 932 MB (90% low-water) and the backend recovered.

## StallGuard stall detection + tuning
A push to make TMC2209 StallGuard actually usable on the chute — both the tuning workflow and runtime enforcement. (StallGuard4 only reads a valid load at smooth constant cruise, in StealthChop; this work makes the tooling honest about that and the enforcement robust.)

- **Chute stress test halts into a terminal `stalled` state** on a confirmed stall instead of rolling into the next leg as if the move completed. Only the firmware's sticky per-channel DIAG latch counts (`_confirmStall()` requires the latch set on several consecutive clean reads), so a garbled UART ack on the marginal chute bus can't fake a stall.
- **Speed-matched, data-driven threshold suggestion.** The StallGuard page now filters the unloaded sweep, loaded stall test, and stress-run baseline to the *same speed* as the run being viewed (no more pairing a 2000 µs/s floor with an old 3000 stall test). It **derives TCOOLTHRS** from the measured cruise TSTEP (fastest-sustained ×1.75) instead of echoing a typed value, and computes SGTHRS as the geometric midpoint of the floor/dip gap. Save writes both to `machine.toml`; the button shows both values.
- **Reliability warning.** Cross-checks the proposed trigger against real (stress-test) motion at that speed; if normal reversing motion dips into the trip range (the cruise SG baseline isn't separable from a stall at that speed), it returns `reliable: false` and the page shows a hard "not reliably tunable at &lt;speed&gt;" box instead of a confident Save. Also flags loaded runs that never actually stalled.
- **Rebuilt load chart** (`StallGuardChart`): stacked SG (0–512) / TSTEP (log, with the tcoolthrs gate line) / CS panels sharing the time axis, replacing the old min-max-normalized single-axis cram; green gate-open shading across panels, a gate readout (open time / accel ramp), and a hover crosshair + value box.
- **Per-stepper stall state + derived incident.** The monitor mirrors the firmware DIAG latch onto `stepper.stalled` each poll as the single source of truth; the blocking `stepper_stall` incident is now a pure projection of it (raised while any motor is latched, auto-resolves when none) — no separate ack bookkeeping that could leave a stale latch stuck. A fresh stall invalidates the chute home reference (lost steps → re-home required).
- **Clear from anywhere.** Each station page shows a red "motor is stalled / Clear stall" banner at the top, polled live via new `GET /steppers/stall-state`; `POST /stepper/{name}/clear-stall` clears one latch and the dashboard incident resolves indirectly; the global Acknowledge (`/stall-incident/clear`) now resets all armed latches, not just the incident flag. `stalled` is also exposed in `/api/stepper/{name}/tmc`.


spencerhhubert:
Superseded by #180 — `spencer/two-piece-tracker-01` is descended from this branch (branch point 353a01f4) and includes all of this work plus the two-piece classification channel and the servo / bin-layouts work since. Rolled into #180.

PR #232, opened 2026-08-01
Supervisor: self-heal backend crash loops, show crash output in the UI

spencerhhubert:
Prompted by the bbs-01 outage on 2026-08-01, where a power loss corrupted `perception/__pycache__/capture.cpython-312.pyc` and the backend crash-looped for ~18h — invisible to systemd (the service wraps the supervisor, which stayed "active") and opaque in the UI (generic "Backend Unavailable").

**Crash-loop self-heal.** The supervisor counts consecutive fast exits (process dies <30s after launch; window configurable via `BACKEND_SUPERVISOR_FAST_CRASH_WINDOW_S`). After 3 in a row it clears the backend tree's `__pycache__` dirs (skipping `.venv`) before the next restart, once per streak. A corrupt `.pyc` passes Python's cache-header check (magic + source mtime + size) while its body is garbage, so a crash loop caused by one never recovers on its own. A healthy run or a slow crash resets the streak.

**Crash visibility.** `main.py`'s stderr is teed — lines still reach journald, and the last 60 are kept in a ring buffer. On an unexpected exit the tail is exposed on `/api/supervisor/status` as `last_crash_output`, alongside `crash_looping`, `consecutive_fast_crashes`, and `pycache_cleared_at`. The Backend Unavailable dialog shows the actual traceback instead of "check that the machine is powered on".

**Systemd units.** Same failure mode one layer down: the backend and UI units (dev + prod, plus the sorteros overlay copies) now survive power-cut corruption of their generated caches rather than failing to start.

**Hive fleet analytics.** Also carries a fix for `/admin/machines` — three machines named "Lego Sorter" made the by-machine donut throw `each_key_duplicate` and blank the whole analytics section. Charts key on ids now, and colliding machine names get a short id suffix. Already deployed to hive prod.

Stacked on #230.

### DIFF
--- a/software/hive/frontend/src/lib/components/charts/AnalyticsDashboard.svelte
+++ b/software/hive/frontend/src/lib/components/charts/AnalyticsDashboard.svelte
@@ -83,16 +83,34 @@
 	}
 
 	const statusSegments = $derived<DonutSegment[]>(
-		(data?.distributions.by_status ?? []).map((s) => ({ label: s.label, value: s.value, color: statusColor(s.label) }))
+		(data?.distributions.by_status ?? []).map((s) => ({ key: s.label, label: s.label, value: s.value, color: statusColor(s.label) }))
 	);
-	const machineSegments = $derived<DonutSegment[]>(
-		(data?.distributions.by_machine ?? []).map((m, i) => ({ label: m.label, value: m.value, color: PALETTE[i % PALETTE.length] }))
-	);
+	const machineSegments = $derived.by<DonutSegment[]>(() => {
+		const rows = data?.distributions.by_machine ?? [];
+		// Machine names aren't unique across the fleet — tag the collisions so the
+		// legend doesn't show three identical rows.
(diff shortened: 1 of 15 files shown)


spencerhhubert:
Second commit extends this to the full power-cut-corruption class after bbs-01 broke again on 2026-08-03 (corrupt `.svelte-kit/tsconfig.json` → vite died instantly 3× → systemd start limit gave up → UI dead on 5173):

- All four systemd units: `StartLimitIntervalSec=0` — never permanently give up on a headless field machine.
- `sorter-ui-dev`: `ExecStartPre` wipes `.svelte-kit` + `node_modules/.vite` every start (vite rewrites them on startup; they're exactly what a power cut catches mid-write). Prod UI untouched — it serves the built app from `.svelte-kit/output`.
- Backend units: `PYTHONPYCACHEPREFIX=/tmp/sorter-pycache` (tmpfs on the Pis) — bytecode caches can no longer survive a power cut at all; the supervisor self-heal also clears the prefix dir.
- sorteros overlay unit copies kept in sync.

Not deployed to any machine.

PR #273, opened 2026-08-09
harness: adopt Jon's drawings as the source of truth

basically-balloon:
Jon (`@effreek`) posted three WireViz projects in the #general call chat during the 8 Aug electronics sync: `Stepper_Harness.zip`, `DC_PSU_Harness.zip` and `DC_basicallyV1_3_Harness.zip`. Spencer said to treat his drawings as the source of truth rather than adopt them piecemeal, so this replaces the guessed values in `electronics/wire_harness/` with his wherever the two disagree.

Replaces `sorter-wiring` [PR #2](https://github.com/basicallysource/sorter-wiring/pull/2), which was opened against the repo the harness sources moved out of a few hours later, and is now closed.

A follow-up PR on top of this one carries what the call settled that never made it into a zip.

## Steppers, cable S

```
             before                  after (Jon)
motor end    XH2.54 4-pin clone      JST PHR-6, the motor's socket
             (StepperOnline lead)    positions 2 and 5 unpopulated
mapping      1,2,3,4 -> 1,3,2,4      1,2,3,4 -> 1,4,3,6
length       40 in                   1 m  (same length, standard unit)
colours      BK RD GN BU             BU GN RD BK
contacts     not specified           SPH-002T-P0.5S, per populated pin
sleeving     none                    PVC over the bundle
```

The StepperOnline shipped lead drops out of the spec entirely: the cable plugs into the motor's own 6-position socket. `CH` (chute) is untouched, his drawing does not cover it.

## Power

- **Board 24V input is JST VH**, `VHR-2` + `SVH-21T-P1.1`, and `W1` moves 22 AWG to 18 AWG. `power.yml` had this marked `GUESS - verify XH vs VH` and `rfq.txt` was still asking a vendor for `XHP-2`, so as it stood the RFQ would have gone out for the wrong connector.
- **PSU end gets a real terminal spec**: 18 AWG spade/fork (fork/split ring), M3.5 stud, 8 mm wide max, sized to fit the LRS-350-24 screw terminals. The 8 mm max is the load-bearing part, plenty of terminals are wider and will not fit between the screws.
- The pigtails come already attached to the panel-mount jacks, so they are a buying decision rather than something to have made.

## Left as an open item, not silently changed

Jon's `DC_PSU_Harness` is drawn as **x3 per PSU build**; `power.yml` has six jacks feeding five loads plus a spare. That is a real disagreement about the machine, not a drawing detail, so it is flagged on the harness page and in the order spec's guess list. (The follow-up PR has the reason behind it.)

Also still open: which pin is +24V (his drawing does not say), and the coil order itself, which is from his drawing rather than a measurement. His own note says to check coils with a multimeter first, and the docs now say so too.

## Docs

Pin-swap diagram redrawn as a 4-to-6 position map, wire schedule, order spec, connector BOM and the guess list all follow the drawings. Separately, per Spencer: the WireViz page is cut back to just the zip and the drawings, and the setup photos page is deleted along with its nav entry.


basically-balloon:

spencerhhubert (approved):

PR #280, opened 2026-08-09
harness: say which guesses Jon's drawings did not cover

basically-balloon:
Spencer asked why values still say `GUESS` when they came from Jon's drawings.

**Audit result: nothing Jon specified is still marked `GUESS`.** The instances visible on the live WireViz page are a stale CDN object, the 8 Aug render, which is a separate problem (the `?v=` cache-buster is inert because the zone ignores query strings in the cache key).

What is still marked `GUESS`, and why each one is legitimately still a guess:

```
value                     covered by Jon?
LED feed gauge L1-L3      no  - he sent nothing for LEDs
LED pigtail gauge L1P-L3P no  - same
LIM gauge                 no  - same
CH gauge and length       no  - his sheet is the channel cable only
W2 hub gauge              no  - his DC files are the pigtail and board feed
W3 buck gauge             no  - same
LED / limit dupont pin order   no
which of COM/NC/NO on the switch   no
```

So rather than change any value, each of those now says on the drawing that Jon's drawings do not cover it. Otherwise a reader sees `S` with a hard 24 AWG and `CH` right above it marked `GUESS` on the same sheet and has no way to know why.

One thing did become stale and is removed: the `verify hub jack size` flag on `W2`. That was settled in the call off the Waveshare wiki (`DC-044` is 5.5 × 2.1), and the pin-map section two lines below already says so.


basically-balloon:
Superseded by #281. Both PRs ended up editing the same note lines in `power.yml`, `steppers.yml` and `leds.yml`, so rather than leave them to conflict, everything here is folded into #281: the "not covered by Jon's drawings" labelling on each remaining GUESS, and the settled `verify hub jack size` flag on W2.

PR #281, opened 2026-08-09
harness: split the PSU drawings out as sub-harnesses under the overview

basically-balloon:
Jon's suggestion, relayed by Spencer: keep the big overview drawing, and put the individual harnesses it is made of below it. The 24V sheet is a whole-system diagram, but what a cable vendor actually quotes is one assembly.

So `power.yml` stays the overview, and the two assemblies come straight from Jon's own files:

- **`psu-pigtail`** (his `DC_PSU_Harness`) — spade terminals onto the LRS-350-24 screw block, panel-mount jack the other end, x3 per PSU build.
- **`board-power`** (his `DC_basicallyV1_3_Harness`) — barrel plug to `VHR-2`, the board's own feed, `W1` on the overview.

## How the page knows

Entries in `docs/src/liquid/_data/harness.yml` can now carry `of: <parent>`. The WireViz page renders those one heading level down, so:

```
## 24V power distribution        <- overview
###   PSU output pigtail         <- buildable
###   24V to basically board     <- buildable
## Stepper cables
## LED drops and limit switch
```

Both are in the `drawings` list in `build-harness.sh`, so the supplier zip carries them and their BOMs too.

## The catch, written into AGENTS.md

Sub-harnesses restate values that also appear on the overview: `W1` is 18 AWG, 36 in in two places now. That is a real drift risk and the only defence is remembering to change both, so the "adding a drawing" section says so.

Steppers and LEDs are left flat. Jon's point was about the PSU, and neither of those has an obvious overview/assembly split yet.

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -176,6 +176,14 @@ harness URL must carry). So a branch preview shows the branch's drawings,
 production shows main's, and there is nothing to update in `docs/` when a
 drawing's content changes.
 
+**`resolveHarness()` reads the host's branch env var, and that is not
+optional.** Both Pages and Vercel build from a detached HEAD, so the git
+fallback returns `HEAD` and everything collapses to `main`: previews then show
+main's drawings and any drawing added on the branch 404s. It reads
+`CF_PAGES_BRANCH` first, `VERCEL_GIT_COMMIT_REF` second, git last. If the site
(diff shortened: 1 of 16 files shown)


basically-balloon:

spencerhhubert (approved):

PR #282, opened 2026-08-09
Show each harness BOM inline on the WireViz page

basically-balloon:
Every drawing on `/hardware/electronics/wireviz/` now shows its bill of materials as a table, directly under the image and the download line. Same content the generated `.html` carries, without leaving the page.

### The choice: fetched in the browser, not baked in at build time

The `.bom.tsv` files are build artifacts in the assets bucket, never in git, so the page has to fetch them from somewhere. I looked at both options and picked client side:

- **The harness workflow only fires on changes under `electronics/wire_harness/`.** A docs-only branch (like this one) never renders, so there is no `harness/<ref>/` prefix in the bucket for it at all. A build-time fetch would 404 on all five drawings for most previews, not occasionally.
- **On a harness branch it does run, but on `pull_request`, while Vercel builds on push.** The render lands ~90 seconds later. The docs build reliably wins that race, so a prerendered table would bake in whatever was in the bucket beforehand and then sit there stale. That breaks the property the whole pipeline is designed around: push the YAML, the preview self-corrects. The drawings above it already work that way because the browser fetches them.
- It also keeps the docs build from depending on the bucket being reachable.

**Verified CORS before committing to this**, since it is the thing that could have killed the approach: `scripts/img-worker/worker.js` sets `access-control-allow-origin: *` on every response unconditionally, and it is present on a cache MISS, on a HIT, and with a query string.

### Caveats

- **The BOM needs JavaScript.** That is the same deal the drawings take (they are remote images), and the `BOM (TSV)` download link stays in the download line as the fallback.
- **If the render has not published for the ref yet**, the page says `Not published for this build yet.` with a link to the TSV, rather than an empty table or a broken layout. Reload once the render lands and the table fills in.
- The `?v=` cache-buster is carried for consistency with every other harness URL, but it does nothing (Cloudflare ignores the query string, and the worker strips it going upstream). A re-rendered BOM can sit behind the CDN cache exactly like a re-rendered PNG can.

### Parsing

Designed against the five real TSVs on `harness-sub-drawings`, not guessed. Handles the optional `Manufacturer`/`MPN` columns (`power`, `board-power` and `steppers` have them, `psu-pigtail` and `leds` do not), empty cells, CRLF, and rows short a trailing column (the width is the widest row, every cell padded out to it). Cells are set with `textContent`, so nothing coming out of the bucket can inject markup.

Mobile: the site's tables already scroll inside themselves rather than dragging the page sideways, and `.bom-table` drops the default 140px minimum cell width so a 7-column BOM compresses instead of forcing a scroll at all.

### Checks

- `npm run build` passes, all five `.bom` containers are in the prerendered HTML.
- `validate_frontmatter.py` reports the same 5 pre-existing errors as the base branch, none new.
- Table construction exercised against all five published TSVs plus ragged/CRLF/injection cases.

**Files:** `docs/src/lib/bom.ts` (new, the parse and table build), the effect in `docs/src/routes/[...path]/+page.svelte`, the `.bom` markup in `wireviz.md`, styles in `layout.css`, and a note in `docs/AGENTS.md`.

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -163,6 +163,17 @@ harness URL must carry). So a branch preview shows the branch's drawings,
 production shows main's, and there is nothing to update in `docs/` when a
 drawing's content changes.
 
+Each drawing's bill of materials is shown inline on the WireViz page, fetched
+**in the browser** from `<harness_base>/<name>.bom.tsv` and turned into a table
+by `src/lib/bom.ts` (driven by an effect in `src/routes/[...path]/+page.svelte`,
+markup is the `.bom` div in `wireviz.md`). It is not baked in at build time on
+purpose: the harness workflow only runs on harness path changes and finishes
+well after Vercel starts, so a prerendered BOM would be missing or stale on most
+previews and could never correct itself. `img.basically.website` sends
+`access-control-allow-origin: *` on everything (`scripts/img-worker/worker.js`),
+which is what makes the fetch legal. When there is no render for the ref, the
+page says so and links the TSV.
+
 `src/liquid/_data/harness.yml` is just the display list (name, title, caption
 per drawing). Adding a drawing: new YAML source, an entry there, and an entry
 in the `drawings` list inside `build-harness.sh`. Full pipeline doc:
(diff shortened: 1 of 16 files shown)


basically-balloon:

basically-balloon:
This has the approvals to merge, but the branch does not merge cleanly into `harness-sub-drawings` right now. Rebase it and the board will pick it up again.

basically-balloon:
Closing as superseded by #281. This PR and #281 both add the same inline-BOM feature to the WireViz page, both client-side for the same reason (the TSVs are bucket artifacts, and a build-time fetch would race the render job). #281 already carries it, verified on the preview, and is the PR that takes `harness-sub-drawings` into main. This branch is built on the pre-rebase `harness-sub-drawings`, which is why it now conflicts. Nothing here is lost: the feature ships in #281.

For the record: this was duplicated work on my side. The sub-agent that produced this branch did run, and I later re-implemented the same thing directly in #281 without realising this existed.

PR #332, opened 2026-08-20
docs: roll the socket/button "A" tag out to the rest of the screws

basically-balloon:
Rolls the socket/button **"A"** tag and naming out to the rest of the interchangeable screws, matching what went in for M5×16 (#319/#321).

- **Consolidate** the M5×12 and M5×20 socket/button card pairs into one card each: correct "socket / button head" name, green "A" tag, the separate button card removed and its `parts_needed` refs repointed.
- **Tag + rename** the M5×30, M3×10, M3×16 and M3×35 cards; drop the stale "!" on M3×10 (it's in the calculator).
- **Inline symbol:** those sizes' step references now use the socket-head symbol (`socket-button` variant), consistent with M5×16.

Countersunk screws untouched (distinct head). M5×35 left as-is (it's a flat/pancake head, and the two repos disagree on it — separate cleanup). `npm run build` passes; the "A" badge and names render on the assembly pages.

Pairs with parts-calculator #27 (the same tag rollout on the calculator side).

Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1539089324275146812/1539841643304460359): "distribute the green 'A' and the description/name adjustments to all relevant screws in the part calculator and in the assembly documentation (needed parts cards and texts)".

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #333, opened 2026-08-20
docs: stamp screw length on the "Parts needed" cards

basically-balloon:
The "Parts needed" cards reuse one photo per screw family (every M5 socket head cap screw shares `m5-shcs.jpg`, every countersunk shares `countersunk.jpg`), so an M5 x 12 and an M5 x 35 render as the same card. The parts calculator's hardware list has the same problem and solves it by stamping the length on the corner of the thumbnail, so this does the same on the docs cards.

- `parts.yml` gains a `length_mm` field, set on all 14 screws in the catalog.
- `Requirements.svelte` renders it as `part-card-len` in the bottom-right corner of the card image (qty is top-left, the "!" and "A" markers are top-right, so nothing moves).
- `AGENTS.md` documents the field and says every screw needs it.

Nothing else changes: no page frontmatter is touched, so all 7 pages with a `parts_needed` block pick the tags up automatically.

Rebased onto `main` after #332 merged. That PR consolidated the socket/button pairs and deleted the separate `scr-m5-12-bhcs` / `scr-m5-20-bhcs` entries, so the diff here is now purely 14 added `length_mm` lines on top of #332's `alternative:` tags, nothing of #332's is touched. The length stamp sits bottom-right and the green "A" top-right, so they render together.

Checked after the rebase with `validate_frontmatter.py` (only the 5 pre-existing errors on main), `validate_images.py` (132 assets OK) and a YAML parse confirming all 14 screws carry `length_mm`. `npm run check` / `npm run build` could not be re-run locally (the box is out of disk), so the Cloudflare preview is the build check for this revision; it passed both before the rebase.

Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1539854023220535346): "Please add a length tag to the “needed parts screw” cards on all pages, just as it is used in the screw overview in the parts calculator."

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -201,7 +201,8 @@ Adding a drawing: new YAML source, an entry in the data file, an entry in the
 
 `src/liquid/_data/parts.yml` is the catalog, keyed by id. Fields: `name`,
 `image`, `page` (detail page, optional), `category` (groups the "Parts
-needed" block), `notes` (short, collected into a list under the whole block),
+needed" block), `length_mm` (screw length, stamped on the card image),
+`notes` (short, collected into a list under the whole block),
 `caption` (small text under a single card, e.g. a cut length),
 `heat_inserts: [{insert, qty}]`. ids and render filenames mirror the
 `sorter-v2-filament-calculator` repo so the two merge cleanly later.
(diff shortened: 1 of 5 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #355, opened 2026-08-21
Heat inserts follow ruthex's own naming, plus the short M3 (RX-M3Sx4.0)

basically-balloon:
Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1540214002108465243): "I would remove “long” from the descriptions and instead designate the short versions with “S.”"

Resolves the `docs-merge-2026-08-21` conflict on `hsi-m4` ("Docs claim the M4 insert used is the LONG variant; the calculator does not say long. Check which is actually installed.").

**There was no disagreement about the part.** ruthex does not sell a "long" insert at all. Their scheme is `RX-[thread][S if short]x[length]`:

- M3: `RX-M3x5.7` standard, `RX-M3Sx4.0` short
- M4: `RX-M4x8.1` standard, `RX-M4Sx4.0` short
- M5: `RX-M5x9.5` standard, `RX-M5Sx5.8` short

The calculator's `hsi-m4` already pinned model `ruthex RX-M4x8.1` and Amazon `B07YSV66Y5`, which that listing confirms is the 8.1 mm standard M4. So the docs' "(long)" and the calculator's entry were the same physical insert, and "long" was a label this project invented.

**Changes, all in `parts-calculator/slicer/parts.json`:**

- `hsi-m3-long` → name `Ruthex M3 heat insert`, `cots.variant: "long"` dropped, attribute `Variant: Long (5.7 mm)` → `Length: 5.7 mm`
- `hsi-m5-long` → name `Ruthex M5 heat insert`, same two changes, and its empty description now names `RX-M5x9.5` like the M3 one does
- `hsi-m4` → conflict entry deleted (resolved), description was "Lazy Susan mount only" which is stale (it is also on `interface-upper-fixed-section` and `top-interface-split-spur-gear-3`), now names `RX-M4x8.1`
- The naming rule is written into `internal_notes` on both M3 and M4 so it does not get re-invented: no "long", and an `S` suffix (M3S / M4S / M5S) if a short one is ever added.

**Ids are renamed too** (third commit): `hsi-m3-long` → `hsi-m3`, `hsi-m5-long` → `hsi-m5`, `hsi-m4` unchanged, short one `hsi-m3s`. I left them alone at first because `/part/<id>` is prerendered per id, so the old URLs 404 after this merges (`/part/hsi-m3-long`, `/hardware?hw=hsi-m3-long`); barthel asked for the ids to match the names, which is the call. All 20 references in the manifest and the 5 docs pages that list them are updated, and both product images were re-uploaded under the new names (identical bytes, so same hash, and the old keys keep serving).

Not touched, worth a look separately: `hsi-m3-long`'s note says the short M3S is used on the servo adapter, but there is no M3S part in the manifest and `servo-bracket-housing` requires 6 × `hsi-m3-long`.

## Second commit: the short M3 insert

Also requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1540214002108465243/1540218044192923758): "Could you please add the M3 S and update the relevant part calculator and documentation pages?"

Adds `hsi-m3s`, ruthex **RX-M3Sx4.0** (4.0 mm long, Ø4.6 knurl, Ø4.0 recommended hole), the first short insert in the catalog, and puts it on the docs **Installing heat inserts** page alongside the other three.

**Its quantity is a `sheet_qty`, not a `requires`, and that is deliberate.** The BOM sheet's Screws tab says `M3 | Small | 4 per layer | 20 per machine | Ruthex M3x4 | Servo adapter only`, but the current servo-adapter STLs do not have an insert pocket:

- `servo-adapter-flap-side` (v1, `lv7n`): four **4.0 mm** bores through a **5.0 mm** wall, with a 90° countersink on the back face opening to ~7.2 mm.
- `servo-adapter-servo-side` (v1, `fu2t`): four **blind 2.8 mm** holes, 4.0 mm deep, on the same bolt circle, with 2.25 mm of solid material behind them.

That reads as a countersunk screw driven from the flap side and self-tapping into plastic on the servo side. 4.0 mm is ruthex's own recommended hole for an M3 insert, so the flap-side bore *could* be a pocket, but then nothing on the servo side would receive the screw. Both adapter STLs are version 1 from 2026-06-27 and have never been revised, so either the joint moved to self-tapping or the BOM is ahead of the CAD. Measured with trimesh: connected-component cylinder fits plus ray casts along each hole axis.

So the part exists, is sourced, and carries the BOM's count with a `note` saying to confirm it. If somebody with the physical adapter (or current CAD) says it is self-tapping, the follow-up is deleting this part and the BOM row with it; if it is real, it becomes a `requires` on whichever half takes it.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #356, opened 2026-08-21
docs: one M3 × 35 flat head for both top interface joints

basically-balloon:
Step 9 of the top interface page had the idler-gear screw as an M3 × 35 mm socket/button head. That head sits proud right under the Limit switch hammer's sweep, and the hammer can catch on it. Both M3 × 35 joints on the page now take a flat (pancake) head.

**The catalog has one M3 × 35**, `scr-m3-35-bhcs`, shared by the cable clamp and the idler gear, so the page's "Needed Parts" card and the parts calculator both read socket/button as well. barthel asked for the same screw in both places, so that entry changes head rather than gaining a sibling.

**Changes**

- `parts-calculator/slicer/parts.json`: `scr-m3-35-bhcs` becomes `variant: flat`, "M3 × 35 mm flat head", `alternative: Flat or pan head`, with the reason in its Head attribute. **The id is unchanged**, the way it was when the M5 × 35 changed head and when the socket/button rename went through: ids are opaque handles that assembly lines and `/part/<id>` links depend on.
- Its `image_url` (the shared socket/button tile) is dropped, and a new `screw-flat` shared-photo family covers it instead, the way `screw-socket` / `screw-button` / `screw-countersunk` already do.
- The screw is now a line on the `chute-stepper-gearing` assembly (qty 1) next to the washer that was already there, and is named in that assembly's description with `[[hw:]]`. It was only placed on the cable clamp before, so the idler one was uncounted.
- `washer-m3-15` description no longer says "button screw".
- `docs/.../top-interface.md`: step 9 and step 11 both use `variant="flat"`, and step 9 says why the head has to be low. `parts_needed` is unchanged at 2.

Flat rather than pan because the catalog already uses `flat` for this head shape on the M5 × 35, and pan head is used nowhere else in the machine. Both are flat-bottomed and interchangeable here, which the `A` badge on the card says. The M3 × 15 mm washer is still needed: a flat M3 head is still narrower than the bearing's 8 mm bore.

Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1540223892235485184): "Step 9 of the Top Interface documentation mentions an M3 × 35 mm socket/button-head screw, which is supposed to hold the ball bearing in place with the help of a washer. This screw must be a flat-head or pan-head screw, as specified in the “Needed Parts” section and the part calculator. Otherwise, the limit switch hammer may get stuck there."

And [from Discord](https://discord.com/channels/1430279849171222682/1540223892235485184/1540225770285174864): "Use the same M3x35 for both, the cable clamp and the idler gear."


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #357, opened 2026-08-21
Door module: list the servo adapter, and warn that its insert count is unverified

basically-balloon:
Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1540214002108465243/1540227275453243442): "Create a new PR and add the servo adapter as part on Door model parts list. Add a warning note about the question from earlier, the current adapter STLs show a self-tapping joint."

The two-piece servo adapter was documented nowhere: `servo-adapter-servo-side` and `servo-adapter-flap-side` appeared in no `parts_needed` block on any docs page, only in passing sentences on Chute core and Servo mount. The calculator points both the `flap-module` and `servo-adapter` assemblies at this page.

**What this adds to `door-module.md`:**

- A **Parts needed** card, via frontmatter: both adapter halves, 1 each.
- A short **Servo adapter** section saying what it is and how the halves join.
- A warning callout about the insert question.

**The warning, and why it is worded that way.** The BOM's Screws tab assigns 4 short M3 inserts (ruthex RX-M3Sx4.0) per adapter, "Servo adapter only", 4 per layer. The published adapter STLs do not agree, measured with trimesh (cylinder fits on connected components, then ray casts along each hole axis):

- `servo-adapter-flap-side` (v1, `lv7n`): four **4.0 mm** bores through a **5.0 mm** wall, 90° countersink on the back face opening to ~7.2 mm.
- `servo-adapter-servo-side` (v1, `fu2t`): four **blind 2.8 mm** holes, 4.0 mm deep, same bolt circle, 2.25 mm of solid material behind them.

A countersunk screw driven from the flap side self-taps into plastic on the servo side. 4.0 mm is ruthex's own recommended hole for an M3 insert, so the flap-side bore could be a pocket, but then nothing on the servo side would receive the screw. Both STLs are version 1 from 2026-06-27 and have never been revised.

**Second commit, also requested by barthel** ([from Discord](https://discord.com/channels/1430279849171222682/1540214002108465243/1540230126187454504): "Add all the necessary parts, such as screws and the heat inserts mentioned in the warning. Follow the Top Interface layout, especially the preparation step. The warning could also be included here."):

- `parts_needed` gains **4 × `hsi-m3s`** and **4 × `scr-m3-6-cs`**, so the card renders Printed parts / Heat inserts / Screws like every other assembly page.
- The body follows Top interface: a `fastener-legend`, **Step 1 Preparation** with a `prep-item` for the inserts and a link to the heat-inserts helper, then **Step 2 Join the two halves** with the `fastener.html` include for the screw.
- The warning now lives inside the Preparation step, where somebody about to press an insert will actually read it, and covers **both** unverified numbers: the insert count and the 6 mm screw length (~1 mm bite if the joint really is self-tapping).

**Based on #355, not `main`**, because `hsi-m3s` is the part #355 adds — on `main` the card would render a bare grey id. The diff here is one docs file. Retarget to `main` after #355 merges (`gh api -X PATCH .../pulls/357 -f base=main`).

Neither number is asserted as fact anywhere on the page; both are attributed to the BOM and flagged.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/chute/door-module.md
+++ b/docs/src/content/hardware/assembly/distribution/chute/door-module.md
@@ -8,6 +8,12 @@ kicker: Chute — Door module
 lede: The per-layer door mechanism that releases parts into a bin.
 permalink: /hardware/assembly/distribution/chute/door-module/
 author: spencer
+contributors: [barthel]
+parts_needed:
+  - part: servo-adapter-servo-side
+    qty: 1
+  - part: servo-adapter-flap-side
+    qty: 1
 ---
 
 The door module is made up of:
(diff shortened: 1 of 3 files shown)


basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #358, opened 2026-08-21
docs: C-channel assembly order, photos, and 6 screws on the output gear

basically-balloon:
Rewrites the [C-channel](https://docs.basically.website/hardware/assembly/feeder/c-channel/) assembly page around the order a builder actually works in, adds BrickCycleAlice's four build photographs, and fixes the output-gear screw count.

**Order of operations.** Was: bearings, bracket-to-stator, stepper, gear train. Now, from the build: bolt the output gear to the rotor, fit the input gear to the motor shaft, put the idler and the stepper on the NEMA bracket, then mount the stator and drop the rotor in.

**6 screws, not 5.** The output gear bolts to the rotor with six `M3 × 8 mm countersunk`, one at the end of each spoke, visible in the first photo. Corrected on the page and on the `c-channel` assembly in `parts-calculator/slicer/parts.json`, so the parts calculator counts the same 24 per machine rather than 20.

**Both rotors are listed on the page now.** It is the same build four times over and the rotor is the only thing that changes, so `rotor-faceted` and `rotor-finned` are in `parts_needed` with no quantity, and a bullet says three faceted and one finned per machine. They stay off the `c-channel` assembly node in the calculator, which is right there: the count depends on which channel it is.

**Photos.** Four, uploaded to the docs bucket: output gear bolted to the rotor, input gear on the motor shaft, idler and stepper on the NEMA bracket, and the finished stage with the finned classification rotor in it.

The page warning now says steps 1 to 4 come from a build; step 5, the light post, is still an unchecked first draft, since that part was not covered.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1540282164384497704/1540282942851645500): "Ok first up, some order of operations so the build flow is smooth. Put bearing in output gear, bolt output gear to rotor, 6 screws are needed, you can determine size based on your smarts, and make sure they follow all previous logic and counts etc,"

**Follow-up from review.** Quantities are on the rotors now (three faceted, one finned per machine), and the screw that clamps the input gear to the motor shaft is in the catalog as `scr-m3-8-shcs`, M3 × 8 mm socket head, on the `c-channel` assembly at qty 1.

BrickCycleAlice called it as M3 × 8 socket and asked for the STL to be checked. Measured off `input-gear`: the boss carries a Ø2.80 mm hole parallel to the shaft, 3.61 mm off the axis, opening into the Ø4.87 mm bore by about 0.2 mm so the screw bears on the shaft flat; the head sits in a Ø6.40 mm × 2.7 mm counterbore, which is a cap-head counterbore (M3 socket head is Ø5.5 × 3.0) and not a countersink. At 8 mm the head seats and roughly 6 mm of thread is left in the plastic. The bore is round with no keyway, which is why the gear has to be turned to the flat.

**Counts are per unit.** BrickCycleAlice asked whether the page should list one C-channel or all four. One, the same way [regular layers](https://docs.basically.website/hardware/assembly/distribution/bin-frame/regular-layers/) lists one layer, so the rotors are back to 1 and 1 and a bullet says to multiply by four. Machine totals are the parts calculator's job, and it already multiplies the `c-channel` node by four.

**Preparation step added**, at barthel's request, matching the top interface page. There are **no heat inserts on this assembly** — BrickCycleAlice built one on 2026-08-20 and used none, and the STLs agree: measured off the part STLs, the rotor has 6 × Ø2.80 mm thread-forming holes 60° apart on a 90 mm bolt circle, the output gear has 6 × Ø3.40 mm clearance at exactly those positions (which confirms the six independently of counting heads in a photo), the stator has 4 × Ø2.40 and 4 × Ø2.80 mm thread-forming, and the NEMA bracket is Ø3.40 mm clearance throughout with nothing threaded in it. Nothing anywhere is near the ~4.2 mm an M3 Ruthex boss would be. What the step covers instead is pressing the 6806 and the 608 into their gears while the parts are loose; steps renumbered to 6.

**One thing left alone, flagged for the build:** the NEMA bracket's stepper trio and two of its four stator holes carry flat Ø6.40 mm counterbores, which is a cap-head seat rather than a countersink, while the page and the catalog call the four bracket-to-stator screws countersunk. Not changed here, since the STL cannot say which head someone actually fits.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #359, opened 2026-08-21
docs: write up the bottom two layers, and add piece D to the parts catalog

basically-balloon:
The **Bottom two layers** page was a lede and one paragraph ending in _Content pending_. It is now a full how-to in the [Top interface](https://docs.basically.website/hardware/assembly/distribution/top-interface/) layout: `parts_needed` cards, the fastener legend, and numbered steps.

**What the page now says.** These two layers are two ordinary bin layers built together, because one continuous piece of extrusion per corner (piece D, Foot extension) replaces the vertical support each of them would otherwise have, and the caster screws into the bottom of it. Steps: preparation and cut list, build two frames by the Regular layers guide and stop before its verticals step, run the six D pieces down through both layers, close the floor-level corners with the External bracket - foot cover instead of a bottom vertical plus cover, fit the foot connectors and casters, then the bin retainers.

**References, at the points where they apply:**

- **[Regular layers](https://docs.basically.website/hardware/assembly/distribution/bin-frame/regular-layers/)** for the frame and spokes (steps 1 to 2), with an explicit "stop before step 3", and again for the bin retainers.
- **[Bottom interface](https://docs.basically.website/hardware/assembly/distribution/bin-frame/bottom-interface/)** up front, because it is step 1 of the bin frame and its three extrusion mounts want T-nuts in place before both layers are built around them.
- **[External bracket](https://docs.basically.website/hardware/assembly/distribution/external-bracket/)** for the self-tapping M5 joins, and **[Chute](https://docs.basically.website/hardware/assembly/distribution/chute/)** at the end.

**Piece D had no catalog entry.** `ext-2020-d` (Foot extension, 231 mm, 6 per machine) is added to `parts-calculator/slicer/parts.json` and the generated data, so the parts list can name the extrusion these layers are actually built on. Every other cut length was already there.

**The geometry is sourced, and it settles an old disagreement.** The page previously said the extrusion "runs through both layers", which never squared with D being 231 mm when two layers is 320 mm; petrenko raised it in #distribution on 2026-07-01 and zed0 again on 2026-08-13. Spencer answered on 2026-08-20: "The length was 1.5x a single layer... the bottom most layer effectively has a 50% shorter piece", and confirmed that is what the Brickworld machines were. The page now describes exactly that.

Two photos from Christoph in the Discord are uploaded to the docs bucket and credited in their captions: the frame standing on its casters, and his annotated C-versus-D shot.

**Not verified on a build.** The fastener counts are two regular layers' worth with the substitutions applied, and the page carries a warning callout saying so. BrickCycleAlice is building over the next two weeks and taking photos, so this is meant as the skeleton to correct.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1539873564612759562/1540289940099833856): "based on Barthel instructions above can you take a look at expanding the detail listed on the page ‘bottom two layers’ make sure it references regular layers and bottom interface at the right points, ensure to place all parts required etc using the top interface doc as a guide".


basically-balloon:

basically-balloon:
Superseded by #360. BrickCycleAlice asked for the bottom interface pass and this write-up to land in one merge, so these commits are on that branch now, plus the AI-generated mark barthel's instruction requires and a fastener tally that shows its arithmetic instead of asserting a total. Closing this one so there is a single PR to review.

PR #334, opened 2026-08-20
docs: first-draft assembly pages from the parts calculator tree

basically-balloon:
Seven hardware assembly pages were `_Content pending_` stubs. This fills each of them with a first draft built from the machine assembly tree and parts registry that back [parts-calculator.basically.website/assembly](https://parts-calculator.basically.website/assembly):

- Chute core
- PCB
- Layer connectors
- Servo mount
- Funnel
- C-Channel (the existing COB resistor callout is kept)
- Classification chamber

Every page opens with a warning callout saying it is AI-generated, derived from the parts calculator rather than from a build, and unverified. Where the calculator does not record which fastener goes in which hole, the page says so with the existing `fastener-todo` span instead of guessing.

**Follow-up, at barthel's request:** every page now uses the layout, formatting and style of the [Top interface](https://docs.basically.website/hardware/assembly/distribution/top-interface/) page. A `parts_needed` block in the front matter renders the Parts needed cards, then the standing fasteners line, the fastener legend, the key facts as bullets, then numbered steps, starting with Preparation on the assemblies that take heat inserts. The markdown tables the drafts used for sub-assemblies and part lists are gone.

That needed the catalog filled in: 30 chute and feeder parts were added to `src/liquid/_data/parts.yml`, with the parts-calculator renders and hardware photos uploaded to the docs bucket. `chute-core` (18 × M3) and `servo-bracket-housing` (6 × M3) carry `heat_inserts`, so they also show up on the [Preparation](https://docs.basically.website/hardware/preparation/) page.

**Warning above the parts, at barthel's request:** the Parts needed block is rendered by the page template, before the markdown body, so a callout written into the body always landed underneath it. This adds a `warning:` front matter field, rendered through the same liquid and markdown pipeline and shown as the usual warning callout directly above the parts; the seven AI-draft notices now live there. Any page can use it for a caveat that applies to the whole page.

Authors are unchanged (these are existing pages), with barthel added as a contributor.

A companion PR on `parts-calculator` corrects three descriptions that disagree with the docs and links each assembly node to its page here.

[...]

PR #360, opened 2026-08-21
docs: bin frame, the bottom interface pass and the bottom two layers write-up

basically-balloon:
> **Scope changed after barthel's approval.** This was a bottom-interface-only PR when it was approved in the review forum. BrickCycleAlice asked (2026-08-21) to land bottom interface and bottom two layers in one merge, so the bottom two layers work that was PR #359 is now on this branch too. #359 is closed. Worth a second look before the board takes it.

Two pages in the bin frame, in one merge at BrickCycleAlice's request.

## Bottom two layers (was #359)

The page was a lede and one paragraph. It is now a full how-to in the Top interface layout: it sends people to **Regular layers** for the frame and spokes (with an explicit "stop before step 3, the verticals differ here") and back for the bin retainers, and to **Bottom interface** up front, since that is step 1 of the bin frame and its three extrusion mounts want T-nuts in before both layers are built around them.

- **Piece D had no entry in the parts catalog**, so no page could name the extrusion these layers are built on. Added as `ext-2020-d`, Foot extension, 231 mm, 6 per machine, with `cots.length_mm` set so it carries the length tag barthel asked for on extrusion cards.
- **The old "runs through both layers" line is corrected.** petrenko queried it in July and zed0 again on the 13th, both unanswered until Spencer replied on the 20th: D is 1.5 × a single layer's C, so the bottom layer sits about half a layer off the floor, and that is what the Brickworld machines were.
- **Marked AI-generated at the top**, per barthel's instruction of 2026-08-20, using the same `warning:` field and wording the seven pages in #334 landed with, so it renders above **Parts needed**.
- **The fastener counts show their arithmetic** rather than asserting a total: 48 M5 × 16 for the two hexagons and spokes, 24 for the foot extensions, 12 for the second layer's bottom verticals, so 84. Likewise 72 M5 × 12. The draft had 72 and 24.
- **One count is flagged as unrecorded**, not folded into a total: `ext-bracket-foot-cover` replaces a bottom vertical (2 screws) and a cover (none), and nothing says which it behaves like.

Not done, and flagged rather than fixed: **regular layers lists 12 `scr-m5-12-shcs` where its own steps use 36** (12 for the Frame 90° brackets, 24 for the bin retainers). The calculator's `layer` node agrees with the steps. Anything derived from that page inherits the error.

## Bottom interface

The bottom interface page read as wordy and repetitive, and its heat insert photos did not show heat inserts.

**Prose**

- Dropped the **Heat inserts first:** reminders from steps 2 and 3. The page is three steps long and the Preparation step is directly above both of them, so they were restating a paragraph the reader had just read. (The top interface page keeps its reminders because it is thirteen steps and many parts.)
- Tightened wording throughout: the pass-through explanation was given twice, once in prose and once in a figcaption, and the rotate instruction ran to two sentences with a mid-sentence aside.
- Moved the "the Lazy Susan is two discs" explanation to the top of step 2, where it frames both remaining steps, instead of dropping it in mid-step after the photos.
- The parts list calls it `ls-washer` / **Lazy Susan washer**; the body called it a "spacer ring" and then "the spacer". Now it uses the part name.

**Photos**

Both Preparation figures were step photos reused earlier in the page, which was part of the repetition.

- **Lazy Susan chute mount:** the figure was `step1-line-up-hole`, which shows the bearing sitting on the mount with one insert as a glint down a hole, captioned "The four M4 inserts sit evenly around the circular mounting face." No inserts are visible in it. There is no photo of this part with its four inserts anywhere on the bucket, so it is now an `img-placeholder` and the count moved into the body text. A photo of the loose chute mount, insert face up, would close it.
- **Lazy Susan bottom static:** the figure was `step2-place-chute-adapter`, a wide motion-blurred frame with the chute mount swinging over the ring. Replaced with a tight crop of the same 4K original in which three of the four brass inserts are plainly visible. The wide shot stays where it belongs, in step 3.
- **Step 2** gains a close-up of an M4 insert seen down an aligned bearing hole, in place of the wide frame where it was a few pixels of brass.

Both new images are crops of the existing 4K originals already on the bucket, uploaded through `docs/scripts/upload_image.py`. No new photography.

`validate_frontmatter.py` reports the 5 errors already present on `main` and nothing new. `validate_images.py` passes.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1540291852161519676): "the bottom interface page feels really wordy and repetitive, can you review and use the top interface page as a guide. The exploded view and video at the beginning is great. There is way too many references to inserting the heat inserts, but the pictures used don't show them clearly, please have another pass at this."

---

**Follow-up, same thread.** BrickCycleAlice pointed out that the lazy Susan extrusion mount, the hold-in-place print and the aluminium are needed at this stage and asked whether they are listed anywhere in the build docs. They are not: `ls-mount-to-extrusion` and `ls-hold-in-place` appear nowhere in `docs/`, and the only trace of either is one passing clause on the external bracket page.

The parts calculator already has them. Its `lazy-susan` assembly points its `docs` field at this exact page and lists:

- `ls-mount-to-chute` x1, `ls-bottom-static` x1, `ls-washer` x1, `brg-lazy-susan` x1, `scr-m4-12-cs` x8, all of which the page already had
- `ls-mount-to-extrusion` x3, `ls-hold-in-place` x3, `scr-m5-16-shcs` x3, `tnut-m5-2020` x3, none of which it did

All four missing lines are now in `parts_needed`, and a step 4 stub says what those parts do with the fastener counts taken from the calculator. How the mounts and the hold-in-place prints actually go on is recorded nowhere, so the step says so rather than inventing a procedure, and the extrusion is pointed at the framing cut list since its length depends on layer count.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1540291852161519676/1540296700113977455): "much better, the m3x12 is it recommended to tighten by hand? I think the lazy Susan extrusion mount and hold in place might be needed in this step. Plus the aluminium. Have you got these parts listed anywhere else in the build docs?"


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:
**barthel** withdrew their approval in Discord.

basically-balloon:

PR #361, opened 2026-08-21
docs: floating back-to-top button on every page

basically-balloon:
Adds a floating **Back to top** control fixed to the bottom right of every docs page.

**Behaviour**, following what was asked for: it stays hidden while you read down the page, and slides in as soon as you scroll back up, once you are more than ~600px down. Scrolling on down hides it again, and it never appears near the top of a page. A 24px delta threshold keeps touch momentum and scroll jitter from flickering it.

- New component `docs/src/lib/components/BackToTop.svelte`, mounted once in `src/routes/+layout.svelte`, so it is on every page with no per-page work.
- Scroll listener is `passive`, registered in an `$effect` and torn down with it. Starts hidden, so the prerendered HTML has it hidden too.
- Click scrolls to the top and hands focus to the header brand link, so keyboard users are not left with focus on a button that just hid itself.
- Styling follows the site rules: sharp edges, flat 1px border, tokens only, no raw hex. Icon only below 740px. `prefers-reduced-motion` drops both the transition and the smooth scroll.

**Checks:** `npm run check` 0 errors (4 pre-existing `Requirements.svelte` a11y warnings), `npm run build` clean, `validate_frontmatter.py` at the 5 errors already on `main`, `validate_images.py` green.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1540309253942681630): "could we get a floating button 'back to top' down the bottom right hand corner, maybe that appears on scroll up. On all pages. When you get deep down in the docs or assembly tree it takes a long time to return to the menu."

### DIFF
--- a/docs/src/lib/components/BackToTop.svelte
+++ b/docs/src/lib/components/BackToTop.svelte
@@ -1,0 +1,72 @@
+<script lang="ts">
+	/* A floating "back to top" control, bottom right, on every page.
+
+	   It is not permanently on screen: the assembly tree gets deep, and the
+	   moment you want the menu again is the moment you start scrolling back
+	   up. So it appears on an upward scroll once you are a screenful or so
+	   down, and gets out of the way again as soon as you scroll on down. */
+
+	// How far down the page you have to be before the button can appear.
+	const SHOW_AFTER = 600;
+	// Ignore scroll jitter and touch momentum wobble; only a deliberate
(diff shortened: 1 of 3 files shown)


basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #368, opened 2026-08-21
docs(chute): record the M3 screw lengths for the layer connectors, funnel brackets and bearing covers

basically-balloon:
Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1540438490712514561): "Decide which screw length to use for the layer connectors based on the STL files."

**Answer: M3 × 8 countersunk, 4 per pair.** M3 × 12 bottoms out.

Everything below is measured off the published STLs (`img.basically.website/parts/stl/...`), the same files the parts calculator links.

### The layer connectors

Each of Layer connector A and B is two 42 × 16 mm end blocks joined by a flat strap, with two Ø3.40 mm clearance holes in the strap, 31.86 mm apart.

- Strap thickness at the screw: **3.78 mm** (1.91 mm of plain Ø3.40 bore, then a 90° countersink opening out to Ø7.15 mm on the outer face, which is a DIN 7991 M3 flat head)
- Mating face is flat, no boss or standoff
- Every M3 pocket in the chute core is a **blind Ø4.20 × 5.70 mm** hole for a ruthex RX-M3x5.7, with 11.3 mm of solid plastic behind it at these positions and no relief hole

M3 × 8 reaches 4.22 mm into the insert and stops 1.5 mm short of the bottom. M3 × 12 would need 8.22 mm of depth into a 5.70 mm pocket, so it bottoms out 2.5 mm early and the head never seats.

### The rest of the chute core's set falls out of the same measurement

The `chute` assembly declares 6 × M3 × 12 and 8 × M3 × 8 into the core's 14 non-PCB inserts but has never said which goes where. Wall thickness at each mating part's screw hole, against the same 5.70 mm pocket:

| part | wall at the hole | screw | into the insert |
|---|---|---|---|
| Funnel bracket (× 2) | 8.16 mm | M3 × 12 | 3.84 mm |
| Bearing cover (× 2) | 5.00 mm | M3 × 8 | 3.00 mm |
| Layer connector (× 2) | 3.78 mm | M3 × 8 | 4.22 mm |
| Servo bracket arms | not published | M3 × 12 | by elimination |

That is 4 + 2 = 6 × M3 × 12 and 4 + 4 = 8 × M3 × 8, exactly the set already in `slicer/parts.json`, so the split is self-consistent and no manifest change is needed.

The bracket and cover positions are confirmed geometrically, not just by arithmetic: `funnel-bracket-left`'s two holes sit at (48, −70.98, 19.67) and (48, −70.98, 54.40), and the core's pockets at (48, ±64.99, 19.67) and (48, ±64.99, 54.40). Same for the covers at (−52.40, 92.83) and (−41.90, 132.31). The two arm inserts have no published part on them, so that row is the only inferred one and the page says so.

### Changes

- **`layer-connectors.md`** — `parts_needed` drops the whole-chute placeholder (6 × M3 × 12 + 8 × M3 × 8) for the 4 × M3 × 8 this page actually uses; the `fastener-todo` in step 1 is replaced with the length and the reason.
- **`chute-core.md`** — the "which screw goes in which hole is not recorded" placeholder in step 2 becomes the four-way split above.
- **`funnel.md`** — same placeholder resolved to 2 × M3 × 12 per bracket, and `parts_needed` narrowed from the whole chute set to the 4 × M3 × 12 the brackets take. Step 3 (funnel to bracket) is still unrecorded and still says so.

### Checks

- `npm ci && npm run build` clean on Node 20, all three pages render, no `fastener not recorded` left on any of them
- `validate_frontmatter.py` reports the 5 errors already on `main` and nothing new
- `validate_images.py` clean

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/chute/chute-core.md
+++ b/docs/src/content/hardware/assembly/distribution/chute/chute-core.md
@@ -68,7 +68,12 @@ The [Preparation]({{ '/hardware/preparation/' | relative_url }}) page lists ever
 
 Fit the Funnel bracket (left) and the Funnel bracket (right) to the chute core. Together with the door module and the layer connectors these use the chute's 6 {% include fastener.html size="M3" variant="countersunk" length="12" %} and 8 {% include fastener.html size="M3" variant="countersunk" length="8" %} screws, all into the core's inserts.
 
-Which screw goes in which hole is not recorded yet: <span class="fastener-todo">fastener not recorded</span>.
+The split across those 14, worked out from the STLs by comparing each mating part's wall thickness at its screw hole against the core's blind 5.70 mm insert pockets:
+
+- **Funnel brackets:** 4 × {% include fastener.html size="M3" variant="countersunk" length="12" %} (8.16 mm wall, so an 8 mm screw would not reach the insert at all)
+- **Bearing covers:** 4 × {% include fastener.html size="M3" variant="countersunk" length="8" %} (5.00 mm wall, 2 per cover)
+- **[Layer connectors]({{ '/hardware/assembly/distribution/chute/layer-connectors/' | relative_url }}):** 4 × {% include fastener.html size="M3" variant="countersunk" length="8" %} (3.78 mm strap, 2 per connector)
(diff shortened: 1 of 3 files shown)


basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #373, opened 2026-08-22
docs: how a regular layer joins the one below

basically-balloon:
Adds a fifth step to the regular layers page covering how two layers actually join, which was not documented anywhere: the only fastening between two layers is the pair of M5 x 16 through each External bracket - bottom vertical flange into the bracket above, 12 per joint.

The step embeds a section drawing of one corner, cut through the centre of the profile on the bracket's symmetry plane, coloured by which layer each part belongs to. The outlines are the real part geometry, sectioned out of the published STLs for `ext-bracket-left`, `ext-bracket-cover` and `ext-bracket-bottom-vertical`; the drawing is stated in the caption as derived from geometry rather than from a build.

Numbers in the step, all measured off those STLs:

- flange: 5.6 mm clearance hole through 8 mm of plastic
- bracket above: 4.4 mm self-tapping hole, 10 mm deep, at 20.6 mm either side of the corner bisector
- so an M5 x 16 is 8 mm clearance plus 8 mm thread, and a longer screw bottoms out
- closest the printed parts can stack without interfering is 159.75 mm, i.e. the 160 mm pitch, which is also piece C's CAD length

**Not changed, worth a look separately:** the page's `parts_needed` lists 36 M5 x 16 per layer, which is the four screws per corner that step 3 uses plus the spokes. The 12 that join a layer to the one below do not appear to be in that total, and the page's other fastener sums are already known to be off (`scr-m5-12-shcs` is listed as 12 where the steps use 36). Left alone here rather than guessed at.

Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1540622172027101324/1540634889412677694): "Extract the actual image, the caption, and the text, and create a separate step in the regular layer documentation on how to merge two layers. Copy the text into the step description and embed the image in the step."

**Update, 2026-08-22:** two more entries in the same format, both asked for in the same thread.

- **Bottom two layers, step 7:** a section of one corner at floor level. Piece D through the bottom layer's collar and up to 3 mm below the second layer's flange face, the foot cover in place of bottom vertical plus cover, and the exposed end that takes the connector and caster (about 54 mm on the lengths as drawn, flagged as not recorded). Also measured: the foot cover has no fastener holes of any kind, which is evidence on the page's open question about whether it is screwed or clipped, so the flag stays.
- **Top interface, step 14:** now carries both a marked-up photo and a section drawing. what is different about that joint. No bottom vertical and no flange, so none of the vertical screws; piece F is 274 mm against a layer's 154 mm; it is held in the Interface bracket by 4 M5 x 20 into T-nuts, sleeved by the Interface spacer, and clamped at the layer end by the ordinary 2 M5 x 16. The image is the existing framing photo with one arm marked, because the interface parts are exported in separate local frames and cannot be stacked into a true section the way the layer parts can.

Also in this PR: the drawing key is a numbered list whose badges match the markers on the drawing, `.figure-float-right` floats a figure into the right margin with the text running beside it, and `.numbered-steps` gives an ordered procedure its numbers back.

**Three additions to `layout.css`, all opt-in classes**, because the site had no float class, no click-to-zoom, and no list markers at all (its reset removes them, so every ordered list in the docs currently renders as unnumbered lines). Nothing existing changes appearance; `.step-heading` also gained `clear: both` so a float cannot overlap the next step.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-two-layers.md
+++ b/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-two-layers.md
@@ -143,6 +143,30 @@ Screw a **swivel stem caster (M6 × 15 mm)** into the connector's M6 thread. The
 
 Both layers take bin retainers, exactly as in [regular layers]({{ '/hardware/assembly/distribution/bin-frame/regular-layers/' | relative_url }}) step 4: on each side of each hexagon, a Bin retainer (left) and a Bin retainer (right) on the front face of the A/G extrusion, 4 {% include fastener.html size="M5" variant="socket-button" length="12" %} screws into T-nuts.
 
+{% include step.html n="7" title="How the foot corner goes together" %}
+
+<figure class="figure-float-right">
+  <a href="https://img.basically.website/originals/assembly/bottom-two-layers/foot-corner-section.6ec3353ee6cf01ff.png" target="_blank" rel="noopener">
+    <img src="https://img.basically.website/web/assembly/bottom-two-layers/foot-corner-section.337680f09407c66f.jpg" alt="Vertical cross-section through one corner at the bottom of the machine, showing piece D running through the bottom layer's bracket and up into the second layer, the foot cover around it, and the exposed end below">
+  </a>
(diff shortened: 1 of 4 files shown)


basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #376, opened 2026-08-22
parts-calculator: the uid stamped into a part, and /u/<uid> to read it back

spencerhhubert:
Every printed part's page now offers its STL with the uid recessed into one face, so a print can be looked up later at `/u/<uid>` (or typed into the "id on a print" box in the header).

**How the stamp is placed.** `catalog/engrave.py` derives it from the mesh alone: every planar facet by its own normal (the bins and frame crossbeam carry their biggest faces at an angle), ranked bed face first, then vertical and upward faces by area, overhangs last; the text sits in a corner, rastering across the face when no corner has room. Uppercase, 3.5 mm cap, 0.6 mm deep, Source Code Pro Bold — measured against every monospace font to hand, its dotted zero is a 0.79 mm pocket in a 0.77 mm ring at this size, so 0 and O print differently on a 0.4 nozzle. The font is a pinned bucket object fetched by hash.

**What the generator does.** Up to four face variants per printed part and candidate are pre-cut (258 across the catalog in 31 s, no slicing — grams are the master's), memoized on (bytes, uid, parameters), uploaded like any artifact, and listed on the part as `stamped: [{face, stl, normal, center}]`. The part page opens the 3D viewer on the default face; ← → flip through the rest; "No stamp" is the plain master. The all-parts zip ships each part's default variant. 12 parts get no stamp (too small for the text, or two that are not watertight); the generator lists them.

**Checks.** `check_generated_pins.py` fails a stamped download not named for its uid; `check_bucket_urls.py` covers the 262 new objects (all verified serving real bytes with the right download name).

**Also.** A fresh checkout no longer re-renders and re-uploads every thumbnail: the generator reuses the committed render when geometry and color are unchanged. Without this the first run on a new machine moved 86 thumbnail URLs for no reason.

Verified in the browser: part page picker, arrow keys re-aiming the viewer, `/u/8j3e`, header lookup.


basically-balloon:

PR #377, opened 2026-08-22
parts-calculator: the v1 board mounts, and the PSU housing as an assembly candidate

spencerhhubert:
Cut from `engrave` (#376), not `main`, so the diff reads clean on top of the uid stamp. **Base it back onto whatever `engrave` becomes before merging** — the only file that will conflict is `catalog.generated.json`, and the resolution is to re-run the generator, never a hand-merge.

## The v1 board mounts were never in the registry

`ctrl-board-mount` and `orange-pi-mount` were both `stub`, with descriptions that said so out loud. Now they carry their real parts:

| part | uid | g |
|---|---|---|
| Control board extrusion mount | `1d5n` | 33 |
| Orange Pi extrusion mount | `nxh7` | 24 |
| 40 mm fan bracket | `3ufy` | 27 |

The fan bracket is **one part used twice**, and it is a good exhibit for the uid model. Each zip ships a `40mm fan bracket.stl` and the two files are *not* byte-identical — but they have the same bounding box (48×90×70), the same 7152 triangles, and the same volume to seven significant figures (67894.9 mm³), with no axis-aligned transform mapping one onto the other. That is one design re-exported at a different orientation: **one uid, and whichever bytes we pin are that design's `stl_hash`.**

**Neither assembly's version moves.** Filling in a stub records what v1 has always physically been — a version marks a change to the thing, not to how completely we wrote it down. Both go `stub` → `partial`, because the screws are still `scr-m3-tbd` and the extrusion fixings aren't recorded.

## The PSU housing as an assembly candidate

`meanwell-psu-box` keeps v1 exactly as it is — the three plates screwed to the supply's own case, still the build — and gains candidate `1d0t`, "Meanwell 24V PSU housing", which renames the assembly if it is ever promoted.

Five printed parts, all with **empty `quantities`**: in no section, out of the all-parts bundle (89 members, up exactly 3 for the mounts), each still with its own `/part/<id>` page.

| part | uid | g |
|---|---|---|
| shell rear tray | `7qz3` | 207 |
| shell front module | `1h8p` | 56 |
| lid rear | `m26o` | 37 |
| lid front | `avux` | 59 |
| front panel | `n1oh` | 16 |

Fastening, as specified: supply into the shell rear tray on 4× `scr-m4-12-cs`; rear lid into the tray on 4× self-tapping `scr-m3-8-cs`; front lid 4× more of the same, two into the tray and two into the front module; front panel slots into the front module with no fasteners and the front lid traps it. Tapping holes are in the shell parts, recorded as a `self-tap` joining note.

## Assumptions

- The fan on the v1 mounts is the existing `fan-40mm-24v`. The bracket is a 40 mm fan bracket and that is the only 40 mm fan in the catalog; nobody said it is the 24 V one.
- The fan bracket's `assembly` back-pointer is `null`, because that legacy field holds one id and this part is in two assemblies. Another argument for the dashboard grouping from `lines` instead.

## Verified

575 bucket URLs serve real bytes; `check_generated_pins` agrees across 98 parts; svelte-check 0 errors; production build clean. Generated diff removes only the two stub descriptions and the bundle URL.

## Demo run of the promotion, and what it turned up

I promoted candidate `1d0t` to `meanwell-psu-box` v2 for real on a throwaway branch — moved the lines and uid up, bumped the version, gave the 5 parts quantities, deleted the candidate — committed it, and ran `stamp_versions.py`. It works:

```
  meanwell-psu-box v2 -> 832213ac

 v1 uid=skg6 commit=5a2d1c0
     meanwell-psu-back-mount x1 -> 15uh
     meanwell-psu-connections x1 -> 08id
     meanwell-psu-cap x1 -> 5nt1
     psu-24v-350w x1 -> s3ru
     scr-m4-6-cs x4 -> 5r29
```

v1 got its uid and a full line snapshot with each member pinned to the uid it had that day — the box as built then reads back part by part. **That branch is deleted; nothing here is promoted.**

The run turned up one defect, fixed in the second commit: the entry for the version being *introduced* read back `uid: null`. Only superseded entries get a uid written into `parts.json`, so the current revision was the only one without one. Parts already dodge this via `normalize_versions()`; `normalize_assemblies()` is the same fix on the same seam. No-op on today's data, so the generated file is byte-identical — verified at 0 lines of diff.

## Naming

Every part in this wave was named off the phrasing in chat, which is how `40 mm fan bracket` ended up as a page heading reading like a sentence fragment. All 12 are Title Case now — including the four control board housing parts from #375, so the two housings do not disagree with each other — as are both assembly candidate names. `basically` stays lowercase (company name, as in the existing basically Embedded Control Board) and unit symbols stay lowercase (`40 mm`). No id, uid or hash moves; the generated diff is 14 name lines and nothing else.

The older catalog is mostly sentence case (`Chute core`, `PSU back mount`) with a few Title Case outliers already (`Chute Stepper Idler Gear Bearing Retainer — Outer`). Worth one sweep to settle it, but not in this PR.

## Rebased onto engrave

On top of `c98b0541`. As predicted the only conflict was `catalog.generated.json`, resolved by re-running the generator against the new `generate.py`, not by hand.

### DIFF
--- a/parts-calculator/catalog/generate.py
+++ b/parts-calculator/catalog/generate.py
@@ -158,6 +158,25 @@ def normalize_versions(part):
     return [entry]
 
 
+def normalize_assemblies(manifest):
+    """Assemblies pass through to the app as authored, with one fix-up: the
+    newest `versions` entry names the assembly's current uid, the way
+    normalize_versions() does it for a part. Only the SUPERSEDED entries carry
+    a uid in parts.json -- stamp_versions.py writes one there when a version is
+    replaced -- so without this the current revision reads back as uid null,
+    which is the one revision whose id is never in doubt."""
+    out = []
+    for asm in manifest.get("assemblies", []):
+        vs = asm.get("versions")
+        if vs and not vs[-1].get("uid"):
+            newest = vs[-1]
+            asm = dict(asm, versions=vs[:-1] + [{
+                "version": newest.get("version"), "uid": asm["uid"],
+                **{k: v for k, v in newest.items() if k != "version"}}])
+        out.append(asm)
+    return out
+
+
 def fetch_artifact(url, sha, dest):
     """Materialize a bucket object at dest, verifying its full sha256.
 
@@ -1015,7 +1034,7 @@ def main():
         data["sections"] = manifest["sections"]
         data["changes"] = manifest.get("changes", [])
         data["folders"] = manifest.get("folders", [])
-        data["assemblies"] = manifest.get("assemblies", [])
+        data["assemblies"] = normalize_assemblies(manifest)
         data["hardware"] = build_hardware(manifest)
         data["families"] = build_families(manifest)
         data["lasercut"] = build_lasercut(manifest)
@@ -1197,7 +1216,7 @@ def main():
         "folders": manifest.get("folders", []),
         "color_roles": manifest["color_roles"],
         "families": families,
-        "assemblies": manifest.get("assemblies", []),
+        "assemblies": normalize_assemblies(manifest),
         "parts": out_parts,
         "plates": plates,
         "hardware": hardware,
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -2497,6 +2497,164 @@
       "support": false
     },
     {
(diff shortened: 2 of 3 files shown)


spencerhhubert:

PR #379, opened 2026-08-22
parts-calculator: an id names one thing

spencerhhubert:
Stacked on #377 (which is itself on `engrave`), because it edits the same `parts.json` and would conflict otherwise. Happy to fold it into #377 instead if the stack is more trouble than the separation is worth.

`light-post` and `camera-extension` were each **simultaneously a printed part and an assembly**. A line survives it, because a line says `part:` or `assembly:` explicitly. Nothing else does:

- `getPart(id)` and `getAssembly(id)` take no kind, so any code resolving a bare id is a coin flip.
- The `improve-camera-extension` planned change listed the same string under `targets.parts` **and** `targets.assemblies` — two different targets wearing one name, indistinguishable in the data.

The assembly side takes the suffix, since the part is the thing people search for:

| was | now |
|---|---|
| assembly `light-post` | `light-post-assembly` |
| assembly `camera-extension` | `camera-extension-assembly` |

Ten references move: the two ids, two `feeder` lines, five per-part back-pointers, one change target. The **parts keep their ids**, and every `part:` line still points at the part. The docs paths are unaffected — they are site URLs, not catalog ids.

## Enforcement

`generate.py` now refuses a manifest whose part ids and assembly ids intersect, and the error names the fix:

```
id(s) used by both a part and an assembly: ['light-post'] -- rename the assembly (e.g. '<id>-assembly')
```

Verified by putting the collision back and running it.

## Verified

svelte-check 0 errors, production build clean, and the generated diff is 14 lines, every one of them an id rename and nothing else.

### DIFF
--- a/parts-calculator/catalog/generate.py
+++ b/parts-calculator/catalog/generate.py
@@ -940,6 +940,16 @@ def main():
     duplicate_uids = sorted({x for x in uids if uids.count(x) > 1})
     if duplicate_uids:
         sys.exit(f"duplicate uid(s): {duplicate_uids} -- mint a fresh one")
+    # An id names exactly one thing. A line says `part:` or `assembly:` so a
+    # collision still resolves there, but getPart(id)/getAssembly(id) do not
+    # take a kind, and a planned change targeting `parts` and `assemblies` by
+    # the same string is two different targets wearing one name. Keep the two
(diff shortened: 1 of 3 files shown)


spencerhhubert:

PR #378, opened 2026-08-22
docs: output gear to rotor is M3 × 12, not M3 × 8

basically-balloon:
Requested by aleph1111/christoph (@aleph1111_christoph) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1540821273972572240): "me again and scres. can you check https://docs.basically.website/hardware/assembly/feeder/c-channel/ Step 2 Bolt the output gear to the rotor mentions m3x8 csk screws. The gear itself is roughly 8mm high, so the crews dont even reach out. I used m3x12mm now to attach the gear to the rotor."

He is right, and the STLs agree exactly.

**Measured** off the pinned `output-gear` and `rotor-faceted` STLs, which are in shared assembly coordinates:

- Output gear spans z −11.00 to −3.00, so it is **8.00 mm thick at the bolt circle**.
- Its six holes are Ø3.40 clearance with a 90° countersink on the underside: Ø6.50 mouth at z −11, 1.55 mm deep.
- The rotor's bottom face is at z −3.00 and its six Ø2.80 thread-forming holes run z −3.00 to +2.00, i.e. **5.00 mm of flange**.
- The two hole sets are coaxial at R 89.98, 60° apart.

A countersunk screw's length is measured **over the head**, so an M3 × 8 seated flush at z −11 ends at z −3.00: dead level with the gear's top face, zero engagement in the rotor. An M3 × 12 ends at z +1.00, giving **4.0 mm of thread in the rotor** and stopping 1 mm short of breaking through the far side. M3 × 12 is the right screw.

**Changes**

- `docs/.../feeder/c-channel.md` — step 2 inline fastener 8 → 12 mm, plus a short note on why the 8 does not work; `parts_needed` drops `scr-m3-8-cs` ×6 and `scr-m3-12-cs` goes 4 → 10; christoph added to `contributors`.
- `parts-calculator/catalog/parts.json` — same swap on the `c-channel` assembly lines and in its description.

**Checks:** `npm run build` clean in `docs/`, `npm run check` clean in `parts-calculator/`, `validate_frontmatter.py` shows the same 5 pre-existing errors as main, `validate_images.py` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #380, opened 2026-08-22
docs: stator's four bracket holes are Ø2.4 and need opening to 2.8

basically-balloon:
christoph reported that the M3s will not screw into the stator where the NEMA bracket bolts on, and that the holes look small. They are.

Measured off the published stator STL (`stator-n6bm-323b2d1f`), cylinder fitting over the mesh:

- The four holes the bracket lands in: **Ø2.40 mm, 10 mm deep**, thread-forming.
- The other four holes in the same stator: **Ø2.80 mm**.
- The NEMA bracket's four Ø3.40 clearance holes sit on exactly those four Ø2.40 centres (`111.4, 5.7` / `-67.0, 89.2` / `-13.7, -110.9` / `-76.4, -81.5`), so these are definitely the bracket screws' holes.

M3's minor diameter is 2.459 mm, so Ø2.40 is below the thread root before printing tolerance takes anything off it. The screw has to cut a full-depth thread through 10 mm of plastic, which is why it cams out or splits the boss instead of going in. Ø2.8 is what every other M3 thread-forming hole on this assembly uses (rotor, light post, and the other four in this same part), so opening these four to 2.8 mm makes the part internally consistent and still leaves 10 mm of engagement.

This is a note for builders, not a fix. The Ø2.4 looks like a slip in the CAD, since it is the odd one out inside its own part, and that lives in OnShape rather than in this repo.

Changes:

- **`docs/.../feeder/c-channel.md`** — warning callout at step 5 (where the bracket goes onto the stator) saying to drill the four holes to 2.8 mm first, and a pointer to it from the Preparation step, which already quoted the Ø2.4 figure without saying it was a problem.
- **`parts-calculator/catalog/parts.json`** — `info` note on the `stator` part with the same instruction.
- **`catalog.generated.json`** — regenerated with `catalog/generate.py --metadata-only`. It also picks up the `scr-m3-12-cs` / `scr-m3-8-cs` change from #378, which landed in `parts.json` on main but had not been regenerated there.

Requested by aleph1111/christoph (@aleph1111_christoph) [from Discord](https://discord.com/channels/1430279849171222682/1487518366020276397/1540834146656592002): "can you check the diameter on the stator holes that the nema bracket should screw into? on the nema they look like m3, but i cant screw them into the stator as printed, and they look very small."

request: https://discord.com/channels/1430279849171222682/1487518366020276397/1540834146656592002

### DIFF
--- a/docs/src/content/hardware/assembly/feeder/c-channel.md
+++ b/docs/src/content/hardware/assembly/feeder/c-channel.md
@@ -56,7 +56,7 @@ The fasteners and quantities are in the parts list above and are called out inli
 
 {% include step.html n="1" title="Preparation" %}
 
-**No heat inserts on this assembly.** Every screw here threads straight into a printed part, except the three that go into the stepper's own tapped holes. That is how one was built, and the part STLs agree: the rotor's six holes are Ø2.8 mm and the stator's are Ø2.4 and Ø2.8 mm, all thread-forming, while the NEMA bracket and the output gear are Ø3.4 mm clearance throughout with nothing threaded in them.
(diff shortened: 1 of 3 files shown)


basically-balloon:
**barthel** rejected this in Discord.

That is 0 of 2 approvals, with 1 rejection(s). A rejection stops the automatic merge: this one needs a person now.

PR #381, opened 2026-08-22
parts-calculator: what to print, as a count, and a manifest that says so

spencerhhubert:
Stacked on #379. Structural option **B**: the assembly owns the relationship through its `lines`, the part stores nothing, and membership is derived.

## The bug, measured

The root list grouped parts by a single field on the part naming its assembly. **38 of 98 parts were filed wrong:**

| count | problem |
|---|---|
| 20 | loose at the bottom of a section, though the tree knows where they go (the stator, the NEMA bracket, the whole c-channel drive) |
| 10 | hidden inside a folder, so their assembly never showed |
| 4 | pointing at an assembly whose lines never mention them |
| 4 | used in two places, able to show in only one |

Grouping now comes from `lines`. `C-channel drive` went from an empty idea to holding Stator, Output gear, NEMA bracket, Idler gear and Input gear.

## Where the numbers come from

The tree cannot supply quantity: it reaches 73 of the 89 build parts and undercounts 5 more (`frame-90deg-bracket` is in both the `layer` and `interface` sections; the tree has only the layer copies). So `$lib/uses` reconciles them — **tree says where, sections say how many**, and whatever the tree cannot account for becomes one visible "not in the assembly tree yet" use.

The invariant the UI leans on: **a part's uses always sum to exactly what the machine needs.** Nothing can hide in the gap, and the gap is on screen rather than silent.

## Selection is a count now

A checkbox cannot say "one of the two fan brackets". A part defaults to what the machine needs and can be edited to anything including 0, and only *edited* counts persist — so a new part in the catalog needs no migration, and a v1 config's `false` folds into a `0`.

`buyList` takes a quantity instead of a boolean, with colour splits scaled by largest-remainder so an edited quantity never produces a fractional part.

## Two views

- **By assembly** — what bolts to what. Repeats are fine here per your call; a shared part is listed under the first assembly and carries a `used in 2 places` marker naming the others on hover.
- **By unique part** — every part once, editable quantity, reset icon back to the machine's number, running totals in the footer.

## The manifest

A zip of forty files named `<part>-<uid>-<hash8>.stl` says nothing about how many to run. `print-manifest.txt` now rides inside the zip and downloads on its own:

```
 QTY  PART                           UID      EACH     TOTAL  COLOR     FILE
────────────────────────────────────────────────────────────────────────────
   1  40 mm Fan Bracket              3ufy     27 g      27 g  Charcoal  fan-bracket-40mm-3ufy-d4757b35.stl
   4  NEMA bracket                   x09t     72 g     290 g  Charcoal  nema-bracket-x09t-8b2bea87.stl

4 parts · 10 pieces · 1.05 kg of filament

Edited quantities (the machine calls for a different number):
  40 mm Fan Bracket: printing 1, machine needs 2

Used in more than one place:
  40 mm Fan Bracket: Control board mount, Orange Pi mount
```

The uid on each line is the exact design revision, so a manifest kept beside a finished machine still says what went into it.

## Verified in a browser

Set the fan bracket to 1: its total went 55 g → 27 g, the footer went 361 → 360 pieces and 14.37 → 14.35 kg, and the reset offered "Back to 2, what the machine needs". 89 quantity boxes, one per build part, with the 9 candidate-only housing parts correctly absent. svelte-check 0 errors, production build clean.

## Not done

Per-use checkboxes in the by-assembly view. The quantity box covers the same ground and the checkbox stays a plain on/off, which is where I stopped.

### DIFF
--- a/parts-calculator/src/lib/config.ts
+++ b/parts-calculator/src/lib/config.ts
@@ -13,7 +13,12 @@ export const CONFIG_KEY = 'sorter-filament-config-v1';
 export type StoredConfig = {
 	printBins: boolean;
 	surplus: number;
-	selected: Record<string, boolean>;
+	/** Parts whose printed count has been edited away from what the machine
+	 *  needs, id -> count (0 = not printing it). Absent means "follow the
+	 *  machine", so a new part in the catalog needs no migration. */
+	qty: Record<string, number>;
+	/** v1's per-part checkbox. Read once and folded into `qty`, never written. */
+	selected?: Record<string, boolean>;
 	inclSupport: Record<string, boolean>;
 };
 
--- a/parts-calculator/src/lib/filament.ts
+++ b/parts-calculator/src/lib/filament.ts
@@ -808,10 +808,35 @@ export function effectiveGrams(part: Part, inclSupport: boolean): number {
 /** Group the SELECTED parts' filament by resolved color, with bulk-tier pricing.
  *  `inclSupport(id)` decides whether a part's support material counts.
  *  `surplusPct` adds a buffer (incidental parts / failed prints) before spool counts. */
+/** Split `total` across `weights` as whole units, keeping the proportions and
+ *  summing to exactly `total` (largest remainder). Used when a part's printed
+ *  quantity has been edited away from what the machine needs and its colours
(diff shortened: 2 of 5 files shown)


spencerhhubert:

PR #385, opened 2026-08-22
parts-calculator: one way to search, and ⌘K to reach it

spencerhhubert:
The header carried a 4-character box that did one thing: type an id off a print, land on `/u/<uid>`. It could not find a part by name, it 404'd on a typo, and the three long lists on the site had no filter at all.

## One matcher backs everything

`$lib/search.ts` splits in two:

- a **scorer** that knows nothing about the catalog — name, uid, slug, keywords, prose in; a score plus the character ranges that matched out;
- an **index** of every thing the catalog names: parts, their superseded versions, their candidates, assemblies and theirs, hardware, laser-cut sheets, framing pieces, sections, pages.

So the palette, the parts list, the hardware list and the assembly tree all rank identically — a part found one way is found the other way.

## The ranking is the point

A **complete** uid outranks every other kind of match, because four characters read off a print are unambiguous. A **partial** uid sits just below a name starting with the same letters. So:

- `s` → **Stator** first
- `s2` → one row: the Interface bracket's superseded v1, `S2O9`

Neither is special-cased — it falls out of the weight table. Under 3 characters only prefixes match, so `a` doesn't return the whole catalog. `m3x8` finds `M3 × 8`; `cchannel` finds `C-channel`.

## The palette

⌘K / Ctrl-K / `/` over any page. Scopes: Everything · Parts · Assemblies · Hardware · **Stamped id**; `#` or `id:` forces id-only from the keyboard; Tab cycles; recents when empty.

Kind is carried by an icon and a chip whose colour means **state** — inked for current, grey for history, amber for under test — so a part and a rejected test print of it, which share a name, never read alike. Renders left, uid hard right in mono, so you can scan the column while reading an id off a print.

## Filters on the long lists

`SearchField` at the top of the parts, hardware and assembly lists.

- **Parts** — both views. Groups auto-expand so a match can't hide inside a collapsed folder. Totals stay the whole build, labelled as such, rather than quietly restating what you're about to print.
- **Hardware** — `m3x8` → 2/69.
- **Assembly** — a real tree filter: the path down to a match stays on screen, and an assembly matching by name keeps its whole subtree. `m3 nut` → 2 matches across 6 nodes.

Each offers "Search everything for X →", so a narrow list is never a dead end.

Skipped laser cut (3 parts), framing (10 rows) and changes (15) — a filter there is furniture.

## Also

A typo now says the id carries nothing and to check for a 0/O mix-up, instead of 404ing.

Removed the "Grouped by what bolts to what." hint next to the view toggle — both halves of it, since the toggle buttons already carry those exact sentences as tooltips.

## Checks

`svelte-check` 0 errors, production build clean, 213 `/u/<uid>` pages still prerendered. Cherry-picked onto current `main` with no conflicts.

**Not merged yet** — I'm chasing one interaction bug found after the branch was cut: after the palette is closed with Escape, the ⌘K chord doesn't reopen it (the trigger button still does). Fix incoming before this lands.

### DIFF
--- a/parts-calculator/src/lib/components/search/Highlight.svelte
+++ b/parts-calculator/src/lib/components/search/Highlight.svelte
@@ -1,0 +1,24 @@
+<script lang="ts">
+	import type { Range } from '$lib/search';
+
+	// Text with the matched characters marked. The matcher hands back a range
+	// rather than a substring so the highlight lands on the actual characters
+	// that matched, not on every other place the same letters appear.
+	//
+	// Ranges are computed against the normalised (lowercased, accent-folded)
+	// form. For everything the catalog actually contains that is character-for-
+	// character the same length as the original, so the indices carry over; the
+	// bounds check below is what keeps an exotic decomposition from painting the
+	// wrong letters — it drops the highlight rather than misplacing it.
+	let { text, range = null }: { text: string; range?: Range | null } = $props();
+
+	const parts = $derived.by(() => {
+		if (!range) return null;
+		const [a, b] = range;
+		if (a < 0 || b > text.length || a >= b) return null;
+		return { before: text.slice(0, a), hit: text.slice(a, b), after: text.slice(b) };
+	});
+</script>
+
+{#if parts}{parts.before}<mark class="bg-primary/[0.14] font-semibold text-text">{parts.hit}</mark
+	>{parts.after}{:else}{text}{/if}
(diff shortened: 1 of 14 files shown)


basically-balloon:

basically-balloon:
This has the approvals to merge, but the branch does not merge cleanly into `main` right now. Rebase it and the board will pick it up again.

basically-balloon:

basically-balloon:

PR #386, opened 2026-08-23
parts-calculator: P0 for the output gear fastener, and one collapsed notice per part

basically-balloon:
Requested by Spencer (@spencerhubert) [from Discord](https://discord.com/channels/1430279849171222682/1540821273972572240/1540821273972572240): "i dont think updating in the docs is the right way to triage issues with parts. can you instead just mark this as a p0 thing under "changes" . additionally whereever we render a part with a problem (broken feature) we should put a tiny warning box... perhaps should collapse those into the same thing for n number notices on a part."

Reported by aleph1111/christoph and scres from a build.

## The P0

The six screws holding the Output gear (130T) to the rotor were specified `scr-m3-8-cs`. Measured off the pinned `output-gear` and `rotor-faceted` STLs, which share assembly coordinates:

- Output gear spans z −11.00 to −3.00 — **8.00 mm thick at the bolt circle**.
- Its six holes are Ø3.40 clearance with a 90° countersink on the underside (Ø6.50 mouth, 1.55 mm deep).
- The rotor's six Ø2.80 thread-forming holes run z −3.00 to +2.00, **5.00 mm of flange**, coaxial at R 89.98.

A countersunk screw's length is measured over the head, so an M3 × 8 seated flush ends at z −3.00: level with the gear's top face, zero thread in the rotor. M3 × 12 ends at z +1.00 — 4.0 mm of engagement, 1 mm short of breaking through. Whether the joint settles as a 12 mm screw or the gear boss changes is left open in the entry.

## The notices

`ChangeStatus` already put a popover badge next to a name on parts, hardware, laser-cut parts, assemblies and sections. Two gaps:

- **One badge per change.** A part with three notices got three badges. Now one control: tone and priority from the worst notice, panel lists them all, count in the label when there is more than one.
- **Nothing on the image.** New `variant="marker"` renders a small corner mark on the image tiles — both parts-list renders, the hardware photo, the laser-cut outline — with the same panel. Quiet on purpose (warning tone, translucent chip); the alarm lives in the panel, which says what it is, what the issue is, and links `/changes#<id>`.

## Also

`targetHref` for an assembly on the changes page now points at `/assembly?focus=<id>` instead of `/#assembly-<id>`, the same destination the hardware page already uses for "where does this go".

To be straight about how this one surfaced: on the older base I built against, `/#assembly-c-channel` did not exist — that anchor only appeared when some part carried the assembly in its own `assembly` field, and c-channel's members live only in its `lines` — so the prerender failed the build on a dead link. #383 changed the grouping and the anchor exists now, so this is no longer a fix, just the better target: the explorer opens focused on the assembly rather than scrolling the dashboard to a collapsed row. Happy to drop it if you'd rather keep the anchor.

## Checks

`npm run check` and `npm run build` clean, `check_generated_pins.py` agrees.

**Generated data was refreshed with `generate.py --metadata-only`**, which carries `stamped`, grams and render URLs over from the committed file and only rewrites authored fields. The diff on `catalog.generated.json` is the 20-line changes entry and nothing else. A full run is not reproducible off Spencer's machine — it rewrote 284 stamped-STL URLs, 12 normals, 10 face picks and 9 renders with no source change.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -245,6 +245,26 @@
           "interface-spacer"
         ]
       }
+    },
+    {
+      "id": "fix-output-gear-rotor-fastener",
+      "name": "Output gear to rotor screw does not reach the rotor",
+      "priority": "P0",
+      "condition": "broken",
+      "description": "The six screws holding the Output gear (130T) to the rotor were specified as M3 x 8 mm countersunk. The gear is 8.00 mm thick at the bolt circle and a countersunk screw's length is measured over its head, so an M3 x 8 seated flush in the countersink ends level with the gear's top face and takes no thread in the rotor at all. Fit M3 x 12 mm countersunk for now: 4.0 mm of thread in the rotor's 5.0 mm flange, 1 mm short of breaking through the far side. Whether the joint settles as a 12 mm screw or the gear boss changes is still to be decided, so treat the length as provisional. Reported from a build by aleph1111/christoph and scres; measured off the pinned output-gear and rotor STLs.",
+      "targets": {
+        "parts": [
+          "output-gear",
+          "rotor-faceted",
+          "rotor-finned"
+        ],
+        "assemblies": [
+          "c-channel"
+        ],
+        "hardware": [
+          "scr-m3-12-cs"
+        ]
+      }
     }
   ],
   "folders": [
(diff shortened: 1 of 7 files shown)


basically-balloon:

basically-balloon:

That is 2 of 2 approvals. This admin approval merges the approved commit if the pull request is open, not a draft, and conflict-free.

basically-balloon:
**Merged.** Approved in Discord by allowlisted admin barthel, Spencer, so the board merged the approved commit.

PR #387, opened 2026-08-23
The control board housing becomes the mount, and two docs pages for it

spencerhhubert:
Two commits, one for the catalog and one for the docs.

## 1. `ctrl-board-mount` v2 — the housing replaces the bracket

The printed housing has been test-printed and built, so candidate `rns2` is promoted. It is now `ctrl-board-mount` v2 and the assembly takes the candidate's name, **basically Embedded Control Board Housing**.

- The board bolts straight to the base on four of its eight heat inserts and the cover closes onto the other four, so the v1 bracket, its four 10 mm standoffs and the cantilevered fan bracket are all gone. The cover carries the 40 mm fan itself.
- v1 is kept as a `versions` entry pinned to uid `u12m`, with its lines snapshotted at each member's uid of the day, and its Onshape render moved onto that entry.
- Fasteners are the measured ones now, not placeholders: 4 × M3 × 6 button head for the board (was "length TBD"), 4 × M3 × 12 button head self-tapping for the fan (was 16 mm, in both the cover's and the fan's descriptions). `scr-m3-shcs-tbd` has no caller left; its uid stays because it was minted, and its text says so.
- Quantities follow: the four housing parts enter the electronics section at 1 each, `ctrl-board-extrusion-mount` drops to none, and `fan-bracket-40mm` goes 2 → 1 because only the Orange Pi mount still uses it.
- All four housing parts got the Onshape version link they were missing.

`check_generated_pins.py` and `check_bucket_urls.py` both pass; the rebuilt bundles are uploaded.

## 2. Two docs pages, written from a real build

The old **Control board mount** page documented the v1 bracket and is deleted. In its place:

**[Preparing the control board](docs/src/content/hardware/electronics/installation/control-board-prep.md)** — seating the five TMC2209s and the Pico, and the ten jumpers that address the drivers. Why they exist gets three sentences (shared single-wire UART, MS1/MS2 are the address bits, four per bus is one short of five steppers, so two buses), then the address map as a table taken from the firmware's `hwcfg_basically_v1_2.h`, an annotated photo of the two jumper states, and a second annotated photo of the one real trap: the fifth driver's header block is mirrored, MS2 on the left and pin 1 at the top, which is why the board prints **!Double check!** there. The RESET button gets two sentences from the schematic: it pulls the Pico's RUN pin low, and it is not BOOTSEL.

**[Control board housing](docs/src/content/hardware/electronics/installation/control-board-housing.md)** — the eight screws into the base, the fan self-tapped into the cover, the plunger and its retainer over the reset switch, closing it up, and bolting it to the extrusion. The fan step names the port and why it is a port and not a permanent feed (24 V low-side switched by a MOSFET on GPIO1 or GPIO6), and warns to bridge that port's bypass jumper, since each one feeds +V through a 180 Ω resistor sized for an LED strip.

Inbound links, `nav.yml` and the installation index are updated.

## One thing to check

In the build photos the fan is plugged into `LED_1_2` and **JP4, the bypass jumper for that port, is not bridged** — so the fan is running through R28's 180 Ω. The netlist is unambiguous (`R28` sits between `+24V` and `J11` pin 1, `JP4` shorts it), so the page says to bridge it. Worth confirming against the real fan before this is taken as gospel.


basically-balloon:

basically-balloon:

PR #388, opened 2026-08-23
docs: a part card opens the part, the way the calculator does

basically-balloon:
Every card under **Parts needed** is a control now. Clicking the image or the name opens a modal with what the parts calculator's own part modal shows:

- the **3D model** for a printed part, orbit and zoom, CAD feature edges or bare shaded, reset view
- the **print facts**: version, updated, filament, print time, and the attributes the catalog carries
- any **open planned change** against the part, priority and all, so a P0 "broken" reaches somebody standing at the bench with the assembly page open
- what the part **takes** (heat inserts and the rest), resolved to names rather than ids
- **where to buy** an off-the-shelf one, and the STL/DXF download
- a link through to the part's page on the calculator for everything else: versions, candidates, build plates, the cart maths

Both sites read the same generated catalog, so the modal cannot drift from the calculator. `content.ts` carries the extra fields straight off `catalog.generated.json`, and the open changes use the same rule `plannedChangesFor()` applies (not complete, targets the id, or targets a section the printed part is built into).

**Notes on the implementation**

- `Modal.svelte` is a native `<dialog>`, so the backdrop, Escape and focus trapping are the browser's job. It is generic; the part detail is its only caller today.
- `StlViewer.svelte` imports `three` dynamically on mount, so a reader who never opens a part never downloads the library. It is a trimmed copy of the calculator's viewer, lit the same way (flat ambient plus a headlight on the camera, no tone mapping).
- `three` and `@types/three` added to `docs/package.json`, pinned to the same `^0.185` the calculator uses.
- The card image became a `<button>` carrying only the image; the corner badges stay siblings painted over it, so their popovers are still reachable and nothing nests a control inside a control.

`npm run build` and `npm run check` are both clean (0 errors; the 4 a11y warnings in `Requirements.svelte` are the pre-existing badge `tabindex`es on `main`).

Requested by Spencer (@spencerhubert) [from Discord](https://discord.com/channels/1430279849171222682/1540963547943804959/1540978622775697528): "make it so that the parts here get a similar modal to the one on the parts calculator".

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -13,6 +13,7 @@
 				"@sveltejs/vite-plugin-svelte": "^6.2.1",
 				"@tailwindcss/vite": "^4.1.17",
 				"@types/js-yaml": "^4.0.9",
+				"@types/three": "^0.185.4",
 				"js-yaml": "^4.1.0",
 				"liquidjs": "^10.21.0",
 				"rehype-raw": "^7.0.0",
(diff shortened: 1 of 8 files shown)


basically-balloon:
Spencer marked this ready for review in Discord.

basically-balloon:

basically-balloon:

PR #402, opened 2026-08-24
parts-calculator: P0, the stator's four bracket holes are under M3's root diameter

basically-balloon:
Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1540821273972572240/1541292837914677268): "Could you please create a CAD issue regarding the holes that are too small?"

Underlying defect reported by aleph1111/christoph, who could not screw the NEMA bracket to the stator as printed.

## The measurement

Off `stator-n6bm-323b2d1f`, the currently pinned STL (still v1, 2026-06-27, so nothing has changed since the report):

- **Four Ø2.40 mm × 10 mm thread-forming holes**, the ones the NEMA bracket bolts into. Fitted diameter is exactly 2.40 on all four.
- **M3's minor (thread-root) diameter is 2.459 mm.** The hole is already narrower than the root before print tolerance takes another 0.1–0.2 mm off it, and the screw has to form a full-depth thread through 10 mm of plastic.
- **Every other M3 thread-forming hole in this assembly is Ø2.80** — the rotor's six, the light post's five, and four more in this same stator. Ø2.40 is the odd one out inside its own part, which is what makes it read as a CAD slip rather than an intent.

Centres, in the shared assembly coordinates the bracket's four Ø3.40 clearance holes sit on: `(111.4, 5.7)`, `(-67.0, 89.2)`, `(-13.7, -110.9)`, `(-76.4, -81.5)`.

## The entry

`open-stator-bracket-holes`, **P0, broken feature**, targeting the stator, the C-channel drive assembly and the M3 × 12. Asks for the four to be opened to Ø2.80 in CAD, with drilling to 2.8 mm (7/64 in) as the builder's interim — 10 mm of engagement means nothing is lost.

**Worth knowing:** this was not tracked anywhere. #380 carried it as a docs warning plus an `info` note on the part, and that PR was closed rather than merged, so `stator` currently has no `info` and no changes entry. It is on the board now, which is where a part problem belongs.

## Checks

`npm run check` and `npm run build` clean, `check_generated_pins.py` agrees. Generated data refreshed with `generate.py --metadata-only`, so the diff is the 18-line entry in both files and nothing else.


basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #420, opened 2026-08-25
docs+parts: idler gear bearing retainers replace the loose washer

basically-balloon:
Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1541701360922140734/1541712000298786926): "Update the documentation accordingly. Remove the washer from the part calculator. Create a step in the preparation step for assembling the gear with the ball bearing and the printed washers/caps. Import the three renderings. Rotate the rendering in the first image so that the recess for the ball bearing in the gear is visible, and rotate the third image accordingly."

## What

The chute stepper idler gear's 608 2RS bearing was retained by a generic M3 x 15 mm washer under the M3x35 axle screw. The catalog already had two printed retainer parts for this joint (`chute-stepper-idler-bearing-retainer-outer`/`-inner`) sitting unused since 2026-08-17 — never wired into the assembly or the docs. This PR connects them:

- **`parts-calculator/catalog/parts.json`**: `chute-stepper-gearing` assembly drops the washer line, adds the outer retainer's 4 screws (`scr-m3-tbd` — head is countersunk, length isn't recorded anywhere) and the 4 heat inserts they need. The washer part itself stays in the catalog (unused, `sheet_qty` removed) rather than being deleted outright, so its uid stays resolvable. Regenerated `catalog.generated.json` with `--metadata-only`.
- **`docs/.../top-interface.md`**: new prep-item on the Preparation step (step 1) for building the gear + bearing + both retainer caps as a sub-assembly before it goes on the bracket, with three renders. Step 9's text updated to reference the now-prepared gear instead of the washer.

## The three renders

No assembly photo of this joint exists (it's a new-to-docs part), so these are reconstructions built from the part STLs — outer and inner retainer, the idler gear's own geometry, and the 608 bearing as a plain placeholder cylinder (no bearing STL in the catalog). All three are rotated to look up from underneath, so the bearing pocket recess and the outer retainer's four countersunk screw holes are visible, per the rotation ask.

## Not done here

The 4 countersunk screws' length isn't recorded anywhere (STL geometry gives head shape and a rough stack-up estimate, not a confirmed spec), so they're the `scr-m3-tbd` placeholder with a `fastener-todo` note, same convention as the rest of this page's unmeasured fasteners.

## Aside: a pre-existing bug in `catalog/generate.py`

`--metadata-only` currently throws `NameError: name 'rebased_render' is not defined` on main (confirmed on a clean checkout) — a leftover from the #418 asset-service refactor that removed the helper's definition but missed two of its three call sites. I patched it locally to produce the regenerated data in this PR, then reverted the source fix, since `catalog/generate.py` is flagged as not mine to open a PR against. Flagging here rather than silently working around it — worth a real fix from someone: the two dead calls are `v["render"] = rebased_render(...)` and `c["render"] = rebased_render(...)`, both should just become `= ov.get("render")` / `= oc.get("render")` the same way the third call site (right above them) was already fixed to `live_render = old["render"]`.

---

**Follow-up review (barthel):** restructured the idler-gear prep-item to match the NEMA 23 bracket / chute mount pattern, dropped the middle render, reordered the remaining two exploded-first, highlighted the bearing in blue, specified M3 x 8mm countersunk for the outer retainer's screws (same as the spur gear two lines up in this assembly), removed the washer references and the closing sentence per the review.

**Small follow-up:** moved the reconstruction-caveat sentence into the prep-item body (was a stray paragraph right after the div), and confirmed the screw/insert counts in the front matter (scr-m3-8-cs 5->9, hsi-m3 12->16) match the assembly lines.

**Another small follow-up:** barthel clarified the washer wasn't a design that got superseded, it was added to the catalog to match a documentation description that was itself wrong from the start. Reworded washer-m3-15's description to say that plainly rather than implying a legitimate prior design.

**Another small follow-up:** added a callout right under step 9's video saying it predates the bearing retainer caps and shows the old washer method instead. Re-checked step 9's prose against the current design while I was in there, it already matched.

**Another small follow-up:** moved the stale bearing photo from step 9 into the step 1 prep-item, with a caption noting it predates the retainer caps, and reworded the prep-item's caveat sentence to distinguish the real photo from the two STL reconstructions.

**Another small follow-up:** reordered the prep-item figures so the real photo sits between the two renders (exploded, photo, assembled) instead of first.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/top-interface.md
+++ b/docs/src/content/hardware/assembly/distribution/top-interface.md
@@ -413,6 +413,11 @@ Loosen the screws attaching the Limit switch housing to the extrusion. Slide the
     loading="lazy"></iframe>
 </div>
 
+<div class="callout callout-warning">
+  <span class="callout-icon" aria-hidden="true">⚠</span>
+  <p>This video predates the idler gear's bearing retainer caps. It shows the idler gear built the old way, with a loose washer under the M3 × 35 screw instead of the two printed caps from step 1.</p>
+</div>
+
 Slide the prepared [Timing pulley]({{ '/hardware/helpers/pulley-gear-mod/' | relative_url }}) over the shaft of the NEMA 23 stepper motor and tighten the two worm screws on its side. Slide the Interface spur gear onto the Timing pulley and secure it with five {% include fastener.html size="M3" variant="countersunk" length="8" %} screws, half tapped into the Interface spur gear and half bracing against the Timing pulley.
 
 Push the prepared Interface idler gear — bearing, inner retainer cap, and outer retainer already fitted, see step 1 — onto the Interface NEMA 23 bracket with the bearing facing outwards.
--- a/docs/src/content/hardware/assembly/distribution/top-interface.md
+++ b/docs/src/content/hardware/assembly/distribution/top-interface.md
@@ -40,6 +40,10 @@ parts_needed:
     qty: 1
   - part: interface-idler-gear
     qty: 1
+  - part: chute-stepper-idler-bearing-retainer-outer
+    qty: 1
+  - part: chute-stepper-idler-bearing-retainer-inner
+    qty: 1
   - part: interface-cage-bracket
     qty: 5
   - part: interface-cage-bracket-cable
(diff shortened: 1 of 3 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #421, opened 2026-08-25
docs: orange-pi-mount.md matches the catalog, bracket is no longer missing

basically-balloon:
Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1541701360922140734/1541754683230781481): "Yes. Use the style and layout from top interface." (following [this message](https://discord.com/channels/1430279849171222682/1541701360922140734/1541753223918911498) asking what changes I'd suggest, after I'd flagged the page as stale in [this thread](https://discord.com/channels/1430279849171222682/1541701360922140734/1541749689232269393)).

## What

`orange-pi-mount.md` still described the printed bracket (`orange-pi-extrusion-mount`) as missing from the parts registry. It hasn't been since 2026-08-22: it has a real STL, it's wired into the `orange-pi-mount` assembly, and it's counted in the electronics totals. The page just never caught up.

- **Front-matter `warning:`** — dropped the false "not in the parts registry" claim, kept the rest (AI-generated first draft, unchecked against a machine).
- **`parts_needed`** — added `orange-pi-extrusion-mount`, it wasn't listed at all even though it's required.
- **Removed the callout** right under the intro paragraph describing the (now resolved) blocker.
- **Replaced both "Image coming" placeholders** with real renders built from the STL, same technique as the idler gear bearing retainers in #420: an iso view for the step 1 prep-item, a straight-on view of the standoff face for step 2. Neither shows the Orange Pi 5 itself since it isn't a 3D part in the catalog, bracket only.

A third "Image coming" placeholder in step 3 (extrusion orientation) is untouched — that needs the frame context, not just this part's own geometry, and wasn't part of what was asked.

Style and layout match top-interface.md per the request: `prep-item`/`prep-item-body`/`prep-item-figure`, `fastener-todo` spans for the still-unmeasured screws, plain STL-reconstruction renders with a caption noting what they show.

---

**Follow-up review:** three more asks on the same page —
- Insert-hole rings were missing from the prep-step render, added (4, all visible from a straight top-down framing).
- Measured the M5 frame screw off the STL (5.5mm clearance, no countersink, 8mm-thick section) and swapped the placeholder for scr-m5-12-shcs (a stack-up estimate, labeled as such). The M3 board-retention screw genuinely can't be measured the same way, explained why in the text.
- Added a render of the bracket with 4 standoffs mounted, drawn as a parametric placeholder since neither the standoff nor the Orange Pi 5 has a 3D model in the catalog.

**One more small addition:** added the 4 board-retention screws (blue) on top of the standoffs in the render, same parametric-placeholder treatment, length illustrative only.

**Committed to a real spec for the last placeholder:** M3x6 socket/button head for the Pi retention screws, not measured but a well-evidenced call, PCB mounts are never countersunk and the control board's own board-to-standoff joint measured out to the same size for the same kind of connection.

**Rendering fix:** the prep-item render marked the insert holes with rings but the holes themselves were invisible under straight top-down orthographic lighting, no shadow, no depth cue. Retried at several camera tilts, picked one where the pocket wall actually catches shadow. Only 2 of 4 holes read cleanly from any single angle (the other 2 open away from whichever camera is used), so the image shows those 2 clearly and the caption says the other 2 mirror them.

**Text compacted:** removed the paragraph-length justification behind each inferred screw spec (M3x6, M5x12) and the render captions, the page banner already says nothing here is validated against a build, so the reasoning was redundant. States the spec and that it is not measured, nothing more.

---

**Correction, this one mattered:** the insert holes I'd been ringing (and mounting standoffs on) were the wrong holes entirely — a feature on the bracket's middle rail, not the actual M3 insert holes. Re-surveyed the STL from scratch and found the real 4 insert holes near the four corners of the frame, 95mm x 56mm apart, matching a normal SBC standoff footprint. Rebuilt both renders at the correct positions; all 4 are visible with real shadow in one shot now, since the corner positions aren't tucked behind the raised wall the old (wrong) positions were.

**More trimming:** removed six specific sentences/clauses (the control-board-precedent aside, the standoff-length caveat, the heatsink-clearance note, the parametric-placeholder trailing clause on the standoff caption, the M5 length-estimate caveat, and the fan-power aside in step 4), replacing one with a short factual line as asked. Nothing left but the steps and the facts.

**Rendering polish:** rotated both renders so the bracket lies on its long edge (landscape, not portrait), and dropped meshfig's built-in caption bar (the black box) by compositing the panel directly instead of calling Panel.finish(label) — the HTML figcaption already carries the text, the burned-in one was redundant and, per this comment, in the way.

**Layout:** floated the standoff render to the right margin with figure-float-right, step 2's text now wraps around it on the left instead of sitting as a full-width block below the paragraphs. Same pattern top-interface.md already uses for its corner-section drawing.

### DIFF
--- a/docs/src/content/hardware/electronics/installation/orange-pi-mount.md
+++ b/docs/src/content/hardware/electronics/installation/orange-pi-mount.md
@@ -12,11 +12,12 @@ contributors: [spencer]
 warning: >-
   **AI-generated first draft.** Written from the machine assembly tree in the [parts
   calculator](https://parts-calculator.basically.website/assembly?focus=orange-pi-mount), not
-  from an actual build. The parts and quantities are real, but the printed bracket the Pi stands
-  off is **not in the parts registry**, and no step here has been checked against a machine. The
-  steps below are placeholders with the gaps marked.
+  from an actual build. The parts and quantities are real, but no step here has been checked
+  against a machine. The steps below are placeholders with the gaps marked.
 parts_needed:
   - part: sbc-orange-pi-5
+    qty: 1
+  - part: orange-pi-extrusion-mount
     qty: 1
   - part: standoff-m3-10mm
     qty: 4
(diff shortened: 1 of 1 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #422, opened 2026-08-25
docs: drop the standalone External bracket page, fix top-interface BOM bug

basically-balloon:
The standalone External bracket page (`/hardware/assembly/distribution/external-bracket/`) described a build order (bottom-vertical flanged to the side bracket first, cover last) that doesn't match either real usage: `regular-layers.md` step 3 builds it cover-first, extrusion, bottom-vertical last; `top-interface.md`'s corner uses only the side + cover, no bottom-vertical at all (step 14). First round deleted the page; barthel asked for it back (see the follow-up commits below), so it now lives at `/hardware/parts/external-bracket/` instead, next to Lazy Susan, with its original four steps and photos intact.

While checking the original deletion, found the parts-calculator's `interface` assembly pulled in the shared `external-bracket` sub-assembly x6, which bundles a bottom-vertical + 4 screws per set — 6 bottom-verticals and 24 screws the top interface doesn't actually use (it needs 12 screws total, per step 13). The `ext-bracket-bottom-vertical` part's own catalog note had this flagged "interface qty ASSUMED 6, confirm" since 2026-06-27 and it was never resolved. Fixed the assembly line to the two real parts + correct screw count, corrected the same stale assumption on the affected parts' `quantities`/`docs_page`/notes, and regenerated `catalog.generated.json` (`generate.py --metadata-only`). That fix stands regardless of where the page itself lives.

Also worth flagging separately: `catalog/generate.py`'s `--metadata-only` path was broken on main (`NameError: rebased_render`, a dead call left over from #418's refactor that deleted the local helper but only updated one of its three call sites). Fixed it locally to regenerate this PR's JSON, then reverted the source change since `catalog/generate.py` isn't mine to PR against — flagging here so a person applies the two-line fix (drop the `rebased_render(...)` wrapper on the two remaining call sites, same as the third one already does with `old["render"]`/`ov.get("render")`/`oc.get("render")`).

**Follow-up commits, same thread:**
- Brought the original page back at the new Parts-section location, verbatim (four steps, all photos), after being told the first deletion went further than asked.
- Added a cross-sectional view under step 2, rendered from the actual STLs (side bracket + bottom-vertical tube in section, no next-layer parts, no baked-in text), then rebuilt it around the extrusion's real start/end and both screw positions per a follow-up ask, using the same measured numbers `regular-layers.md` step 5 already publishes.

Checked: `python3 docs/scripts/validate_frontmatter.py` and `validate_images.py` clean (same 5 pre-existing frontmatter errors as main), full `docs` build succeeds with no new broken links, `parts-calculator` `npm run check` is 0 errors and `npm run build` succeeds, `scripts/check_generated_pins.py` agrees.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #425, opened 2026-08-25
docs: light-post fastener details measured off the STLs

basically-balloon:
First of the feeder pages barthel asked to improve (light-post, camera-mount, classification-chamber — the ones still skeleton/AI-draft past C-channel).

**light-post** — measured off `light-post`, `light-post-cap-adapter`, `light-post-cap` (a shared STL frame):
- The three M3×12 screws join the cap adapter to the post only — the cap has no screw holes anywhere in its geometry, so it's a fastener-free fit over the assembly (and needs no heat insert).
- Post's three pilot holes: 2.8 mm self-tap, 120° apart, 12.1 mm radius. Adapter's matching holes: 3.4 mm clearance, same positions.
- One boss carries an asymmetric lobe the other two don't — a real orientation cue, even though nobody's confirmed which adapter feature it keys to yet.

**camera-mount** — `part6` (A/B connector) and `part37` (arm A) share a 12-hole bolt circle: 3.3 mm, 30° apart, 27 mm radius. Only 3 of the 12 are used at once, so the joint sets the arm's angle in 30° steps, not its reach like the draft guessed. `part37b` (arm B) doesn't share this circle at all, so how it joins is still open.

**classification-chamber** — went through two rounds after barthel supplied the camera module's own spec (M2 screws, 1.57 mm board), which didn't fit the pattern I'd first (wrongly) guessed was the camera mount:
- The Camera & LED insert's 4-hole 46 mm/3.4 mm pattern is M3, and matches `camera-extension-mount`'s own holes at the same spacing — that's the mount-ring joint, not the camera board.
- The camera board's real mount is on `camera-extension` itself: a two-post bracket, 2.6 mm holes, a real M2 fit. barthel confirmed the joint is self-tapping straight into an 8 mm-deep post. M2×8 mm socket/button head, worked from board thickness (1.57 mm) + most of the post's depth for a solid self-tap bite.
- M2 doesn't exist anywhere else in this catalog (everything else is M3/M4/M5/M6) — flagged plainly rather than inventing a catalog part id.
- The dome itself has zero screw holes anywhere in its own geometry.

All renders were originally flat top-down outlines; barthel flagged them as looking strange, so all three got redone at a slight angle with `parts-calculator/scripts/meshfig.py` (matching the chute-core insert views' style). Doing the classification-chamber one properly at an angle is what surfaced the insert's 10-hole outer ring in the first place — a narrower area filter on the first pass had missed it entirely.

Checked at each round: `validate_frontmatter.py` and `validate_images.py` clean (same 5 pre-existing errors as main), full `docs` build succeeds.

### DIFF
--- a/docs/src/content/hardware/assembly/feeder/camera-mount.md
+++ b/docs/src/content/hardware/assembly/feeder/camera-mount.md
@@ -11,8 +11,8 @@ author: barthel
 warning: >-
   **AI-generated first draft.** Written from the machine assembly tree in the [parts
   calculator](https://parts-calculator.basically.website/assembly?focus=camera-mount), not from an
-  actual build. The parts are recorded; the assembly order, every screw's position, the camera fixing
-  and all photographs are missing. Fill it in as you build.
+  actual build. The A/B connector's joint is now measured off the STLs. The rod mounts, the camera
+  fixing and every photograph are still missing. Fill it in as you build.
 parts_needed:
   - part: camera-mount-part-6
     qty: 1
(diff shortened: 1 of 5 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #406, opened 2026-08-24
docs(chute-core): mark the 18 heat-insert locations in Preparation

basically-balloon:
Puts three rendered views of the chute core into the **Preparation** step, in place of the `Image coming` placeholder, marking where all 18 M3 heat inserts go.

One long side, 8 inserts

### What is in it

- **8 + 6 + 4 across three faces**, marked on one view each: the two long sides and the rear panel. Every pocket position was measured off `chute-core`'s STL rather than taken from the assembly tree, and they are all the same hole: Ø4.20 mm, blind, 5.70 mm deep, with 11.3 mm of solid plastic behind and no relief hole.
- Each view is turned about 18° off its face. Straight on, a blind flat-bottomed pocket shades identically to the plate around it and is invisible; the tilt is what puts enough of the wall in shadow for it to read as a hole. Only a thin circle marks each one.
- Each pocket is ray-tested per view, so a circle is only drawn where nothing is standing in front of it.
- Layout follows the **Top interface chute mount** prep item on `top-interface.md`: one `prep-item` whose figure column is a `prep-item-figure-split` holding a `figure` per view, each with its own caption.

The renders come from `parts-calculator/scripts/meshfig.py` against the published `chute-core` STL, and are hosted on the asset service. They are in the `sorter-parts` namespace rather than `sorter-docs` because `basically-balloon`'s token has no `write:sorter-docs` scope. That, and the ~75 minute rendition stall that held this PR on placeholder URLs, are tracked in #408.

### Checks

- `npm run build` in `docs/` on Node 20, clean.
- `docs/scripts/validate_images.py`: 152 assets, passes, including these three.
- `docs/scripts/validate_frontmatter.py`: the 5 errors already on `main`, no new ones.

Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1541303416951804014): "In the Docs, there is a Chute Core page where the 18 heat inserts are mentioned in the “Preparation” section. Please create views from the Chute Core STL file that show the locations where the heat inserts are to be inserted, and mark them. You can create multiple images. It’s important that it’s easy to see where the heat inserts need to be placed. Add the rendered images to the “Preparation” section of Chute Core."

PR #433, opened 2026-08-26
docs: foot cover on bottom two layers takes 2 M5x16 screws

basically-balloon:
Resolves the "fastener not recorded" gap on the bottom two layers page: whether the External bracket - foot cover takes screws or clips on.

- The foot cover takes 2 M5x16 socket-button screws to the External bracket - side, the same joint as the bottom vertical it replaces, fitted before the extrusion goes in.
- This contradicts the page's earlier note that the foot cover "has no fastener holes of any kind" (apparently read off a geometry that doesn't match what's actually printed and screwed together).
- Updates the M5x16 tally in `parts_needed` from 84 to 96 (12 more, 2 per corner x 6 corners) and clarifies the step 3 corner build order: mount the cover to the bracket first, build the hex ring, partially thread the piece-D collar screws, insert piece D, measure from the bottom, then tighten.
- Adds BrickCycleAlice's four build photos (the joint with the cover mounted, measuring piece D, tightening the collar screw, and the two confirmed screws marked). My token doesn't have write access to the `sorter-docs` namespace on the asset service (tracked in sorter-v2#408), so these are uploaded to `sorter-parts` instead and referenced directly, the same workaround already used in #406/#416/#420.

Independently corroborated by uwe_barthel in the thread ("If I remember right are these two screws M5x16mm socket/button head. Same as used in regular layer to mount two layer together"), and BrickCycleAlice confirmed directly ("Definitely takes the two screws").

Note from BrickCycleAlice: she's still mid-build and plans a fuller rewrite of this page once she has more photos, so treat this as an interim factual fix, not the final word on the page.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1542096336613023875/1542099059232350278):** "Ok so this is the bottom foot cover attached to and external bracket, it is upside down. All 6 screws pictured are m5x16, Build order, Mount the foot cover to the external bracket, the top pictured screws. Create the hex ring, following regular layer to step 2, that's the second screw set to the al frame, still picture 1. Next Partially screw in the third set of bolts, Insert the foot extension, measure from the bottom, image two, then tighten onto the extrusion shaft, third image shows which bolt to tighten. (That's your step 3, part 4)"

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-two-layers.md
+++ b/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-two-layers.md
@@ -129,16 +129,16 @@ On each of the six corners, working on the second layer first:
 
 <div class="img-row">
   <figure>
-    <div class="img-placeholder">Image coming</div>
-    <figcaption>The bottom layer's corner, foot cover already mounted to the External bracket — side, then the bracket bolted into the hexagon extrusion as in regular layers step 2. Photo courtesy of BrickCycleAlice in the basically Discord; pending upload.</figcaption>
+    <img class="doc-figure" src="https://assets.basically.website/sorter-parts/img-4206-full-677bbaa8ee74.jpg" alt="The bottom layer's corner, seen upside down: the foot cover mounted to the External bracket — side, which is bolted into the hexagon extrusion">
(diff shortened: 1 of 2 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

Issue #436, opened 2026-08-26
Frame 90° bracket long-side holes too tight for an M5 to rotate freely


basically-balloon:
## Measured

Off the pinned STL (`frame-90deg-bracket-1081b927cfb5.stl`, via the asset
service, `trimesh` axis-agnostic hole finder): the bracket has exactly **two**
cylindrical holes, one on each side, not two on the long side as this issue
first said.

- **Long-side hole: Ø4.40 mm.** Takes the M5 × 12 that lets the bracket pivot
  against the A/G extrusion (via a T-nut) while the B/H spoke is fitted
  (`regular-layers.md` step 2). M5 clearance elsewhere in this frame measures
  ~5.5 mm, so 4.40 mm is far too tight to rotate freely as printed. This is
  the clear defect: the hole needs to be a real M5 clearance fit.
- **Short-side hole: Ø4.40 mm.** Takes the M5 × 16 that's meant to tap
  directly into the bracket, bracing against the extrusion (same step) --
  and 4.40 mm matches the M5 self-tap diameter already used elsewhere in this
  same frame (e.g. `ls-bottom-static`'s self-tap holes), so by the frame's own
  convention this one is sized correctly, not broken.

BrickCycleAlice's build needed **both** drilled out, though -- she corrected
an earlier version of this issue that (wrongly) described two long-side
holes. Worth checking while this is open whether the short-side hole is
actually printing tighter than 4.40 mm in practice (tolerance loss) even
though it's specified correctly, rather than assuming the self-tap side is
fine because the number matches convention.

144 holes across a full machine either way: 2 per bracket × 12 brackets per
layer × 6 layers.

Reported by BrickCycleAlice from a build.

## Design intent: this is a T-nut joint, not self-tapping

barthel asked whether this bracket was ever meant to be self-tapping with no
T-nuts at all. Checked the server history: nothing supports that for the
whole bracket. What's there instead is aleph1111/christoph hitting this exact
problem on 2026-08-01 in #machine-setup-help, and zed0 (who wrote most of the
current assembly docs) answering:

> "I have no idea what the official expectation is here, but personally I
> used M4s on both holes. On the short side I tapped into the plastic,
> bracing it against the extrusion without a T-nut. On the long side I filed
> off a bit on the inside to let the M4 slide through into a T-nut."

Two builders, three weeks apart, independently landed on "T-nut on the long
side, file/drill it to fit" -- only the short side is self-tapping in either
[...]

Issue #423, opened 2026-08-25
catalog/generate.py --metadata-only: NameError on rebased_render, undefined since #418


basically-balloon:
\`catalog/generate.py --metadata-only\` throws \`NameError: name 'rebased_render' is not defined\` on a clean \`main\` checkout, at two call sites (currently around lines 995 and 1005, inside \`main()\`'s metadata-refresh branch).

**Cause:** #418 (2026-08-24, the asset-service migration) replaced one of three uses of a locally-defined \`rebased_render()\` helper with \`old["render"]\` directly, and deleted the helper's \`def\` — but missed the other two uses a few lines further down in the same function.

**Fix** (same shape as the one #418 already applied to the first call site):

\`\`\`python
# both of these:
v["render"] = rebased_render(ov.get("render"), f"{source['id']}-v{v.get('version')}")
c["render"] = rebased_render(oc.get("render"), f"{source['id']}-{c['uid']}")
# become:
v["render"] = ov.get("render")
c["render"] = oc.get("render")
\`\`\`

Not something I can PR directly (\`catalog/generate.py\` is one of the files a PR against it "only creates work," per the \`github\` skill's parts-calculator section), so flagging here instead. Hit this independently on two unrelated PRs so far while trying to regenerate \`catalog.generated.json\`: #420 and #422, both worked around it by patching locally, regenerating, then reverting the source file before committing.

PR #437, opened 2026-08-26
parts-calculator: mark Frame 90° bracket broken (M5 holes too tight)

basically-balloon:
Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1487518366020276397/1542144753641066616): "mark the 'Frame 90deg bracket' as broken in the parts calculator. Step 2, in regular layers just informed me to drill the holes larger to allow clearance for an M5. That means 2x12x6 drills. Or 144 holes that have been printed the wrong size."

Adds a P0 `broken` entry to `changes` in `catalog/parts.json` for `frame-90deg-bracket`: the long-side holes need M5 clearance to let the bracket pivot while the B/H spoke is fitted, but print too tight for that, so the regular-layers assembly doc already has builders drilling or stripping all 144 of them by hand. Tracked as sorter-v2 issue #436, which also carries BrickCycleAlice's separate note that the bracket could stand to be ~1mm thinner to ease T-nut install.

Regenerated `catalog.generated.json` with `generate.py --metadata-only`; `check_generated_pins.py` and `check_asset_urls.py` both pass. Hit the known `rebased_render` NameError again ([sorter-v2#423](https://github.com/basicallysource/sorter-v2/issues/423)), worked around it the same way as before: patched the two dead calls locally just long enough to regenerate, then `git checkout --` reverted `generate.py` before committing, so this diff doesn't touch it.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -324,6 +324,21 @@
           "scr-m5-12-shcs"
         ]
       }
+    },
+    {
+      "id": "widen-frame-90deg-bracket-holes",
+      "name": "Frame 90° bracket long-side holes too tight for an M5",
+      "priority": "P0",
+      "condition": "broken",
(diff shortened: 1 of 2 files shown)


basically-balloon:
Closing: Spencer confirmed 2026-08-26 that this hole is a deliberate fallback, not a defect — the bracket is designed to hold with no screw or T-nut at all ("This entire assembly has no hardware," "Which was not actually broken, it was intentionally designed to avoid nuts"). See sorter-v2#443 for the corrected assembly method. Closing this PR without merging.

PR #474, opened 2026-08-30
docs: fix Chute review findings

basically-balloon:
Fixes the editorial review findings for the Chute landing page and its five sub-pages (chute core, door module, funnel, layer connectors, PCB/layer adapter board).

## What changed

- **Chute (index)**: added the sorting mechanism explanation (the whole chute stack rotates as one unit on the top interface's Lazy Susan; bins are fixed), deduplicated the "don't build chutes into layers" instruction, clarified the "This last part" reference, and noted when to connect each chute's ribbon cable.
- **Chute core**: clarified that the core's 18 inserts are separate from the bearing assembly's own 10 (previously easy to read as the same set), moved the inferred 12mm servo-bracket screw length next to where it's first used instead of burying it at the end of the section, fixed passive phrasing, and gave the placeholder image a caption.
- **Door module**: added the door's mechanism up front (bearings, servo, PCB timing), moved the "no assembly steps written for 3 of 4 sub-assemblies" disclosure before the fastener breakdown instead of after (previously the reader hit detailed fastener counts first, implying steps existed), fixed inconsistent capitalization of part names in running prose, noted the 6704-2RS bearings are sealed on both sides so they're symmetric (no orientation to get wrong), added a snug-don't-overtighten note for the bearing covers, explained what "clock the horn" actually means instead of relying entirely on the embedded video, and added a dry-fit check for the door's swing before final tightening.
- **Funnel**: stated plainly that funnel size and bin set are a matched pair (previously just said it "changes both"), gave the placeholder image a caption, and reworded the unrecorded funnel-attachment step to ask builders to share how theirs attaches.
- **Layer connectors**: gave a concrete pass/fail for "lines up" (flush, no visible gap or twist), tied the "don't fully tighten yet" warning directly to the fastening step instead of leaving it to a separate step with no cross-reference, added the N+2 total connector pairs for an N-layer machine, and gave the placeholder image a caption.
- **PCB → Layer adapter board**: renamed the page's title/kicker from "PCB" to "Layer adapter board" to match how every other page on the site refers to it (and updated the two cross-page link labels plus the nav.yml sidebar entry to match), added handling guidance for the bare board, gave a concrete answer for ribbon-cable timing (connect it before installing the chute into the frame if the harness is easier to reach on the bench), and gave the placeholder image a caption.

## Two findings checked and left alone

The review flagged two things that don't actually hold up once checked against the rendered site, so I didn't touch them:

- `funnel-half`/`funnel-third` already have no `qty` and a catalog `caption` instead ("For a layer of 12 half bins" etc.) — that's the established fix for an either/or part from sorter-v2#409, and adding `qty: 1` back (as the review suggested) would misrepresent that both are needed.
- `layer-connector-1`/`layer-connector-2` already render as "Layer connector A"/"Layer connector B" on the page, via the catalog's own `name` field (`Requirements.svelte` shows `part.name`, not the raw id), matching the body prose. No mismatch a reader would actually see.

Verified with a full local `npm run build` (Node 20, clean), `validate_frontmatter.py`/`validate_images.py` (no new errors), and grepped the built HTML for each changed string.

Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1543523992437137518): "Review: Chute / Collecting all review findings regarding Chute and the pages below from here [...] and let's start to work on those findings."

## Correction (barthel caught this)

The first version of this PR had the rotation mechanism backwards: it said the chute is stationary and a "bin ring" rotates underneath it. That's wrong, and it came from an unverified purpose/structure suggestion in the review rather than checking the existing pages. `top-interface.md` and `bottom-interface.md` both already describe the real mechanism: the whole chute stack rotates as one unit, stepper-driven through a gear train, on the Lazy Susan at the top interface; there's no separate "bin ring." Restored the original "rotating chute" lede and fixed the two mechanism sentences (chute index.md and door-module.md) that had it backwards.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/chute/chute-core.md
+++ b/docs/src/content/hardware/assembly/distribution/chute/chute-core.md
@@ -35,17 +35,17 @@ One chute is the chute core plus four things that bolt onto it:
 - **Door module**, one per chute: the door, the bearing assembly, the servo adapter and the servo in its bracket, built as a unit. See [Door module]({{ '/hardware/assembly/distribution/chute/door-module/' | relative_url }}).
 - **Funnel bracket (left)** and **Funnel bracket (right)**, one of each. See [Funnel]({{ '/hardware/assembly/distribution/chute/funnel/' | relative_url }}).
 - **Layer connector A** and **Layer connector B**, one of each. See [Layer connectors]({{ '/hardware/assembly/distribution/chute/layer-connectors/' | relative_url }}).
-- **Layer adapter board**, one per chute. See [PCB]({{ '/hardware/assembly/distribution/chute/pcb/' | relative_url }}).
+- **Layer adapter board**, one per chute. See [Layer adapter board]({{ '/hardware/assembly/distribution/chute/pcb/' | relative_url }}).
 
 Every screw on this page lands in one of the chute core's own M3 heat inserts. Nothing here taps into bare plastic.
 
 {% include step.html n="1" title="Preparation" %}
 
-Press the heat inserts into the chute core before anything is mounted to it. Once the funnel brackets and the door module are on, several of the insert positions are hard to reach with an iron. See [installing heat inserts]({{ '/hardware/helpers/heat-inserts/' | relative_url }}) for the technique.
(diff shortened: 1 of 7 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #511, opened 2026-09-01
install: stop attempting to copy example env that doesn't exist

zed0:
We were atttempting to copy $SOFTWARE_DIR/sorter/frontend/.env.example however that file was removed in 50e4ae94

### DIFF
--- a/software/install.sh
+++ b/software/install.sh
@@ -136,12 +136,6 @@ EOF
     ok ".env written"
 fi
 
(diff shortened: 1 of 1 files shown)


spencerhhubert:
Wouldnt be surprised if this script experienced other decay since it was made FYI @zed0 , I personally never used it

spencerhhubert (approved):

PR #514, opened 2026-09-01
docs: copy-link button beside every article heading

basically-balloon:
Every heading in an article gets a small chain-link button beside it that puts a direct link to that section on your clipboard.

**Requested by zed0 (@zed0) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1544368367333281844):** "on the assembly guide can we provide a small icon beside the headings which copies the link for that heading? If we're doing something similar elsewhere then use the same icon".

## The icon

The first version of this reused the site's existing code-block copy glyph. zed0 [corrected that](https://discord.com/channels/1430279849171222682/1544368367333281844/1544372050037313557): "The copy button on code blocks is a different function though. I think the usual icon for this function is a chain link". So it is a chain link, drawn inline to match the site's other icons, and `.copy-code-button` is untouched. The two share the `is-copied` flash (green, 1.5s) and nothing else.

## What it does

- Sits on `h2`, `h3` and `h4` inside the article body. `rehype-slug` already gives every heading an id, and the assembly guide's step headings carry their own (`step-1`, `step-2`, from `_includes/step.html`), so the anchor is the id that is already there. Nothing new is invented.
- Copies an absolute URL built from the current page (`origin + pathname + #id`), not from the address bar as it stands, so a link copied off a page you arrived at via a different anchor still points at the heading you clicked.
- Faint at rest, full opacity when the heading is hovered or the button is focused. Hidden in print.
- If the clipboard is unavailable (insecure context, or the browser refuses), it falls back to putting the anchor in the address bar so the link can still be copied by hand.
- A step heading is a column flex of "Step N" over the title, so the button is appended inside the title span rather than to the heading itself, which would have stacked it underneath as a third row.

## Scope

zed0 asked for the assembly guide. The article renderer is shared by every page on the site, so this lands site-wide, which is where a section link is useful anyway. Restricting it to `/hardware/assembly/` would have meant a conditional with nothing behind it.

`npm run check` is clean (0 errors; the 4 warnings are pre-existing in `Requirements.svelte`) and `npm run build` passes.

### DIFF
--- a/docs/src/routes/[...path]/+page.svelte
+++ b/docs/src/routes/[...path]/+page.svelte
@@ -24,6 +24,17 @@
 
 	let contentEl: HTMLElement | undefined = $state();
 
+	// One copy glyph for the whole article: the code-block button and the
+	// heading link button below are the same action, so they are the same icon.
+	const COPY_ICON =
+		'<svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="square" aria-hidden="true"><rect x="9" y="9" width="12" height="12"></rect><path d="M5 15H4a1 1 0 0 1-1-1V4a1 1 0 0 1 1-1h10a1 1 0 0 1 1 1v1"></path></svg>';
+
+	// Copied-state feedback, shared by both buttons.
+	function flashCopied(btn: HTMLElement) {
+		btn.classList.add('is-copied');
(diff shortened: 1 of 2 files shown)


basically-balloon:
zed0 marked this ready for review in Discord.

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #523, opened 2026-09-02
tests: catch 21 stale backend tests up with the code they cover

mneuhaus:
Tests only. On `main` the backend suite currently fails 21 tests that assert behaviour the code no longer has (the tests were never updated alongside the changes). This brings them in line; no production code touched.

Suite on `main`: 21 failed / 698 passed. With this PR: 719 passed.

### DIFF
--- a/software/sorter/backend/perception/tests/test_no_camera_crossover.py
+++ b/software/sorter/backend/perception/tests/test_no_camera_crossover.py
@@ -189,7 +189,10 @@ def test_three_workers_do_not_cross_outputs_when_run_concurrently() -> None:
         rad = math.radians(angle_deg)
         cx = 50.0 + 30.0 * math.cos(rad)
         cy = 50.0 + 30.0 * math.sin(rad)
-        return (int(cx - 5), int(cy - 5), int(cx + 5), int(cy + 5))
+        # The worker infers on the polygon's bounding-rect crop and offsets the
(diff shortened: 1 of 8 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07), triaged**

This PR only updates tests to match the code; Codex's findings are about the production code those tests now describe, not about the test changes themselves. None of them is introduced here, so the verdict for this PR is: mergeable as is. The findings are worth follow-ups of their own:

- `wait_until_stopped()` returns True at once for a software-disabled motor, and the manual-move timeout handler calls `_halt_stepper` without `force=True`, so a forced manual move on a disabled axis releases its lock without a reliable stop (`hardware/sorter_interface.py:379`, `server/routers/steppers.py:810`).
- rev01 discharge has no stop condition while `stopped` stays false: the legacy path waits forever, the perception path returns for `moving` before its total timeout is checked (`discharging.py:276`, `discharging.py:143`).
- Secrets seed first-use initialisation does SELECT then INSERT OR REPLACE; two concurrent first writers can end up with ciphertext nobody can decrypt (`secrets_crypto.py:49–73`). Insert-if-absent plus reread, or a process lock.
- Test-quality remarks kept for a later pass: the C4 direction test derives its expectation from the production `C4_TRAVEL_SIGN` (self-referential), and only the topology where `carousel` is a feeder-tracker role is covered in the vision-manager dynamic-detection test.

PR #524, opened 2026-09-02
stepper: ramp-aware move-time estimate so back-to-back pulses are not dropped

mneuhaus:
The firmware rejects a new distance move while the previous one is still running. The backend's move-time estimate ignored the acceleration ramps, so on the B1 machine 119 of 250 C3 pulses were refused ("Only allow new move when stopped"). The estimate is now a trapezoid (accel + cruise + decel) plus a margin; measured against the firmware on the machine.

Backend suite: no new failures (the 21 pre-existing failures on `main` are fixed in #523).

### DIFF
--- a/software/sorter/backend/hardware/sorter_interface.py
+++ b/software/sorter/backend/hardware/sorter_interface.py
@@ -5,6 +5,7 @@
 # Licensed under the MIT License. See LICENSE file in the project root for full license information.
 
 
+import math
 import os
 import time
 import json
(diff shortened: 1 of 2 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up**

- **No applied limit → the estimate used the caller's speed as the ceiling**, while the firmware cruises at its default 2000 µsteps/s until `set_speed_limits` is called (5000 steps: 1.75 s estimated vs ~2.7 s real). Fixed: the firmware default is the ceiling until a limit is applied.
- **Very short moves came out shorter than one step interval** at the start speed (49 ms for a step that cannot fire before 62.5 ms at min 16). Fixed with a floor of one step interval at v0, plus tests for both.
- **`min_speed = 0` freezes a firmware distance move** (`go_to_angle/flow.py:113` still sets 0): real, but pre-existing and outside this PR. It is the same freeze Spencer fixed for the pulse path in 06c9d3ce; go_to_angle needs the same floor of 16. Left for a follow-up.
- Codex could not run the suite in its sandbox; here: 6 estimate tests pass, the stepper endpoint safety and pulse tests are unchanged.

PR #525, opened 2026-09-02
firmware: software StallGuard for boards without a wired DIAG pin

mneuhaus:
The SKR Pico boards have no usable TMC2209 DIAG pins (the chute's would collide with its home sensor), so StallGuard never worked there. The firmware now polls TSTEP/SG_RESULT over the TMC UART on core0 and latches the same stall flag the DIAG path uses; the backend's stall monitor and incident path are unchanged.

Three commits, in order:
1. the poll itself, gated on the backend's SGTHRS/TCOOLTHRS writes;
2. a compact per-armed-channel debug view in `GET_OBSERVABILITY` (`soft_sg`), which is what made the next fix findable;
3. the fix: the two UART reads must not follow each other back to back (the TMC2209 needs the bus idle after its reply, the second read timed out on ~40 of 44 polls), and the poll samples only the cruise phase, because SG_RESULT reads near zero on ramps.

Verified on the B1 machine: the chute latched on a real jam (SG 10 at cruise) and on the mechanical stop; zero false trips over several hours of sorting. Observability payload stays under the 246-byte cap (worst case measured).

### DIFF
--- a/software/firmware/sorter_interface_firmware/Stepper.cpp
+++ b/software/firmware/sorter_interface_firmware/Stepper.cpp
@@ -32,7 +32,7 @@ Stepper::Stepper(int step_pin, int dir_pin)
       _mc_dir(1), _mc_home_pin(-1),
     _steps_moved(0), _steps_frac(0), _brake_distance(0),
     _current_speed(0), _current_speed_frac(0), _current_dir(1),
-    _jitter_active(false), _jitter_amplitude(0), _jitter_strokes_remaining(0), _jitter_dir(1) {
+    _jitter_active(false), _stall_request(false), _jitter_amplitude(0), _jitter_strokes_remaining(0), _jitter_dir(1) {
 }
 
 void Stepper::initialize() {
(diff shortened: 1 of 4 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up** — verdict was "needs rework"; two findings fixed in the new commit, three recorded as known limits:

- **Core-0 stall latch could be overwritten by core 1** (`latchStall()` stored `STOPPED` while core 1 was mid-transition, so the motor could resume after a detected stall until the 250 ms backend monitor intervened) — fixed: the poll only sets a request flag, core 1 consumes it at the top of its motion tick and performs the same stop the wired-DIAG path does.
- **Observability payload overflow** (every armed channel listed, a fifth entry pushed the reply over the 246-byte cap and the whole `GET_OBSERVABILITY` collapsed to `{}`) — fixed: only software-polled channels are listed, and an oversized reply drops the `soft_sg` detail rather than the hardware facts.
- **Moves that never reach a cruise phase are not polled** — true by design: SG_RESULT reads near zero during ramps (measured 4 on the B1 chute), so only the cruise phase is sampled. Short moves (e.g. one platter sector at 4000 µsteps/s with 2500 µsteps/s²) get no software protection. Documented limitation of the software emulation; a wired DIAG pin has no such gap.
- **Driver mode gating (TCOOLTHRS ≥ TSTEP > TPWMTHRS, StealthChop only)** — the poll checks the TCOOLTHRS floor only; if SpreadCycle is switched on for an armed channel the readings are not meaningful. Left as a follow-up: cache GCONF/TPWMTHRS and gate the poll on them.
- **SGTHRS = 0 treated as "not configured"** — noted; a separate "written" flag would be the fix, low priority.

Both SKR Pico variants build with the ARM toolchain. Not flashed yet: the B1 machine is down for a mechanical repair; the flash is part of the recommissioning.

PR #526, opened 2026-09-02
bins: misc can be routed into an explicitly assigned bin

mneuhaus:
Without a bucket under the chute the misc passthrough has nowhere to go. Positioning now honours a bin the operator assigned to `misc` (the assign endpoint no longer rejects it, and the bin modal offers "Misc (discard)"). Misc still never claims an empty bin on its own; unassigned it passes through as before.

Tests: the old `test_misc_assigned_bin_is_ignored_for_bottom_passthrough` encoded the previous behaviour and is replaced by two tests (routes to an assigned misc bin; never auto-claims an empty bin).

### DIFF
--- a/software/sorter/backend/server/routers/hardware.py
+++ b/software/sorter/backend/server/routers/hardware.py
@@ -3098,12 +3098,9 @@ def assign_bin_categories(
         category_id = category_id.strip()
         if not category_id or category_id in cleaned:
             continue
-        # MISC is the virtual discard passthrough, never a physical bin.
-        if category_id == MISC_CATEGORY:
-            raise HTTPException(
-                status_code=400,
-                detail="Misc is the discard passthrough and cannot be assigned to a bin.",
-            )
+        # MISC may be routed into a bin only by this explicit assignment; the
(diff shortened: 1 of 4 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07), triaged** — verdict "needs rework", but the three findings do not hold against what the PR changes, so: mergeable as is.

- **"A persisted `misc` tag changes routing after upgrade on any machine"** — before this PR the assignment endpoint stripped `misc` from every request, so no persisted `misc` tag can exist; routing only changes when an operator assigns a bin to misc explicitly in the bin picker. That is the feature.
- **"A full or unreachable misc bin releases pieces into nowhere"** — when the assigned bin is not usable, misc falls back to the passthrough to the discard bucket, which is exactly the behaviour every machine had before this PR for every misc piece. The PR adds no new path there.
- **"Not-in-inventory pieces do not reach the misc bin"** — by design and documented in the code: the not-in-inventory pool never mixes with the normal pool, so those pieces stay in their pool or pass through, as before.

Nothing to change; the review is recorded here for the history.

PR #527, opened 2026-09-02
chute: back out of the endstop flag before homing

mneuhaus:
The chute's endstop flag is wide (over 45° on the B1). When homing started with the sensor already active (after a stall or a hard-stop hit) the firmware zeroed the position on its first tick, wherever the chute stood inside the flag, and every bin was aimed up to 45° off — this looked like "homing lost instantly" after each re-home. Homing now probes forward in 22.5° steps until the sensor releases, then searches for the edge as before.

Verified on the machine: "endstop already active before homing — backing out … released after 1 probe" followed by a correct home.

### DIFF
--- a/software/sorter/backend/subsystems/distribution/chute.py
+++ b/software/sorter/backend/subsystems/distribution/chute.py
@@ -42,6 +42,11 @@ DEFAULT_FIRST_SECTION_OFFSET_DEG = 8.25
 
 HOME_SPEED_MICROSTEPS_PER_SEC = -1000
 HOME_TIMEOUT_MS = 15000
+# The endstop flag is wide (>45° chute on the B1). Homing must start outside
+# it, otherwise the firmware sees the sensor active on its first tick and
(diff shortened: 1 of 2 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up** — verdict was "needs rework"; addressed in the new commit:

- **Probe result ignored** (a probe that timed out left the chute possibly still moving while homing went on) — fixed: the result is checked; on a timeout the driver is cut so the unfinished move cannot continue and homing aborts. The next `home()` re-enables the driver.
- **A failed backout left `homed=True`** — fixed: the home reference is dropped when homing starts and only set again by a completed sequence.
- **Probes ran before the driver re-enable** that `StepperMotor.home()` does for older firmware — fixed: re-enabled before the first probe.
- **No tests** — added: sensor release after N probes, no-op on an inactive sensor, stuck sensor exhausting the budget, unfinished probe halting and clearing the home flag.

PR #528, opened 2026-09-02
chute: acceleration and operating speed actually reach the stepper

mneuhaus:
Two config gaps found while taming a skipping chute:

- `[stepper_acceleration_overrides]` (`<stepper> = <µsteps/s²>`, next to the current overrides) is applied where the default acceleration is set at stepper init. The chute used the firmware default of 10000 with no way to lower it from `machine.toml`.
- `[chute] operating_speed_microsteps_per_second` only shaped the move-time estimate; the stepper kept its 3000 µsteps/s init default, so the setting never changed how fast the chute moved. The chute now re-asserts its speed limit before every move (the stepper API endpoints set their own limits).

Verified on the machine: `set_speed_limits max=1500` logged per move, TSTEP matches the configured speed.

### DIFF
--- a/software/sorter/backend/irl/config.py
+++ b/software/sorter/backend/irl/config.py
@@ -954,8 +954,11 @@ def cameraDeviceSettingsToDict(
 def mkStepperConfig(
     default_steps_per_second: int = 2000,
     microsteps: int = 8,
+    acceleration_microsteps_per_second_sq: int = FIRMWARE_DEFAULT_ACCELERATION_MICROSTEPS_PER_SECOND_SQ,
 ) -> StepperConfig:
-    return StepperConfig(default_steps_per_second, microsteps)
+    return StepperConfig(
+        default_steps_per_second, microsteps, acceleration_microsteps_per_second_sq
+    )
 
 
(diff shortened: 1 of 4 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up** — verdict was "needs rework"; two findings fixed, two recorded:

- **Speeds 1–15 accepted but never applied** (`set_speed_limits(16, <16)` is rejected silently by the firmware, leaving the previous limit in force) — fixed: below 16 falls back to the default at parse time, and the chute clamps to the firmware minimum before every move.
- **Legacy physical stepper names ignored in `[stepper_acceleration_overrides]`** — fixed: normalised to the canonical names like the current overrides.
- **Post-home backoff timeout with a low acceleration** — the timeout comes from `estimateMoveDegreesMs`, which #524 makes ramp-aware (it honours the applied acceleration), and #527 drops the home reference when homing starts; with those two merged the scenario no longer leaves `homed=True` behind.
- **Chute-random StallGuard diagnostic loses its requested speed** — true, the diagnostic's limit is overwritten by the operating speed on every move; it is a calibration helper, not a production path, so left for a follow-up.
- Tests added: parse floor, clamp, legacy alias.

PR #529, opened 2026-09-02
c4: stall auto-clear shakes the platter when forward rotation cannot clear it

mneuhaus:
A tyre resting on the C4 fall-off lip rode the 720° auto-clear without leaving, and the channel stayed blocked (`exit_stuck`) until an operator intervened; a manual jitter freed it at once. After an exhausted rotation budget the two-piece auto-clear now walks the existing exit-release shimmy ladder from `classification_channel_config` (calm to firm, occupancy-checked after every stage) before giving up.

New unit tests for the ladder walk (clears mid-ladder, exhausts, skips when already clear / unconfigured).

### DIFF
--- a/software/sorter/backend/irl/config.py
+++ b/software/sorter/backend/irl/config.py
@@ -415,6 +415,16 @@ class ClassificationChannelConfig:
                 acceleration_microsteps_per_second_sq=6000,
                 settle_ms=450,
             ),
+            # The operator test wiggle that freed every lip-stuck piece on the
+            # B1 platter (2026-09-04/05) once the rungs above had failed.
+            ClassificationChannelExitReleaseStage(
+                "operator-proven-kick",
+                amplitude_output_deg=3.0,
+                cycles=3,
+                microsteps_per_second=16000,
+                acceleration_microsteps_per_second_sq=150000,
+                settle_ms=500,
+            ),
         )
         self.exit_release_review_pause_enabled = True
         self.stale_zone_timeout_s = 3.0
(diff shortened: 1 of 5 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up**

Findings and what was done:

- **Jitter timeout or telemetry error counted as finished** (`_waitJitterDone`) — fixed: the wait now returns whether the stepper positively reported the stroke finished; an unconfirmed stroke ends the ladder with `jitter_unconfirmed` instead of reading occupancy or starting the next stage while the platter may still move.
- **`force=True` bypassed `software_disabled`** — fixed: automatic recovery returns `stepper_disabled` before sending any command (a suppressed jitter otherwise returns True without moving, which would have looked like a shake).
- **The operator-proven last rung (3° ×3 @ 16000 µsteps/s) was missing from this branch** — added by cherry-picking 6c7101d5 from b1-belt-feeder; every auto-clear on 2026-09-04/05 exhausted the old ladder at 1.75° @ 1900 and the operator wiggle then freed the piece within a second.
- **Stroke shape**: Codex notes the firmware jitter primitive runs 0 → +A → 0 per cycle while the older exit-release shimmy plan in `running.py` alternates +A / −2A / +A. Left as is: the ladder deliberately uses the same primitive as the operator's "Test Wiggle" button, which is the manoeuvre that demonstrably cleared the stuck pieces.
- Tests added for the disabled-stepper and unconfirmed-jitter paths (5 pass).

History for reviewers: an always-on rotate-to-clear wiring of the two-piece path was reverted on main on 2026-06-29 (b9d3d972). This ladder is different in kind: bounded stages, run only while an exit-stuck incident is being auto-cleared, and it fails closed into the incident hold.

PR #530, opened 2026-09-02
steppers: AS5600 shaft-encoder position check as a second stall source

mneuhaus:
**Stacked on #525 and #528** (their commits are included; only the last commit is new here).

StallGuard only sees the cruise phase; a hit on a ramp or at the target goes unnoticed and the chute silently loses its home. An AS5600 on the motor's rear shaft (I2C0 on GP0/GP1, fixed address 0x36, one per board) lets the firmware compare the unwrapped encoder count with the commanded position every 5 ms; three consecutive readings beyond the tolerance latch the same stall flag StallGuard uses, so the incident / re-home path is shared. Position resets (homing, `SET_POSITION`) re-sync the offset; jitter strokes are skipped.

Backend: `[stepper_encoder.<name>]` (enabled / sign / counts_per_rev / tolerance_fullsteps) → `ENCODER_CONFIG` at stepper init; `ENCODER_STATUS` and the observability `enc` field expose magnet status, AGC and deviation for bring-up. Disabled unless configured. Firmware builds for both SKR variants; sensor bring-up on the machine is pending (hardware arrives tomorrow).

### DIFF
--- a/software/firmware/sorter_interface_firmware/AS5600.cpp
+++ b/software/firmware/sorter_interface_firmware/AS5600.cpp
@@ -1,0 +1,21 @@
+#include "AS5600.h"
+
+static const uint8_t REG_STATUS = 0x0B;
+static const uint8_t REG_RAW_ANGLE = 0x0C; // two bytes, MSB first, 12 bits used
+static const uint8_t REG_AGC = 0x1A;
+static const uint32_t XFER_TIMEOUT_US = 1000;
+
+bool AS5600::readRegisters(uint8_t reg, uint8_t *buf, size_t len) {
+    if (i2c_write_timeout_us(_port, _address, &reg, 1, true, XFER_TIMEOUT_US) != 1) return false;
+    return i2c_read_timeout_us(_port, _address, buf, len, false, XFER_TIMEOUT_US) == (int)len;
+}
+
+bool AS5600::readRawAngle(uint16_t *out) {
+    uint8_t b[2];
+    if (!readRegisters(REG_RAW_ANGLE, b, 2)) return false;
+    *out = (uint16_t)(((b[0] & 0x0F) << 8) | b[1]);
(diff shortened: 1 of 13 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07), triaged** — verdict "needs rework", and here the review is right at the design level, so this PR becomes a **draft** until the encoder is actually commissioned (it is shelved on the B1 machine for now: the AS5600 is wired but not enabled).

The branch was rebuilt on the updated #525 and #528 and now also carries the later re-sync fix from `b1-belt-feeder` (explicit position resets re-sync the encoder instead of guessing from jumps), which answers part of finding 4. Both SKR Pico variants build.

Open before this can leave draft, in Codex's order:

1. **I²C cross-core race** — AS5600 reads on core 0 and PCA9685 servo writes on core 1 share the peripheral without a lock; needs a transaction lock or a single owning core.
2. **A latched encoder stall does not inhibit motion** — `latchStall()` only sets STOPPED, the next `moveSteps()` is accepted, and blocking backend moves return True for any stopped state; completion must distinguish "target reached" from "fault-stopped".
3. **No incident path for encoder-only motors** — the stall monitor only watches StallGuard-armed steppers, and `DISABLE_STALLGUARD=1` silences the monitor but not the encoder.
4. **Small position resets** (below 2000 µsteps, or homing near zero) still look like shaft deviation — needs an explicit sync generation.
5. **Encoder config survives an MCU INIT** — removing the TOML entry and restarting the backend leaves the encoder armed.
6. **Fails open on sensor or magnet failure** — read failures only bump a counter; enabling should probe sensor health and have a fault policy.
7. **Several entries per board silently replace each other** — validate and apply once per physical board.

Plus tests for wrap/reset/latch, I²C errors, restart cleanup and per-board arbitration.

Issue #536, opened 2026-09-03
parts-calculator: delete the unused scr-m3-8-shcs catalog part


basically-balloon:
`scr-m3-8-shcs` ("M3 × 8 mm socket/button head") in `parts-calculator/catalog/parts.json` (uid `ql35`) is unused everywhere in the live assembly tree. It was added on 2026-08-21 as the screw clamping the C-channel input gear (12T) to the NEMA 17 shaft flat; the C-channel overhaul (version 3, 2026-09-01, `441d9114`) established that every screw on that assembly is countersunk, so that joint now carries `scr-m3-8-cs` and this row lost its only use. It has no `sheet_qty` either, so it contributes nothing to any BOM.

Its two remaining references are historical snapshots of superseded `c-channel` versions 1 and 2, which are expected and correct to leave alone.

I can't delete the part object myself: `parts-calculator`'s rule is that a minted uid stays resolvable forever, and `check_generated_pins.py` refuses a change that drops one from `parts.json` ("uid(s) removed from parts.json ... mark them superseded/rejected instead"). That is a CI-enforced invariant, not a call to make from a Discord thread. Same situation as #453 (`scr-m3-tbd`) and #454 (`scr-m3-shcs-tbd`).

So this needs someone who can decide whether to override or adjust that check for this uid, then delete the part object from `catalog/parts.json` and re-run `generate.py`.

In the meantime the part's `description`/`note` say it is retired and unused, and a P4 `changes` entry (`delete-unused-m3-8-shcs-screw`) puts a visible marker on it wherever the calculator renders it.

Requested by barthel (Uwe Barthel) (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1545150490940022864): "Mark and tag scr-m3-8-shcs in parts calculator like scr-m3-tbd to be retired and with P4 and create a github issue for deletion from parts catalog too."

basically-balloon:
Catalog markup (retired description, P4 `changes` entry) is in #537. This issue stays open for the actual deletion of the part object.

PR #479, opened 2026-08-30
parts-calculator: note that washer-m3-15 is unused

basically-balloon:
Requested by Uwe Barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1543649360905244752/1543650309958537376): "Create a new pull request for adding the same note on the washer in the catalog as we did for the unused m3 screws."

## What

`washer-m3-15` was already flagged unused in its `description` (from PR #420), but it didn't carry the `note` field or the "kept because a minted uid has to stay resolvable" explanation that `scr-m3-tbd` (#453) and `scr-m3-shcs-tbd` (#454) got. This adds both, same wording pattern as those two, and bumps `updated_at`.

Tracked for deletion in #478.

- `parts-calculator/catalog/parts.json`: `washer-m3-15` gets a `note` field and an added sentence in `description`.
- `parts-calculator/src/lib/data/catalog.generated.json`: regenerated with `catalog/generate.py --metadata-only`.

## Aside

`generate.py --metadata-only` still throws `NameError: name 'rebased_render' is not defined` on main, same pre-existing bug flagged in PR #420 (a leftover from the #418 asset-service refactor). Patched it locally to produce this regeneration, then reverted the source change, since `catalog/generate.py` isn't mine to open a PR against. Still worth a real fix from someone with access: the two dead calls are `v["render"] = rebased_render(...)` and `c["render"] = rebased_render(...)`, both should become `= ov.get("render")` / `= oc.get("render")` the way the third call site above them already was.

Issue #478, opened 2026-08-30
parts-calculator: delete the unused washer-m3-15 catalog part


basically-balloon:
`washer-m3-15` ("M3 × 15 mm washer", uid `phwe`) in `parts-calculator/catalog/parts.json` is unused everywhere in the catalog — its own `description` already says why: it was added to match the top-interface idler gear joint's documentation, which turned out to be wrong. The joint's actual design uses the printed chute stepper idler bearing retainers instead (PR #420, 2026-08-25), so the part was dropped from the `chute-stepper-gearing` assembly and nothing has referenced it since.

Same situation as `scr-m3-tbd` (#453) and `scr-m3-shcs-tbd` (#454): the part should be deleted outright, but `check_generated_pins.py` refuses a change that drops a minted uid from `parts.json`, so it needs someone who can override or adjust that check, then delete the object and re-run `generate.py`.

Requested by Uwe Barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1543649360905244752/1543650309958537376): "Create a github issue for the washer like for the unused m3 screws and tag it with bom."

basically-balloon:
### Discussed in Discord

▲ **1 upvote** from people in the Discord.

[Open the thread](https://discord.com/channels/1430279849171222682/1543650501034250301)


_This comment is written and rewritten in place by balloon as the thread continues. Reply here and it appears in Discord._

basically-balloon:
Tagged in the catalog as well: PR #537 adds a P4 `changes` entry (`delete-unused-m3-15-washer`, `targets.hardware`) so the notice renders on the part itself, and points its `description` here. This issue stays open for the actual deletion of the part object.

Issue #454, opened 2026-08-29
parts-calculator: delete the unused scr-m3-shcs-tbd catalog part


basically-balloon:
`scr-m3-shcs-tbd` ("M3 socket/button head, length TBD", uid `xe72`) in `parts-calculator/catalog/parts.json` is unused everywhere in the catalog — its own `description` already says why: it was the placeholder for the basically Embedded Control Board's screws before they were measured as M3 × 6 mm button head (`scr-m3-6-bhcs`), and nothing has referenced it since it was minted.

Same situation as `scr-m3-tbd` (#453): the part should be deleted outright, but `check_generated_pins.py` refuses a change that drops a minted uid from `parts.json`, so it needs someone who can override or adjust that check, then delete the object and re-run `generate.py`.

Requested by barthel (Uwe Barthel) (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1543277171865551012/1543290486939189299): "Bitte auch für die catalog Position "M3 socket/button head, length TBD" beides durchführen." ("Please do both for the catalog item "M3 socket/button head, length TBD" too.")

basically-balloon:
### Discussed in Discord

▲ **1 upvote** from people in the Discord.

[Open the thread](https://discord.com/channels/1430279849171222682/1543292162320638004)


_This comment is written and rewritten in place by balloon as the thread continues. Reply here and it appears in Discord._

PR #537, opened 2026-09-03
parts-calculator: mark the unused M3 x 8 screw and M3 x 15 washer for deletion, and badge a removal as one

basically-balloon:
barthel asked for `scr-m3-8-shcs` to get the same treatment as `scr-m3-tbd`: marked retired, tagged P4 in `changes`, and a GitHub issue for deleting it from the catalog. Then the same for `washer-m3-15`. Then, looking at the result, he pointed out that P4 alone is close to invisible and asked whether an icon would make more sense. It would, so the third commit adds one.

**`scr-m3-8-shcs` is genuinely unused.** No current assembly line and no connection edge; its only two references are the `c-channel` version 1 and version 2 historical snapshots, which are correct to leave alone. Spencer retired it at the bench in #515 (`441d9114`): the input gear's clamp screw is the countersunk `scr-m3-8-cs`. Every `M3 × 8` in `docs/` is the countersunk variant, and the Drive BOM sheet lists M3 countersunk in 8/12/16 mm with no M3 socket or button head at all.

**`washer-m3-15` is genuinely unused too.** PR #479 already gave it the retired `description` and `note`, and issue #478 already tracks the deletion, but it had no `changes` entry, so nothing rendered on the part itself. Its last assembly line went in PR #420 (2026-08-25) when the interface idler gear joint turned out to use the printed bearing retainers, not a washer.

**Why P4 alone was not enough.** `PriorityBadge` carries fixed hues for P0 to P3 only; P4 and up fall through to a generated hue, which is why it renders a pale teal nobody picked. Worse, with no `condition`, `ChangeStatus` classified these notices as nice-to-have and rendered "Nice to Improve · P4" with a refresh icon, so a part being deleted read as an optional improvement.

**What changed**

- New P4 `changes` entries `delete-unused-m3-8-shcs-screw` and `delete-unused-m3-15-washer`, both `targets.hardware`.
- Both parts' `description` fields point at their deletion issue; `scr-m3-8-shcs` also gains a `note` warning against attaching a new joint to it.
- `PlannedChange.condition` gains `'retired'`, for an entry whose subject is leaving the catalog rather than being fixed or improved. `ChangeStatus` gives it a trash-can icon and the words "To Be Removed" on the part badge and on the image-tile marker, and the changes page prints "To be removed" under the priority. All four deletion entries (`scr-m3-tbd`, `scr-m3-shcs-tbd`, `scr-m3-8-shcs`, `washer-m3-15`) carry it. The priority number stays as the urgency of the work; the state is now said in words.
- The changes page's "To be removed" caption is 13px and never wraps: at 10px beside a two-line label the trash can read as a smudge and barthel had to ask whether an icon was there at all.
- **The image-tile marker never reached the tile.** It positions absolutely, but the `Popover` root it lives in is itself `relative`, so it anchored to a zero-size inline wrapper after the image and rendered below the picture on every surface that uses it. `Popover` now takes a class for its root; the marker variant uses it to sit at the tile's bottom-left (`-1px`, mirroring the hardware page's length stamp in the bottom-right), and a removal mark is red.
- Regenerated `catalog.generated.json` with `generate.py --metadata-only` in each commit.

**Deletion is tracked in #536 (screw) and #478 (washer).** I cannot delete either part object myself: `check_generated_pins.py` refuses a change that drops a minted uid from `parts.json`. Same wall as #453 and #454.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -352,6 +352,17 @@
       }
     },
     {
+      "id": "delete-unused-m3-8-shcs-screw",
+      "name": "Delete the unused M3 x 8 mm socket/button head screw",
+      "priority": "P4",
+      "description": "This screw is not used anywhere in the machine any more. It was added as the clamp screw holding the C-channel input gear (12T) onto the NEMA 17 shaft flat; the C-channel overhaul established that the screws on that assembly are countersunk, so the joint carries the M3 x 8 mm countersunk (scr-m3-8-cs) instead, and this row has no quantity anywhere. It should be deleted outright, but a minted part id has to stay resolvable once it exists, and the catalog's own consistency check refuses a change that removes one, so deleting it needs that check overridden by someone who can make that call, not an edit from here. Tracked as sorter-v2 issue #536.",
+      "targets": {
+        "hardware": [
+          "scr-m3-8-shcs"
+        ]
+      }
+    },
+    {
       "id": "unconfirmed-servo-adapter-heat-insert",
       "name": "Servo adapter's 4x M3S heat inserts are unconfirmed against the current CAD",
       "priority": "P2",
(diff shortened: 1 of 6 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #538, opened 2026-09-04
Fix the print-colour prose the C-channel overhaul left behind

basically-balloon:
The C-channel overhaul (`e22fce30`, #534) rewrote the `color` fields on the C-channel parts and left every `description`, `internal_notes` and `caption` around them saying the old colours. Extremetaz planned a print off that prose and was going to run the finned rotor in Ivory White; trainphreak asked for the text to be fixed rather than corrected in chat.

Colour prose only. Nothing here changes a `color` field, an STL or a version.

### What the catalog said vs what its `color` field says

| part | prose | `color` |
|---|---|---|
| `stator` | 1 gray + 3 black | `fixed: ash-gray` |
| `rotor-faceted` | 2 white + 1 gray | `fixed: ash-gray` |
| `rotor-finned` | White (x3: description, notes, caption) | `fixed: ash-gray` |
| `output-guide` | Black, 2 needed | `fixed: ash-gray`, 4 (3 feeder + 1 classification) |

`output-guide`'s count was stale in the same sentence as its colour, so that is corrected with it. Regenerated with `catalog/generate.py --metadata-only`.

### Docs

- **c-channel**: "the feeder parts are charcoal, the classification-channel ones ash grey" was the whole basis of the confusion. Replaced with what is actually ash grey (stator, both rotors, output guide, all four channels) and a bullet naming what is still charcoal: the gear train on the `feeder` colour role, and C1's NEMA bracket.
- **classification-chamber**: "in ash grey rather than charcoal" in three places no longer distinguishes anything, and the finned rotor is not white. The dome and the camera & LED insert are still `fixed: ivory-white` and are untouched.
- **output-guides**: "the feeder colour" is now a fixed ash grey.

### Third commit, separable

`bearing-holder-left/right` and the six servo bracket/adapter parts carry "defaulted to frame (black); confirm" in `internal_notes` while their `color` is the `chute-core` role, which defaults to ivory white. Same class of bug, unrelated to the C-channel, and agent-facing text only (`internal_notes` is not emitted to the app, so `catalog.generated.json` is unchanged by it). Drop that commit if you would rather keep the two things apart.

### Deliberately not touched

- **Photo alt text** calling the built rotor white. It describes the photograph, not the prescribed colour.
- **`output-guides.md` still says "Two per machine"** in its frontmatter and body, and the part is now named "Short and Angled Output Guide" at 4 per machine. That is a quantity and naming fix on an already-flagged AI-draft page, not a colour one.
[...]

PR #540, opened 2026-09-04
Four output guides, one per channel, not two at the handovers

basically-balloon:
Extremetaz asked whether the Short and Angled Output Guide is really 4, whether C1 still needs one when the bulk bucket is used, and whether C4 needs one. The catalog answers yes to all three; the docs still describe the pre-overhaul arrangement, so this fixes them.

### What the tree says

All four channel assemblies (`bulk-bucket`, `channel-two`, `channel-three`, `classification-chamber`) carry `{"part": "output-guide", "qty": 1}`, each with the same connection:

> The guide pushes onto the C-channel drive and is held by the fit alone.

Until `e22fce30` (#534, 2026-09-02) there were two, on C2 and C3 only; the overhaul added one to C1 and one to C4 as part of consolidating all four channels to one shape. `classification-chamber` v3's own message is "it gains the output guide".

### What the docs said

- **output-guides.md**: "Two per machine", `parts_needed: qty 2`, and a lede and opening paragraph built on the guide bridging the gap between two channels. Its draft banner also claimed the part "is in no assembly there", which the tree now contradicts.
- **arranging-c-channels.md**: step 4, "One output guide at each handover, C1 to C2 and C2 to C3".

The count was the visible error; the model behind it was the real one. A guide belongs to the channel it is mounted on, not to a handover, which is why four channels take four guides rather than three handovers taking three.

### Sources

- Quantities, placement and the joint: `parts-calculator/catalog/parts.json`, the four channel assemblies.
- Purpose, for the rewritten opening paragraph: mneuhaus in `#c-channels`, 2026-03-04, "purpose of the 'output guide' is to prevent stuff going in infinite circles", and 2026-02-27, "there to prevent stuff from going in circles".

### Still unrecorded, and still flagged as such

Where on the drive the guide seats, at what angle, and how far it projects over the exit. The draft banner stays for that reason.

### Note on #538

That PR touches `output-guides.md` too, one line in step 1 ("Prints in the feeder colour" to ash grey). This branch does not touch that line, so the two should merge in either order without conflict. #538 is the colour pass; this is the count and the purpose.

### DIFF
--- a/docs/src/content/hardware/assembly/feeder/arranging-c-channels.md
+++ b/docs/src/content/hardware/assembly/feeder/arranging-c-channels.md
@@ -88,7 +88,7 @@ The numbers this page most needs, and none of them are written down anywhere tod
 
 {% include step.html n="4" title="Fit the output guides" %}
 
-One [output guide]({{ '/hardware/assembly/feeder/output-guides/' | relative_url }}) at each handover, C1 to C2 and C2 to C3.
+One [output guide]({{ '/hardware/assembly/feeder/output-guides/' | relative_url }}) on each of the four channels. Each one belongs to the channel it pushes onto, not to the gap between two, so the classification channel gets one as well.
 
 <div class="img-placeholder">Image coming</div>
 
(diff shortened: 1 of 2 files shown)


basically-balloon:
Superseded by #601. The framing on these pages was the real error and #601 carries it, but the count that follows from it is two (C2 and C3), not four: C1's bulk cap has a guide built into it and C4's finned rotor sweeps through where one would sit, so #601 takes the part off both assemblies.

Issue #541, opened 2026-09-04
C1: the Bulk cap and the Short and Angled Output Guide interfere by 0.315 cm³


basically-balloon:
The Bulk cap and the Short and Angled Output Guide both go on C-channel 1 in the assembly tree (`bulk-bucket`), and they interfere. Reported by ReveryX in Discord, 2026-09-04, as "the bulk bucket covers the mounting point"; measuring it, the mounting point is fine and the guide is clocked about 5.5 degrees too far round.

![Section at z = 45 mm](https://assets.basically.website/sorter-parts/c1-bulk-cap-output-guide-clash-full-4bda026498a4.png)

## Measured

All three STLs are in one assembly frame. Confirmation that they are: the cap's groove measures r 116.7-120.3 and the stator's wall r 117.0-120.0, i.e. 0.3 mm clearance a side, and `bulk-cap` against `stator` booleans to exactly zero.

| | |
|---|---|
| Interference volume | **0.315 cm³** |
| Location | r 113.45-116.78 mm, az -45.10 to -40.00 deg, z 40.0 to 50.0 mm |
| Cross-section | 32.2 mm², constant from z = 42 to the guide's top face at z = 50 |
| Cap walls at z = 45 | inner r 113.5-116.7, outer r 120.3-123.5 |
| Cap relief cut | az **-84.0 to -45.5**, 38.5 deg wide |
| Guide | az **-63.3 to -40.0**, 23.3 deg wide |

The cap already has a relief cut for the guide and it is 15 degrees wider than the guide needs. The guide simply sits 5.5 degrees past the cut's -45.5 edge, so the top 10 mm of its outer corner buries itself in the cap's inner wall.

## Why this is new

The previous guide was not close to fitting: **v1 (`8j3e`, the original straight guide) booleans against the same cap at 39.7 cm³**, across az -96 to -52 and in to r = 15. C1 carried no output guide in the tree until `e22fce30` (#534, 2026-09-02), which both added one to C1 and made the guide Short and Angled. It very nearly fits; this is the last 5.5 degrees.

## What the model should become

Either clock the guide back into the relief, or extend the cap's relief cut past -40. Which of the two is the right fix is a CAD call, not one I should make: the guide's position relative to the C-channel exit is presumably set by where pieces leave the rotor, and moving it may not be free.

[...]

PR #542, opened 2026-09-04
P0: the output guide fouls the stator, the C1 bulk cap and the C4 finned rotor

basically-balloon:
Two P0 entries on the same part, both from builders fitting it.

**Requested by ReveryX (@reveryx) [from Discord](https://discord.com/channels/1430279849171222682/1545190826551017492/1545473777989853225): "the bulk bucket covers the mounting point, it is not possible to mount the output guide at the same time"**, and [again](https://discord.com/channels/1430279849171222682/1545190826551017492/1545484653744623768): "IMO, the output guide should be on the left side of the opening, not on the right... it at least is on the left side on the other C channels.. all parts would need to move up before they can leave the bluk bucket when it is mounted on the right side".

**Requested by Extremetaz (@extremetaz) [from Discord](https://discord.com/channels/1430279849171222682/1437920132893511740/1545557815064526918): "from fitting, the output guide is not required on C1 when the bulk bucket is in use. The output guide is also not campatible with the finned rotor in C4. So the count should be 2 of the short and angled output guide, not 4."**

Per the part-problem procedure these are `changes` entries rather than docs prose. The CAD work is #541 and #543.

## 1. The guide is clocked wrong: it overlaps the stator, and the bulk cap on C1

Sections at z = 45 and z = -12

| | |
|---|---|
| `output-guide` ∩ `stator` | **3.014 cm³**, r 102.99-119.99, az -45.05 to -40.00, z -19.0 to 50.0 |
| `output-guide` ∩ `bulk-cap` | 0.315 cm³, r 113.45-116.78, z 40.0-50.0 (C1 only) |

The stator overlap runs the full height of both parts and is present on **all four channels**, since one guide STL serves them all. The bulk cap did not cause it; it made it visible on the one channel that has a cap.

| feature | azimuth | width |
|---|---|---|
| stator opening at the wall (z ≥ 20) | -90.1 to -45.1 | 45.0 deg |
| stator opening at the base flange | -87.2 to -45.2 | 42.0 deg |
| bulk cap opening beside its integral tongue | -84.0 to -45.5 | 38.5 deg |
| output guide | -63.3 to -40.0 | 23.3 deg |

Rotating the guide **-5.1 to -18.3 deg** clears the stator and the cap both. That does not settle it: the stator's base flange indexes on a **25.714 deg notch grid** (14 positions, 4.8-5.0 deg wide; measured centres -167.1, -141.4, -38.5, -12.8, 12.9, 38.6, 64.3, 90.0, 115.8, 141.5, 167.2, with -64.2, -89.9 and -115.6 absent into the exit opening and the motor gap). The guide's base lobe pair sits at **-43.5 deg**, off the grid, and the nearest grid position inside the opening (-64.2) needs -20.7 deg, 2.4 deg outside the clear window.

So the exported clocking is not the assembled clocking, and the entry says so rather than picking one of the two fixes. Tracked as #541.

## 2. The guide and the C4 finned rotor cannot both be fitted

Plan section at z = 20 and the radius-against-height elevation

Extremetaz is right, and it is not a fit problem, it is a rotating clash. `output-guide` rotated against each rotor at 5 degree steps through a full turn:

| rotor | positions interfering | max volume |
|---|---|---|
| `rotor-finned` (C4) | **24 of 72** | **0.504 cm³** |
| `rotor-faceted` (C1-C3) | 0 of 72 | 0 |

The clashing positions fall in five clusters 72 deg apart, one per fin. The guide's inner face is at r 67.82; the finned rotor's body at z 20 is r 64.6 (about 3 mm clear) but its five fins reach r 85.4, and the fin sweep is up to **21.2 mm** past the guide's inner face over the band z 8 to 29. The faceted rotor clears by 1.5 to 1.8 mm along its whole height, which is the clearance the guide is dimensioned to.

**No clocking fix touches this**: a rotor covers every azimuth, so the clash is radial. C4 had no guide in the tree until `e22fce30` (#534, 2026-09-02) consolidated all four channels onto one shape and gave `classification-chamber` v3 the guide. The room had circled it from the other side before: Marc [ran C4 without a guide in May](https://discord.com/channels/1430279849171222682/1460692585881534579/1506291102913659021), and Spencer [asked whether pieces were jamming between the finned rotor and the stator before reaching the guide](https://discord.com/channels/1430279849171222682/1460692585881534579/1506376502877491290). Tracked as #543.

**Correction, 2026-09-07.** I originally read the bulk cap's cut as a purpose-made seat for a separate guide, and that is backwards. Spencer has since filed #580, *remove output guide from bulk bucket*, on the grounds that "the original version is built into the bulk bucket", and the geometry agrees: **39.71 cm3 of the original straight guide (v1, `8115840b817f`) lies inside the `bulk-cap` solid, 73% of that guide's 54.13 cm3.** Sampling the cap's solid at fixed radii through a full turn finds material only in a band at azimuth **-96 to -74 deg**, reaching inward from r 103 at the base to r 51 at z 25 and r 15 at z 45. A bucket floor would be present at every azimuth; this is a wall at one azimuth leaning inward as it rises, standing in the stator's exit opening, which is the old guide moulded in. So the 0.315 cm3 cap clash is not a clocking error and no rotation fixes it: C1 does not want a separate guide.

That leaves the count Extremetaz asks about at **3 on the fitting, and 3 in the catalog once #580 lands**: C2 and C3 keep theirs, C1's comes out because the cap already has one, and C4 cannot take this guide in its current shape. Removing the C1 line is a quantity change and #580's own job, so this PR still changes no quantities. It only badges the defects, and the stator overlap above is unaffected by any of this because it is on all four channels.

## Shared frame

Not assumed. For the cap: its groove measures r 116.7-120.3 against a stator wall of r 117.0-120.0, 0.3 mm clearance a side, and `bulk-cap` against `stator` booleans to exactly zero. For the rotors: `output-guide` against `rotor-faceted` booleans to exactly zero at all 72 rotations, which a wrong relative pose would not produce.

## What is in this PR

Two entries in `changes`, both P0 and `condition: "broken"`, with their section drawings. The first targets `output-guide`, `stator`, `bulk-cap` and all four channel assemblies plus `c-channel`; the second targets `output-guide`, `rotor-finned` and `classification-chamber`. Regenerated with `catalog/generate.py --metadata-only`; `check_generated_pins`, `check_asset_urls`, `check_versioning` and `check_connections` pass locally.

No workaround in either entry, and no docs edit: the feeder pages read the warning off `ChangeStatus` once this lands.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

PR #544, opened 2026-09-05
Bin retainers take M5 x 16, not M5 x 12

basically-balloon:
The bin retainers need an **M5 × 16**, not the M5 × 12 the docs and the catalog both call for.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1487518366020276397/1545685109615759371): "can you measure the Stl for the bin retainers, check what screw size is needed to attach them."** She then reported the failure herself, [from Discord](https://discord.com/channels/1430279849171222682/1545685109615759371/1545686385615446108): "That's what I'm getting, I tried the m5x12's and they didn't clear the whole. I think it's and m5x16 needed."

## What the STL says

Measured off the pinned `bin-retainer-left` (`hqz0`) and `bin-retainer-right` (`ybpv`) masters, which are exact mirrors of each other:

- **Holes:** 5.400 mm, two per retainer, 76.00 mm apart, axis square to the mounting face. Standard M5 clearance.
- **Bore the screw crosses:** 12.80 mm. That is 12.00 mm of plastic between the head face and the face that sits on the A/G extrusion, plus a 0.8 mm locating rib (4.68 mm tall, centred on the hole line) that drops into the slot.
- **M5 × 12** therefore stops level with the extrusion face with no thread in the T-nut at all, which is what she hit on the bench.
- **M5 × 16** stands 4.0 mm past the face. A 2020 ball-spring roll-in M5 nut is about 4.3 mm thick ([handsontec datasheet](https://www.handsontec.com/dataspecs/screw-nut/2020M5-T-Nut.pdf), C = 4.3 mm) sitting in a slot about 5.9 mm deep ([2020 profile spec](https://premiumalu.com/product/2020-aluminum-extrusion/)), so its thread starts roughly 1.6 mm down and about 2.4 mm of the screw is in it. Light, and it is the longest standard length that fits.
- **M5 × 20** does not fit: 8 mm into a 5.9 mm slot bottoms on the slot floor before it clamps anything.

The head also wants to be a socket cap rather than a button head. The flat it bears against only reaches 4.13 mm below the hole centre, so a button head (9.5 mm) overhangs the bottom edge and a washer will not sit flat.

## Changes

- `regular-layers.md` step 2 and `bottom-two-layers.md` step 5 now call for M5 × 16, and both parts lists fold the bin-retainer screws into their existing `scr-m5-16-shcs` line (36 → 60 and 48 → 96). The `scr-m5-12-shcs` line is gone from both; it is still correct everywhere else it appears (NEMA 23 mount, PSU box, Orange Pi mount) and is untouched there.
- `catalog/parts.json`: the `layer` assembly's 24 `scr-m5-12-shcs` merge into its `scr-m5-16-shcs` line, 12 + 24 = 36. Stamped as v3 with a fresh uid (`mnpn`, superseding `jog7`), `breaking: false` — an already-assembled layer mates outward exactly the same.
- The joint is now recorded in the assembly's `connections` with its measured `through_mm`, and the sentence enumerating it has come out of the description, per this directory's `CLAUDE.md`.

The counts on these two pages come from [#449](https://github.com/basicallysource/sorter-v2/pull/449), which reconciled them against each page's own steps but took the M5 × 12 at face value.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-two-layers.md
+++ b/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-two-layers.md
@@ -31,9 +31,7 @@ parts_needed:
   - part: caster-wheel-m6
     qty: 6
   - part: scr-m5-16-shcs
-    qty: 48
-  - part: scr-m5-12-shcs
-    qty: 48
+    qty: 96
 tools_needed: [Hex key, Tape measure]
 ---
 
@@ -85,8 +83,7 @@ Here's where the {% include fastener.html size="M5" variant="socket-button" leng
 - **24** for the foot extensions, 4 per corner (step 2 below), two into each layer
 - **12** for the second layer's External bracket — bottom verticals, 2 per corner (step 3 below)
 - **12** for the bottom layer's External bracket — foot covers, 2 per corner (step 3 below), the same as the bottom vertical it replaces
-
-The {% include fastener.html size="M5" variant="socket-button" length="12" %} count is the same arithmetic: **24** per layer for the bin retainers, so **48** across both.
+- **48** for the bin retainers (step 5 below), 24 per layer, so 48 across both
 
 You should already have two [hex frames]({{ '/hardware/assembly/distribution/bin-frame/hex-frame/' | relative_url }}) built, one for each layer. The verticals are the part that differs here, and they are step 2 below.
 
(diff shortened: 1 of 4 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #545, opened 2026-09-05
Machine tree: n bin layers plus one top interface, not n-2 between two

basically-balloon:
The assembly tree counted `n − 2` bin layers while the printed-parts and framing models both counted `n`, so every build's hardware totals were short. This makes the tree agree with them.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1545685109615759371/1545694727955808316): "Ok, let's push this through as a screw length change. Then make a new one for the layer count issue:"** She found it by [checking the M5 × 16 total](https://discord.com/channels/1430279849171222682/1545685109615759371/1545692007920312390) on #544's preview, [corrected my reading of the bracket covers](https://discord.com/channels/1430279849171222682/1545685109615759371/1545699166347333742) ("For external bracket covers you need n x6. They get used in the top interface"), and [settled the last open question](https://discord.com/channels/1430279849171222682/1545685109615759371/1545712373074104341): "you could say the top interface owns the spare layer connector. As it slides onto the external bracket, on that hex ring." She had already caught the same defect from the frame side [on 26 August](https://discord.com/channels/1430279849171222682/1542146533968052367/1542160152768028823) with a build photo showing 6 hex frames on a 5 layer machine.

## The shape of the machine

**n bin layers plus one top interface, so n + 1 hexagons.** The lowest bin layer carries foot covers in place of the bracket covers and bottom verticals every other frame uses, and the bottom interface hangs underneath it with no hexagon of its own — [bottom two layers, step 6](https://docs.basically.website/hardware/assembly/distribution/bin-frame/bottom-two-layers/): *"It does not share piece D's legs and it does not get a hex frame of its own."*

## Changes

- **`filament.ts`: new `non-bottom-layers` quantity**, `max(0, n − 1)`. It mirrors the `non-bottom` half of a part's existing `layer_scope`, which is what the printed-parts side already uses to say the same thing. Without it the tree cannot express "every bin layer but the lowest", and flipping `layer` to `per-layer` alone would put the bracket covers at 6(n + 1) instead of 6n.
- **Four new assemblies** for the lowest layer: `external-bracket-with-foot` (side + foot cover, same 2 screws the bottom vertical uses), `outer-hex-frame-bottom`, `hex-frame-bottom`, `bottom-layer`.
- **`superstructure` v2**: `layer` runs `non-bottom-layers`, plus one `bottom-layer`.
- **`bottom-interface` v2**: drops its `layer-frame` and its `layer-connector`.
- **Layer connector descriptions** corrected from "1 per bottom interface" to the top interface, matching their `quantities` and her call.

Both version entries are `breaking: false`: nothing physical changed, only how many of each the tree says a machine has.

## Verification

The other two models are independent oracles, and the tree now lands on both **exactly, at every layer count from 2 to 8**:

| | before | after |
|---|---|---|
| printed frame/layer parts disagreeing with the 3D-parts tab | 8 | **0** |
| `ext-2020-ag`, `ext-2020-bh` vs `framing.ts` A+G, B+H at n=3 | 18 vs 24 | **24 vs 24** |
| same at n=5 | 30 vs 36 | **36 vs 36** |

Specifically, `ext-bracket-foot-cover` goes from appearing in **no assembly at all** to 6, the bin retainers from 12(n−2) to 12n, and Frame 90° bracket / crossbeam / External bracket — side pick up the missing hexagon.

**M5 × 16 rises by 96** at every layer count: 163 → 259 at n = 3, 283 → 379 at n = 5. That is the shortfall being corrected, not new hardware. My earlier "at least 72" was a floor counting only the retainers and the missing ring; the rest is that ring's brackets and the second layer's flange screws.

`npm run check` 0 errors 0 warnings, `npm run build` clean, and `check_connections` / `check_versioning` / `check_generated_pins` all pass.

## Confirmed on a build

BrickCycleAlice, [from Discord](https://discord.com/channels/1430279849171222682/1545685109615759371/1545716605349134467), after reading the diagnosis: *"that's what happened to me, I printed, cut and ordered then went to build and the count of the hardware didn't match. But the prints and cuts did."*

That is the failure this PR predicts, from the outside: the 3D-parts and framing models fed her prints and her cut list correctly, and only the tree-derived hardware order came up short. Her machine is 5 layers ([76 M5 × 12 on the old catalog](https://discord.com/channels/1430279849171222682/1545685109615759371/1545692007920312390) pins it), so the shortfall she hit is the 96 M5 × 16 above, plus 6 foot covers she had printed and had no screws listed for.

## Not covered

Bins and funnels are in no assembly either, so they still read 0 through the tree. That is a separate gap and out of scope here.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -6728,11 +6728,43 @@
       ],
       "versions": [
         {
+          "version": "1",
+          "uid": "ywb3",
+          "message": "As recorded before the first stamped change.",
+          "lines": [
+            {
+              "part": "top-plate",
+              "qty": 1,
+              "uid": "1g22"
+            },
+            {
+              "assembly": "interface",
+              "qty": 1,
+              "uid": "vhrk"
+            },
+            {
+              "assembly": "layer",
+              "qty": "middle-layers",
+              "uid": "mnpn"
+            },
+            {
+              "assembly": "bottom-interface",
+              "qty": 1,
+              "uid": "67uq"
+            },
+            {
+              "part": "scr-m5-16-shcs",
+              "qty": 6,
+              "uid": "2e0s"
+            }
+          ]
+        },
+        {
           "version": "2",
           "date": "2026-09-05",
           "breaking": false,
           "message": "The stack is n bin layers plus one top interface, not n - 2 layers between two interfaces. The layer node now runs non-bottom-layers (n - 1) and the new bottom-layer node runs once, because the lowest layer carries foot covers instead of bracket covers and bottom verticals. Brings the tree onto the same counts the 3D-parts and framing models already used: n + 1 hexagons, 12n bin retainers, 6 foot covers. No physical change.",
-          "commit": null
+          "commit": "36826f49"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #546, opened 2026-09-05
docs: hex frame is builder-verified, drop both warnings

basically-balloon:
barthel built a hex frame straight from this page and it worked, so the two warnings on it are no longer true and both come off.

**Removed the `warning:` frontmatter banner.** It said the page was "reorganized from a Discord build walkthrough, not yet reviewed by a builder". It has now been checked step-by-step against a build, and the convention is that a page verified against a build loses the note entirely rather than softening it.

**Removed the closing callout** about the page not being reconciled against the parts calculator's assembly groupings. barthel reconciled it by hand: the calculator's `Outer hex frame` node lists `External bracket — cover` and `External bracket — bottom vertical` (they come in via the `External bracket with support` sub-assembly), and those two are not fitted during this build, they go on in [Regular layers](https://docs.basically.website/hardware/assembly/distribution/bin-frame/regular-layers/) and the other layer pages. His call is that this is fine and that the individual element counts match the page, so there is nothing left for the callout to warn about.

**Second ask, same thread:** the "Building the bottom layer's frame?" callout in step 3 is now an information note rather than a warning, at barthel's request. That is a bare `<div class="callout">`, which is how the site already writes a neutral note (`bottom-two-layers.md`, `top-interface.md`); the yellow bar and the warning icon are gone, the text is unchanged.

**Not touched:**

- The other two callouts on the page stay, as build instructions rather than caveats about the page: the T-nut count in zed0's video, and the extra spoke T-nuts a bottom layer needs for the bottom interface.
- No change to `parts-calculator/catalog/parts.json`. The grouping barthel flagged is the recorded one and he approved it as is.
- `regular-layers.md` keeps its own reorganization warning. Nobody has built from that page since the reorg, so it is still unreviewed.

Verified by barthel's build, reported in #balloon-does-docs.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/bin-frame/hex-frame.md
+++ b/docs/src/content/hardware/assembly/distribution/bin-frame/hex-frame.md
@@ -9,10 +9,6 @@ lede: The hexagonal aluminum-and-bracket ring shared by every layer. Build one p
 permalink: /hardware/assembly/distribution/bin-frame/hex-frame/
 author: barthel
 contributors: [brickcyclealice, zed0]
-warning: >-
-  **Reorganized from a Discord build walkthrough, not yet reviewed by a
-  builder.** This page has not been checked step-by-step against a build.
-  Correct it as you go.
 parts_needed:
   - part: ext-2020-ag
     qty: 6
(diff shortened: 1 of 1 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #547, opened 2026-09-05
docs: bin retainers get their own page, T-nuts move to the step that uses them

basically-balloon:
**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1545727470026883232):** *"I'd like us to give more consideration to this feature of the T-nuts and install them right at the step where they're needed. That way, they wouldn't be scattered throughout the entire documentation and would be easier to locate. ... Ideally, I'd like to dedicate a separate, short page to attaching the bin retainers, since this is described in several places."* He opened the topic [here](https://discord.com/channels/1430279849171222682/1536088194557419660/1545721430682108068), noting that the bin retainers are what drives the screw and T-nut count per layer.

The catalog part is `tnut-m5-2020`, "M5 roll-in T-nut, 2020", spring-loaded roll-in. A roll-in nut goes into the slot anywhere along its length, so nothing has to be pre-fitted while an extrusion end is still open. The hex frame page was written as if the opposite were true. The style was chosen for exactly this property, in [#distribution on 2026-03-07](https://discord.com/channels/1430279849171222682/1437144782819426537/1479944747765334220): *"i like these ones a lot because you can take them out and insert them anywhere in the aluminum profile"*.

## New page: Bin retainers

`docs/src/content/hardware/assembly/distribution/bin-frame/bin-retainers.md`, authored to barthel.

The step was written out twice, in [Regular layers](https://docs.basically.website/hardware/assembly/distribution/bin-frame/regular-layers/) step 2 and [Bottom two layers](https://docs.basically.website/hardware/assembly/distribution/bin-frame/bottom-two-layers/) step 5, with the second telling you to follow the first. It is now one page that both call, carrying the retainers, their 24 screws and their 24 T-nuts. Both callers drop those parts from their own `parts_needed` and link instead, the same pattern the hex frame split used.

It is a bin-layer page, not a hex frame page. You build N+1 hex frames and only N of them hold bins: the top interface's frame takes no retainers.

## New page: Fitting T-nuts

`docs/src/content/hardware/helpers/t-nuts.md`, a helper page, authored to barthel. What a roll-in nut is, how it goes in, and the two consequences that matter (fit it at the step that uses it; you can undo a joint without losing the nut). The video is Marc's, [posted in #distribution](https://discord.com/channels/1430279849171222682/1437144782819426537/1479945463192092763) in the thread where the style was decided, embedded from his own timestamp of 1:13.

The page also collects the two places a **slide-in** nut has to be fitted early, so that constraint lives in one place instead of being repeated as a conditional on every page.

## Hex frame page: the 24 T-nuts come off

`hex-frame.md` carried them in four places: `parts_needed`, the intro paragraph, the fastener-total paragraph, and the callout above zed0's video. All four now point at the bin retainers page. What stays is the slide-in exception in step 1, which is the only case where something really does have to happen at this stage, and it now links to the helper page.

The frame's own fastener total is unchanged: 12 M5 x 16, no T-nuts. Step 1's T-nut *option* for the inner holes is unchanged too, and it is now stated as 12 more nuts per frame rather than being netted against the 24.

## Catalog: the retainer T-nuts were missing from the machine entirely

`parts-calculator/catalog/parts.json`, `layer` v3 to v4, `breaking: false`. This is a real BOM bug and it is how the docs and the catalog got out of step, so it goes in with the rest.

The 24 T-nuts the bin retainers bolt into were dropped twice, each time on the assumption that the other node still carried them:

- `layer` v2, 2026-08-29: *"Dropped tnut-m5-2020 qty 24 -- those are the hex frame's own pre-installed T-nuts (bin retainers reuse them, not new ones)"*.
- `hex-frame` v2, 2026-08-31: *"the 24 T-nuts are dropped"*, as part of the Inner/Outer hex frame restructure.

After the second one no assembly in the tree listed them, so no hardware list has included them since. They are restored on `layer`, alongside the 24 `scr-m5-16-shcs` that pass through them, which is also where they now sit in the docs. The `bin-retainer-left` connection note is corrected to match. `catalog.generated.json` regenerated with `generate.py --metadata-only`.

**Not fixed here**, both pre-existing and both worth their own change:

- `layer` is instantiated `max(0, layers - 2)` times in the machine tree while the 3D-printed-parts tab multiplies the same section by the full layer count, so every per-layer fastener on this node reads two layers short. Same defect as the M5 x 16 shortfall in #544.
- `bottom-two-layers.md` lists 6 `ext-bracket-cover` for two layers where the two collars look like they want 12.

Checks: `check_versioning.py` clean, `check_generated_pins.py` clean, `check_docs_refs.py` warns only that `layer` was revised (expected) plus the pre-existing set. Docs build clean on Node 22.

## The External bracket — cover moves to the hex frame page

**Asked by barthel [in the same thread](https://discord.com/channels/1430279849171222682/1545727470026883232/1545731282288517161):** *"why don't we attach the external bracket cover when we're building the hex frame? Does that make a later step in the build more complicated?"* It does not, so it moves. He then [pointed at the page's own words](https://discord.com/channels/1430279849171222682/1545727470026883232/1545731654914678835) as the argument for it: *"Auf der regular layer seite steht sogar: 'Fit the External bracket — cover onto the External bracket — side now, at each corner of your hexagon. It's difficult to slide on once piece C is in place.'"* That warning is the sentence this change removes, because the frame build is a strictly earlier moment than the one it was protecting.

`hex-frame.md` gains 6 `ext-bracket-cover` and a step 9 that fits them; `regular-layers.md`, `bottom-two-layers.md` and `top-interface.md` drop them from their own `parts_needed`, and Regular layers' warning about fitting the cover before piece C becomes a note pointing back at that step. The machine total is unchanged: 6 per frame across N+1 frames, less the bottom layer's frame, is 6N, which is what the pages added up to before.

**The bottom layer's frame is the exception and step 9 says so.** It takes `ext-bracket-foot-cover`, one printed part in place of the cover and the External bracket — bottom vertical together, per `bottom-two-layers.md` step 3.

**`ext-bracket-cover` has no fastener holes**, so this moves no screws. Measured on the STL: silhouette taken along +z, radially inward, and both 45° diagonals, then flood-filled from the border to find voids fully enclosed by material. The cover returns **zero** enclosed voids in every direction. The same pass over `ext-bracket-foot-cover` returns its 28.8 × 29.2 mm bore plus **two 5.6 × 5.6 mm holes at (±20.7, −356.6)**, which are the Ø5.5 clearance pair its step already documents, so the method finds holes where holes exist.

That corrects `bottom-two-layers.md` step 2, which said the cover and the foot cover "both take 2 M5 × 16 screws". Only the foot cover does. Step 3's description of the foot cover's own pair was already right.

**Not fixed here:** `ext-bracket-foot-cover` appears in no assembly's `lines` at all, so the catalog gives every corner of every layer a cover and no machine BOM contains a foot cover. Modelling the bottom layer as its own node is a bigger change than this PR.

## Follow-ups from the same thread

**Regular layers now shows where its 36 M5 × 16 come from.** [Asked for here](https://discord.com/channels/1430279849171222682/1545727470026883232/1545754280391217262): *"Check the screw count on the sides, where bin retainer was described before."* Both pages check out and the split conserved the totals: Regular layers 60 → 36 + 24 on the retainer page, Bottom two layers 96 → 48 + 48, and the machine's M5 × 16 across these pages is 72N − 12 either way. Regular layers had no breakdown of its number (Bottom two layers already had one), so it now lists the three pairs per corner that make up the 36.

**The catalog is a different story and is not touched here.** It carries 60 M5 × 16 per layer against the docs' 72. The missing 12 is one of the two bracket pairs at each corner: `external-bracket-with-support` lists 2 screws where the corner takes 4 (Spencer's own description of the part is *"Has 4 screws. Two that hold the bottom to the top and two that hold the vertical extrusion"*, and the `external-bracket` helper node already says 4). Bumping it is not the fix, because that node sits inside `hex-frame`, which the top interface also uses and which has no External bracket — bottom vertical at all. The bottom vertical and its screws belong to the layer standing on the frame, not to the frame, which is the shape the docs now describe. That is the same structural knot as the `middle-layers` n − 2 count and belongs in its own change. Previously flagged in #373.

**The video starts at 1:13.** barthel [confirmed that timestamp](https://discord.com/channels/1430279849171222682/1545727470026883232/1545735780595408947): *"1:13 is the correct timestamp."* I had briefly moved it to 1:08 off a transcript estimate; his reading of the video wins, so it is back to Marc's original `?start=73`.

**Fitting T-nuts is now linked from every page that calls for T-nuts.** [Asked for here](https://discord.com/channels/1430279849171222682/1545727470026883232/1545736402820669491): *"Please add a link to the 'fiiting t-nuts' page"*. Added on Bottom interface, Bottom two layers, Regular layers, Top interface, PSU box, Control board housing, Orange Pi mount and the electronics installation overview, at each page's own first T-nut mention. Hex frame and Bin retainers already linked it. Same shape as the heat inserts helper: the procedure lives in one place and the steps point at it.

**The T-nut warning above the hex frame video is gone.** [Asked for here](https://discord.com/channels/1430279849171222682/1545727470026883232/1545734093520502895): *"I think we need the video ... and the photo below it showing the four T-nuts, but we no longer need the warning above the video."* The video and the photo stay; the callout telling you which of the T-nuts in them belonged to this page is no longer needed now that none of them do.

**On the cover being on every hex frame** ([barthel](https://discord.com/channels/1430279849171222682/1545727470026883232/1545732952409776268)): every frame except the bottom layer's, which `bottom-two-layers.md` gives External bracket — foot covers instead. Step 9 carries that exception, and after barthel's [“the foot cover is a cover too”](https://discord.com/channels/1430279849171222682/1545727470026883232/1545745199790039120) it says that frame takes no *External bracket — covers* rather than "no covers", which was contradicting the next sentence.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/bin-frame/bin-retainers.md
+++ b/docs/src/content/hardware/assembly/distribution/bin-frame/bin-retainers.md
@@ -1,0 +1,49 @@
+---
+layout: default
+title: Bin retainers
+type: how-to
+section: hardware
+slug: assembly-bin-retainers
+kicker: Bin frame — Bin retainers
+lede: The pair of rails on each face of a layer that a bin slides into. Same on every bin layer.
+permalink: /hardware/assembly/distribution/bin-frame/bin-retainers/
+author: barthel
+contributors: [zed0, brickcyclealice]
+warning: >-
+  **Split out of Regular layers and Bottom two layers, which described this
+  twice, not yet reviewed by a builder in this form.** The step itself is
+  unchanged from those pages. Correct it as you build.
+parts_needed:
+  - part: bin-retainer-left
+    qty: 6
+  - part: bin-retainer-right
+    qty: 6
+  - part: scr-m5-16-shcs
+    qty: 24
+  - part: tnut-m5-2020
+    qty: 24
+tools_needed: [Hex key]
+---
+
+Every bin layer gets the same twelve retainers: a Bin retainer (left) and a Bin retainer (right) on the front face of each of the six A/G extrusions, so each of the six faces has a pair that a bin slides down between. The quantities above are **for one layer**. [Regular layers]({{ '/hardware/assembly/distribution/bin-frame/regular-layers/' | relative_url }}) and [Bottom two layers]({{ '/hardware/assembly/distribution/bin-frame/bottom-two-layers/' | relative_url }}) both send you here.
+
+**Do this after the layer's External bracket — covers are on.** The retainers run right out to the corners, so leave them off until the corner is finished.
+
+{% include fastener-legend.html %}
+
+{% include step.html n="1" title="Put four T-nuts in each face" %}
+
+Drop 4 {% include fastener.html size="M5" variant="t-nut" text="M5 T-nuts" %} into the outward-facing slot of each A/G extrusion, two per retainer. They do not have to be positioned accurately: slide them along to meet the retainer's holes once it is held in place.
+
+The T-nuts specified for the machine are the roll-in kind, so they go into the slot here, at this step, with the frame already built. See [Fitting T-nuts]({{ '/hardware/helpers/t-nuts/' | relative_url }}) if you bought a different style, because slide-in nuts have to go in much earlier.
+
+{% include step.html n="2" title="Fasten the retainers" %}
+
+<figure class="single-figure">
+  <img class="doc-figure" src="https://assets.basically.website/sorter-docs/assembly-regular-layers-bin-retainers-installed-w1600-31bf32089e71.png" alt="Bin retainers fastened to the outer faces of the hexagon frame">
+  <figcaption><cite>Photo: zed0.</cite></figcaption>
+</figure>
+
+On each side of the hexagon, hold both the Bin retainer (left) and the Bin retainer (right) against the front face of A/G (Outer horizontal / Horizontal interface frame). The rib along the back of each one drops into the extrusion's slot and sets the height for you; the hook at the top sits on the top face of the extrusion. Fasten each retainer with 2 {% include fastener.html size="M5" variant="socket-button" length="16" %} screws into the T-nuts.
+
+Each retainer's bore is 12.8 mm deep, so a shorter M5 reaches the extrusion face with nothing left to bite in the T-nut. Use a socket head rather than a button head here: the flat the head lands on stops 4.1 mm below the hole, which a button head overhangs, and a washer will not sit flat on it at all.
--- a/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-interface.md
+++ b/docs/src/content/hardware/assembly/distribution/bin-frame/bottom-interface.md
@@ -244,7 +244,7 @@ Three Lazy Susan extrusion mounts carry the bearing assembly's weight; a hold in
 
 The screw between the extrusion mount and the hold in place is an {% include fastener.html size="M5" variant="socket-button" length="16" %}.
 
-Each mount then bolts up into the B/H spoke (158mm) already in place in the bottom layer's frame, through two more 5.5 mm M5 clearance holes, 30 mm apart, into a T-nut in the extrusion.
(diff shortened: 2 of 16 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #548, opened 2026-09-05
docs: the camera lamp assembly

basically-balloon:
Adds the camera lamp to the assembly docs. ReveryX asked for it in `#c-channels` after barthel pointed out that the lamp's assembly only existed in the parts calculator.

**New page:** `/hardware/assembly/feeder/camera-lamp/`, the arm, the shaded lamp and the camera that hang over a C-channel. Written from the [catalog's assembly tree](https://parts-calculator.basically.website/assembly?focus=camera-lamp), Spencer's photographs (from the `camera-lamp` catalog entry, re-uploaded to `sorter-docs` with their EXIF rotation baked in so they are not sideways on the page), and measurements taken off the published STLs.

### What the STLs settle

- **The 12 M3 × 12 go into 6 holes.** The C-channel arm mount and the arm each carry Ø2.8 mm pilot holes running 19.8 mm straight through the joint; brackets A and B sit on opposite faces (Ø3.5 mm clearance in a 4.5 mm plate) and screw into the same hole from either end. The camera lamp ring does the same at the far end: 4 screws, 2 holes. A 12 mm countersunk leaves ~7.5 mm of thread in a 19.8 mm hole, so two screws nose to nose still have ~5 mm between them.
- **No heat inserts anywhere.** Every hole across the seven printed parts is Ø2.8 (self-tap) or Ø3.5 (clearance). Nothing in the 4.2 mm range an insert would need.
- **The camera clasp has two screw positions, not four**, 43.2 mm apart on the diagonal: Ø3.5 clearance with a countersink on the underside of the bottom half, Ø2.8 × 5 mm pilots in the top. The catalog lists 4 × M3 × 8 and already flags the count as unconfirmed at the bench.
- **How the arm mount attaches to the C-channel is not in the CAD.** The mount has exactly two holes and both belong to the brackets. The page marks it not recorded rather than inventing a fixing.
- The reflector is 150 mm across and the cover 153 mm, friction fit, no fastener anywhere in the cover's geometry; the six LED hooks friction-fit into Ø2.7 sockets 60° apart around the reflector's rim.

### The clasp screw count

Recorded as a **P3 in `changes`** (`camera-clasp-screw-count`) rather than edited in place. Changing an assembly's `lines` is a revision that has to be stamped as a new version per `VERSIONING.md`, and that is a call for somebody with a build in front of them, not something to fold into a docs PR. The docs page states the measurement and the discrepancy in a callout. `catalog.generated.json` regenerated with `generate.py --metadata-only` in the same commit.

### The two retired pages

The light post and the overhead camera mount were retired by #534 on 2026-09-02. Both now open with a callout saying so and pointing here, both are labelled `(retired)` in the sidebar, and they come out of the feeder build order (kept for machines already built that way). C-channel step 6 points at the camera lamp instead of the light post, and its "two of the four channels carry a light post" bullet becomes the camera lamp.

### Open, and deliberately left open on the page

- **Whether C1 takes a camera.** The catalog gives all four channels a full lamp, the software has no C1 camera role, and BrickCycleAlice asked Spencer to double-check on 2026-09-05. The page carries that as a warning callout so nobody buys a third OV9732 on the strength of a docs page.
- **The LED strip.** Which strip, how much, how it is joined and how it is wired are not catalog lines yet. Said so rather than guessed.
- **The order of operations.** Nobody has built one from instructions; the warning at the top says so.

Sources: #534 (the C-channel overhaul), `parts-calculator/catalog/parts.json`, and the STLs at `assets.basically.website/sorter-parts/`.

### Builder confirmation

ReveryX, who has the parts, confirmed the Step 2 / Step 3 arrangement: six pilot holes, a screw driven into each end of every one, twelve M3 × 12 in total. That count was measured off the STLs when the page was written and is now checked against real parts, so the page says so and the top warning no longer claims every count is geometry only (`f817f5f`).

barthel asked for the sizes to be checked so nothing is left as an undefined screw. Every screw on the page carries a size already; `3eb7528` records why each is the length it is:

- **M3 × 12** at both joints. 4.5 mm of bracket leaves 7.5 mm of thread in the 19.8 mm hole, and because a screw enters from each end, a 16 would put 23 mm of screw into a 19.8 mm hole and the two would meet before either seated. ReveryX confirmed the 12 on an arm he assembled.
- **M3 × 8** at the camera clasp is the one length nobody has checked against a build: through the 6 mm bottom half it reaches about 2 mm into the top half's 5 mm pilot, where an M3 × 10 would give 4 mm and still stop 2 mm short of the top face. The callout says so rather than quietly changing the authored part.

The two `fastener not recorded` marks left on the page are not screws: the clasp-to-ring joint is a spigot in a socket, and the arm mount has no fastener to the C-channel anywhere in its geometry.

### Where the images are served from

barthel asked for the lamp's photos to live with the documentation rather than the parts calculator. Every image on the page now comes from the `sorter-docs` namespace: the four photographs were re-uploaded when the page was written (with their EXIF rotation baked into the pixels, since the derived rungs drop the flag and they render sideways otherwise), and the three renders followed in `e8ae887`. So the page does not depend on the `camera-lamp` catalog entry keeping its `images`.

Nothing is removed from the calculator in this PR. Spencer added those seven images to the assembly three days ago in #534 and people are using that page to see the build, so whether the entry keeps them is his call rather than something to fold into a docs change.

### DIFF
--- a/docs/src/content/hardware/assembly/feeder/c-channel.md
+++ b/docs/src/content/hardware/assembly/feeder/c-channel.md
@@ -11,7 +11,7 @@ author: spencer
 contributors: [barthel, brickcyclealice, christoph]
 warning: >-
   **Steps 1 to 5 come from a build**, the order of operations and the photographs are
-  BrickCycleAlice's. Step 6, the light post, is still an AI-generated first draft written
+  BrickCycleAlice's. Step 6, the camera lamp, is still an AI-generated first draft written
   from the machine assembly tree in the [parts
   calculator](https://parts-calculator.basically.website/assembly?focus=c-channel) and has
   not been checked against a machine. Correct it as you build.
@@ -52,7 +52,7 @@ The parts list above is **all four channels' worth** — the count is fixed for 
 
 - **Build 4 per machine**, three in the feeder and one for the classification channel. It is the same build four times over, and the rotor is the only thing that changes.
 - **One rotor per unit, and only one of the two.** The three feeder channels take the Rotor (faceted); the classification channel takes the Rotor (finned), the one with the fins in the photographs below. The list above already splits them the way a real machine needs them: three faceted, one finned. Colour changes with the rotor too: the feeder parts are charcoal, the classification-channel ones ash grey.
-- Two of the four channels also carry a light post, whose screws belong to the light post rather than to the list above.
+- Every channel also carries a [camera lamp]({{ '/hardware/assembly/feeder/camera-lamp/' | relative_url }}), whose parts and screws belong to that page rather than to the list above.
 
 {% include step.html n="1" title="Preparation" %}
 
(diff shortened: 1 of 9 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #549, opened 2026-09-05
docs: add an Install the bins page

basically-balloon:
Adds `/hardware/assembly/install-bins/`, the last assembly step before software setup, asked for by BrickCycleAlice in [#balloon-does-docs](https://discord.com/channels/1430279849171222682/1536088194557419660/1545748719784427580).

**What the page covers**

1. Match the bin size to that layer's funnel (half → 12 bins, third → 18), which is the choice the [funnel page](https://docs.basically.website/hardware/assembly/distribution/chute/funnel/) already makes before printing.
2. Get the bins, printed or laser cut, both given equal space.
3. Drop them in behind the bin retainers, which bolt to the frame and not to the bin, so either kind is held the same way.
4. Turn the chute through a revolution by hand and check the funnel clears every bin.

**Sources**

- Bin sets and quantities: `parts-calculator/catalog/parts.json` (`bin-half-*`, `bin-third-*`, 6 per layer of each, marked optional).
- Print cost: the catalog's own sliced numbers. Half set 12 × 171.5 g / 4.6 h; third set 6 × (121.8 + 112.2 + 121.7 g) and 6 × (3.4 + 3.2 + 3.4 h). Both land near 2.1 kg and 55 to 60 hours a layer.
- Cardboard bins are the design intent for cost and shipping (design proposal meeting, 2025-11-11), and the current tool is [bin-gen.basically.website](https://bin-gen.basically.website/), which `parts-calculator/src/routes/lasercut/+page.svelte` already links as the Laser Cut Bin Generator. Settings named on the page were read off the live tool.
- Cut order (green score, then blue perforations, then red outlines, green side face up) and the hot glue note: [spencerhubert in #distribution](https://discord.com/channels/1430279849171222682/1437144782819426537/1514123416402264199).
- Retainers, 6 of each per layer and 2 × M5 × 16 each: [Regular layers](https://docs.basically.website/hardware/assembly/distribution/bin-frame/regular-layers/).
- "The funnel lands right at the bin entrances": daddyosbricksbill, #distribution, 2026-09-02.

**Update**

BrickCycleAlice asked that the docs not refer to Spencer by name, so both prose mentions on this page are gone: "what Spencer runs" is now "what the machine was designed around", and "hot glue is what Spencer uses" is now "hot glue is what the bins at Basically are held together with". The photo credit lines and `author:` bylines elsewhere on the site are attribution, required by `docs/AGENTS.md` and checked by `validate_credits.py`, and are untouched here.

**Photos**

[...]

PR #551, opened 2026-09-05
docs: the funnel size is a builder's choice, not a consequence

basically-balloon:
[BrickCycleAlice](https://discord.com/channels/1430279849171222682/1545748719784427580/1545765809706631238): *"I just read the funnel page and it goes around in circles, the funnel picks the bins the bins pick the funnel. Both pages need to be clear, it is the builders choice which size to pick, here are the advantages and disadvantages of each, here is the recommendation for which level to put which size on, but it is builders choice."*

She is right, and the circularity had spread: the chute landing page said the funnel's size decides the bin set, the funnel page said the same, and the new install-bins page said the bins follow from the funnel. Nothing said who chooses.

**What step 1 says now.** The choice is the builder's, taken once per layer, and it is a choice about how finely that layer divides its ring: 12 bins or 18. The funnel and the bin set follow from it together, as a matched pair, and sizes can be mixed layer to layer.

Then the trade-off, so the decision can actually be made:

- **Half, 12 bins:** wider mouth, so that layer takes the biggest pieces the machine can sort; twelve destinations rather than eighteen.
- **Third, 18 bins:** half again as many destinations, which suits the small parts that dominate a pile; narrower mouth, and the mouth is what limits piece size on a layer, so anything too big has to be routed to a half-size layer.
- **Neither is cheaper.** A layer's bins are ~2.1 kg and 55 to 60 hours either way (from the catalog's own sliced weights and times), and per-layer capacity works out much the same, since the extra walls take room out of each bin rather than out of the layer.
- **The recommendation:** the parts calculator starts a fresh machine at two third-size layers and one half-size (`DEFAULT` in `parts-calculator/src/lib/layers.svelte.ts`), which is a sensible default: most LEGO is small, but you want one layer that can take the big pieces.

**One thing I could not source:** whether a half-size layer belongs at a particular height. Nothing in the repos, the docs or the Discord ties a size to a level, and the calculator labels layers 1..N with no top or bottom meaning, so the page says the level is the builder's choice too. If there is a house rule (bigger pieces lower down, say), tell me and it goes in.

The matching fix on `install-bins.md` is in #549, since that page only exists on that branch. This PR stands alone.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #552, opened 2026-09-05
docs: show the External bracket cover going on, hex frame step 9

basically-balloon:
Step 9 ("Fit the External bracket — covers") had no photo at all. barthel asked for one showing an external bracket with the cover attached.

Added the two photos Christoph took for the old external-bracket helper page, which was retired in `fbf8ed9b` and took its pictures with it:

- **Cover loose**, beside the open face of the collar it clips onto.
- **Cover pushed on**, collar closed.

Both are of a single bracket off the frame, so the caption says that the tube below it (External bracket — bottom vertical) and the extrusion through it (piece C) are fitted later, on the layer pages, not while building a hex frame.

Both URLs are live on the asset service (200), credits follow the `<cite>Photo: Christoph.</cite>` convention, and `validate_credits.py` passes for this page. christoph added to `contributors:`.


basically-balloon:
Closing at barthel's request: these photos are from the old assembly order and would confuse more than help. The step stays without an image until somebody shoots the cover going onto a hex frame.

PR #556, opened 2026-09-05
docs: chute screws are listed on the page that drives them

basically-balloon:
Every screw the chute section calls for now sits in the Needed Parts of the page where you actually drive it, and Chute core no longer repeats them. Before this, the funnel brackets, the door module and the layer connectors all pushed their mounting screws up onto the Chute core page, each with a line explaining that they "come out of the core's set and are not extra", while the Layer adapter board page carried its own four. So three pages listed screws they never mention driving, and one page listed 14 screws none of whose steps are on it.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1545873599217401998):** "Why are screws listed here under "Needed Parts" when the process of screwing components in place is described on the respective subpages?", scoped [in the thread](https://discord.com/channels/1430279849171222682/1545873599217401998/1545877500620578927) to "I mean the chute subpages only. All pages in chute section.", then [extended](https://discord.com/channels/1430279849171222682/1545873599217401998/1545882679029796904) to Chute core step 2 itself: "Now that we've moved the required screws to their respective subpages, there's no need to list them again in Step 2. That would only increase the maintenance effort when changes are made."

## What moved

- **Chute core**: drops 6 × `scr-m3-12-cs` and 8 × `scr-m3-8-cs`. It keeps the core and all 18 `hsi-m3`, because every one of those inserts is pressed in at its own step 1.
- **Funnel**: gains 4 × `scr-m3-12-cs`, the pair of brackets at its step 2.
- **Layer connectors**: gains 4 × `scr-m3-8-cs`, the pair at its step 1.
- **Door module**: 4 → 6 × `scr-m3-12-cs` and 16 → 20 × `scr-m3-8-cs`, adding the 4 through the bearing covers and the 2 through the servo bracket arms that its step 4 drives. Its "What each fastener is for" split now closes on the full totals instead of pointing elsewhere for the last row.
- **Layer adapter board**: unchanged. It was already doing this.

## Step 2 stops repeating them

Step 2 was listing every sub-assembly's screw count and length a second time, next to the link to the page that also lists them. That is the duplication barthel asked to remove, and it is now four links and nothing else.

The measurements behind each length were only written down in step 2, so rather than delete them they moved onto the page that uses each one: the funnel bracket's 8.16 mm wall onto Funnel, the layer connector strap's 3.78 mm onto Layer connectors, and the bearing cover's 5.00 mm plus the servo arms' inferred 12 mm onto Door module. Each number now sits beside the screw it explains, and no page points at another page for it. The servo bracket arms' 12 mm is still inferred by elimination rather than measured, because no STL exists for that joint, and the Door module page now says so at the point it is used.

Totals are conserved: 4 + 2 = 6 twelves and 4 + 4 = 8 eights, the same set the `chute` assembly declares in `parts-calculator/catalog/parts.json`. Nothing is added to or removed from the machine, so no catalog change and no regeneration.

This follows the pattern the rest of the assembly docs already use: the bin retainers took their own screws and T-nuts onto their own page in #547, and the hex frame split the same way.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/chute/chute-core.md
+++ b/docs/src/content/hardware/assembly/distribution/chute/chute-core.md
@@ -22,7 +22,7 @@ parts_needed:
 
 The chute is what steers a part into the right bin. Build one per layer, N for an N-layer machine. The [bottom interface]({{ '/hardware/assembly/distribution/bin-frame/bottom-interface/' | relative_url }}) doesn't add any extra chutes of its own, it's a mounting stage the bottommost chute sits on, bridged to it by a [layer connector]({{ '/hardware/assembly/distribution/chute/layer-connectors/' | relative_url }}).
 
-The fasteners and quantities are in the parts list above and are called out inline at each step.
+The heat inserts are in the parts list above. The screws that hold the four sub-assemblies on are on their own pages, listed where they are driven.
 
 {% include fastener-legend.html %}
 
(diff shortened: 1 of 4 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #557, opened 2026-09-05
docs: funnel page reordered, the snap documented, and builder-verified

basically-balloon:
Four changes to [/hardware/assembly/distribution/chute/funnel/](https://docs.basically.website/hardware/assembly/distribution/chute/funnel/), all of them from barthel reading the #556 preview.

**1. The size decision moves to step 2.**

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1545883365444419664): "The funnel brackets can be screwed on separately from the actual funnel or the bin divider. In my opinion, the decision to “Pick the funnel size for the layer” can be moved before “Hang the funnel.”"**

The brackets are the same left/right pair whichever size the layer ends up being, and they bolt into the chute core's M3 heat inserts, so nothing about fitting them waits on the size decision. Step 1 is now *Fit the funnel brackets to the chute core*, step 2 is *Pick the funnel size for the layer* (text unchanged from #551), step 3 is *Hang the funnel*. The size choice still has to be made before the funnel or that layer's bins are printed, which the step says in its first paragraph.

**2. Step 3 says how the funnel attaches.**

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1545883365444419664/1545883773969891388): "The funnel brackets have a kind of spring mechanism (the arm without holes) that allows the funnel itself to snap into place and stay secure."**

That step previously said the connection was not documented and asked whoever got there to report back. It now describes the snap: the plain arm is a sprung slide, the funnel has a rail down each side that runs into it, and the arm flexes and springs back to hold the funnel.

The catalog agrees with the report: `funnel-bracket-left` and `-right` in `parts-calculator/catalog/parts.json` are both sourced from `snap_in_funnel_bracket - <side>_funnel_slide.stl` and their internal note reads "Snap-in funnel bracket". Rendering the two published STLs shows the same thing: the bracket's plain arm carries a slot running its length with a stepped catch at the tip, and the funnel has a matching rail on each side.

Front matter changed with it at the time, following the shape `feeder/c-channel.md` uses. Change 4 below then takes the whole note off the page.

Nothing else on the page moves, and no fastener counts change.

**3. Where the brackets bolt onto the core.**

[...]

PR #558, opened 2026-09-05
docs: show where each chute sub-assembly bolts onto the core

basically-balloon:
Three of the four chute sub-assembly pages now show where their part bolts onto the chute core. **The funnel view is in #557 instead**, which already edits that page, [as barthel pointed out](https://discord.com/channels/1430279849171222682/1545873599217401998/1545886135170109460). None of them had an illustration of that joint: Funnel, Layer connectors and Layer adapter board carried only a placeholder for a photo nobody has taken, and Door module had renders of its own parts' heat inserts but nothing of it going onto the core.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1545873599217401998/1545885687625285834):** "Take a look at each subpage of the Chute Core section and see if there's an illustration showing how the individual elements are attached to the Chute Core. If there are no illustrations, find suitable photos or use one or more of the renderings (without the red circles) from the Chute Core page in the Preparation section, copy them to the relevant subpage, and mark with red circles where the parts should be screwed onto the Chute Core." Follows #556, which moved the screws themselves onto these pages. The funnel's own view is on #557.

## The views

Same renderer, camera and marker style as the three heat-insert views already on the [Chute core](https://docs.basically.website/hardware/assembly/distribution/chute/chute-core/) page, so the pages read as one set: light background, thin orange ring, no fill, camera about 18° off the face so a blind pocket shades as a crescent.

- **Layer connectors** — the 2 inserts one connector screws into, one near each end of a long side. Its mirror pair takes the second connector.
- **Door module** — a pair of views, because the two sides differ: the servo side has 4 (its bearing cover plus the two servo bracket arms), the other has 2.
- **Layer adapter board** — reuses the existing `chute-core-inserts-v3-rear` render unchanged. Those 4 circles were already exactly the board's 4 inserts, so nothing new was rendered for it.

No page gained a letter or colour key tying a circle to a part, because that was taken back out of the Chute core views in #406 at barthel's request.

The photo placeholders stay where they are: a render of the bare core is not a substitute for a photo of the finished joint, and those are still wanted.

## Positive control

The insert positions come from the measured map in the knowledge base rather than from eyeballing the render. Before cutting the four views, the same camera was run with **all 18 inserts circled**, and it reproduces the Chute core page's own counts exactly: 8 on one long side, 6 on the other, 4 on the rear face, every ring landing on a visible pocket. The subsets below are slices of that same list.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/chute/door-module.md
+++ b/docs/src/content/hardware/assembly/distribution/chute/door-module.md
@@ -170,6 +170,17 @@ Six of the core's 18 inserts are for this module, and the screws for them are in
 
 The cover screws are 8 mm because that reaches 3.00 mm into the core's blind 5.70 mm insert, measured off the STLs. The arms' 12 mm is inferred by elimination rather than measured, since no STL exists for that joint, so check the fit before committing.
 
+<div class="img-row">
+  <figure>
+    <img class="doc-figure" src="https://assets.basically.website/sorter-docs/chute-core-door-module-side-4-full-f61ab38ed984.png" alt="Render of the servo side of the chute core with four heat inserts circled in red">
+    <figcaption>The long side the servo sits on: two inserts for that side's bearing cover, two for the servo bracket arms. <cite>Render: Balloon.</cite></figcaption>
+  </figure>
+  <figure>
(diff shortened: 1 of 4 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #559, opened 2026-09-05
docs: chute door module servo bracket, corrected from a real build

basically-balloon:
Asked for by **barthel** in #balloon-does-docs: https://discord.com/channels/1430279849171222682/1536088194557419660/1545916365070671944

He corrected two things about the servo bracket from a real machine, and asked for photos and for the black box taken off the two servo adapter renders.

## What his correction changes

The page had the MG995 screwed into its housing through its own mounting ears, and the cover clipping on with no screws. Both are wrong:

- **The servo is not fastened.** It slides into the housing, the cover goes on over it, and the cover's two screws clamp the servo's mounting tabs.
- **The lower arm takes 2 x M3 x 8.**

Confirmed against the STLs, which also fill in the rest of the joint: the housing's 6 heat inserts are two on each of three faces, one screw per insert. Lower arm 2 x M3 x 8 (4.25 mm of arm), side arm 2 x M3 x 12 (8.40 mm), cover 2 x M3 x 12 (6.65 mm). The bracket's own totals are unchanged.

## The screw that falls out of the same measurement

Step 4 said the bracket bolts to the chute core with 2 x M3 x 12, flagged in the page's own words as "inferred by elimination rather than measured, since no STL exists for that joint". There is one, and the two arms are not the same thickness at the ear: the lower arm's is 8.40 mm and does take a 12, the side arm's is 5.00 mm, the same as the bearing covers', and takes an 8. A 12 there bottoms out in the core's blind 5.70 mm insert.

So one screw moves from the 12 column to the 8. The door module is now 5 x M3 x 12 and 21 x M3 x 8, and `chute` is stamped v2 in the catalog for the same change, `breaking: false`.

## Photos

**There are none of this module, and I could not make any.** Nobody in the Discord has photographed the door module or the servo bracket (searched the whole mirrored archive), and the MG995 video the page already embeds cannot be downloaded from CI-style hosts, so its frames are not available to cut. What is available without auth is the video's own poster frame, which happens to show the assembled servo adapter with its four countersunk screws in front of the servo in its bracket, so that is now a figure in step 4.

The rest is renders:

- The cover pulled off the housing, in step 2.
- The bracket assembled, in step 3.

An overview render of the whole module was in the first version of this branch and **has been removed**: barthel says it is wrong (https://discord.com/channels/1430279849171222682/1545916365070671944/1546028686539493446). The parts I placed it from are not all in the assembly frame I assumed they were.

## The servo adapter renders

The two loose half-views are replaced by an exploded view and an assembled one, colour-coded, following the Interface idler gear entry on the top-interface page. It now reads as a build order rather than two parts sitting next to each other: horn, then exploded. The assembled render came out again at barthel's ask (the exploded one carries it), the horn photo is mirrored so its splined sleeve points at the half it goes into, and the text says the sleeve seats in that half, which is the only way it fits. The preparation step is also reordered: the three heat-insert entries first, then the two sub-assemblies that go together there.

## The black box

Both servo adapter renders had a stub of the caption bar left in the bottom-left corner. Painted out and re-published; nothing else in either image changed.

## Later asks on the same page

- The overview render is gone (barthel: it is wrong).
- The two bearing-holder heat-insert entries are one entry with one render: the holders are a mirror pair, so the job is identical on both. The page now also says they are **handed** -- only a reflection maps one onto the other, no rotation does, so you print one of each.
- New preparation entry for seating the two 6704-2RS bearings, with a render. The seat is measured, and it closes exactly: the holder's bore is a **27.3 mm pocket 9.00 mm deep** onto a 24.4 mm step, the bearing is **4 mm** wide, and the bearing cover carries a **5.00 mm** annular boss. 4 + 5 = 9.
- New **step 4, hanging the door on its bearings and the race**, with a render. The order is not invented: both holder bores are exactly concentric with the door's Ø19.8 mm shaft, none of the four parts overlaps another in its own coordinates, and the race touches the holders at a 0.000 mm gap, so they are in one assembly frame and only fit one way. The shaft is captured once it is together, so the holders must go on before anything is bolted down. The door is also handed -- one shaft end is a ~12 mm hex spigot, and that is the servo side. Old step 4 is now step 5.
- **Fitting to the core is now two steps**: step 5 bolts the flap assembly on (the two bearing covers are the only thing holding it to the core), step 6 bolts the bracket on and couples the adapter to the door, and carries the video. The hex joint is measured too: the door's shaft ends in a **12.0 mm across-flats hex spigot** and the flap-side adapter has a **12.4 mm hex socket**, same orientation, so that is a confirmed pair rather than an assumption.
- Step 5 carries a **real photograph** of the flap assembly on the core, from 0:01 of the MG995 video. barthel screenshotted the frame (YouTube will not serve this box the video, so only four fixed frames are reachable here and none of them is that one); cropped here to the core, with the player chrome and the play overlay out of frame. The finished servo bracket is lying beside the core in the same frame, so **step 3 gets a real photo too**. barthel also confirmed the video is Spencer's, so the four captions on this page that said the filmer was not recorded now credit him, stills included.

The `chute` v2 version entry carries `"commit": null` on purpose. This branch was rebased onto main after #558, so the hash `stamp_versions.py` had written no longer exists, and since the board squashes, any hash written here is wrong by the time it lands. `layer` v4 already sits on main the same way; run `catalog/stamp_versions.py` after the merge.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #561, opened 2026-09-06
parts-calculator: the extrusions and the bottom layer's T-nuts go in the machine tree

basically-balloon:
**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1546026272176607232):** *"let's work on these outstanding items"*, pointing at [the deferred list](https://discord.com/channels/1430279849171222682/1545685109615759371/1545724899732234272) she asked me to [hold back from #545](https://discord.com/channels/1430279849171222682/1545685109615759371/1545724603417235526). Items 2 and 6 came out of her question about [why the extrusions read qty TBD](https://discord.com/channels/1430279849171222682/1545685109615759371/1545723554019020831).

Catalog data only. **No physical change: no STL, no geometry, no fastener length.** Three assemblies gain lines they should always have had, and two part ids are renamed.

## 1. The bottom layer never got its bin retainer T-nuts

`bottom-layer` v1 → v2, `breaking: false`.

#547 restored the 24 retainer T-nuts onto `layer` after they had fallen out of the machine entirely. #545 landed alongside it and created `bottom-layer` as a new node, without them. So the lowest bin layer has 12 retainers and 24 M5 × 16 passing through nothing.

**At 5 layers the machine reads 164 T-nuts; it should read 188.** Short by 24 at every layer count.

## 2. Pieces C, D, E and F go in the tree

The Hardware tab resolves a row from the assembly tree, then from `sheet_qty` (a hand count carried over from the old BOM spreadsheet), then gives up and prints **qty TBD**. C, E and F were in no assembly and had no hand count, so all three printed TBD. D had a hand count, so it was right by luck rather than derived.

Each piece now sits on the node that owns it:

- **C (Layer vertical support)** → `layer`, 6. A finished layer already carries the extrusion that spans up to the ring above it (`regular-layers.md`, "Joining two layers"). `layer` runs `non-bottom-layers` times, so 6 there is 6(n−1).
- **D (Foot extension)** → `bottom-layer`, 6. D replaces this layer's C and runs from the caster up through it. Its `sheet_qty` is removed, so the row now has one source instead of two.
- **E (Interface spoke, long)** and **F (Interface vertical support)** → `interface-bracket-mount`, 1 each. That node's own description already said it was *"bolted onto its vertical extrusion (piece F) and its spoke extrusion (piece E)"*; the lines were just never there. 6 brackets per interface, so 6 of each per machine.

[...]

PR #562, opened 2026-09-06
Count the bottom layer's 24 bin-retainer T-nuts

basically-balloon:
The lowest bin layer is the only frame in the machine whose bin-retainer T-nuts are not counted. This adds its 24.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1545685109615759371/1546032913164804168): "This should be fixed in the seperate bin retainer side."** Raised by BrickCycleAlice mid-build, [from Discord](https://discord.com/channels/1430279849171222682/1545685109615759371/1545719833893670922): "are the t-nuts listed as a count hardware here?" — they were not.

## Most of this was already fixed

[`layer` v4](https://github.com/basicallysource/sorter-v2/commits/main/parts-calculator/catalog/parts.json) restored `tnut-m5-2020` qty 24 on the regular layer node on 2026-09-05, while [#545](https://github.com/basicallysource/sorter-v2/pull/545) was still in review. `bottom-layer` was created by that PR and so never got the same line. The result on `main` today is `68 + 24(n − 1)` — every bin layer's retainer T-nuts counted **except the lowest one's**.

So this is 24 T-nuts, not the 120 I quoted in the thread before checking the current state.

## The change

One line on `bottom-layer`, `tnut-m5-2020` qty 24, in the same place `layer` carries its own — on the layer node alongside the 24 `scr-m5-16-shcs` that pass through them, which is where `layer` v4 decided they belong. Stamped v1 → v2, `breaking: false`; nothing physical changes.

Totals go to **`68 + 24n`** at every layer count, one T-nut per retainer screw:

| layers | before | after |
|---|---|---|
| 2 | 92 | 116 |
| 3 | 116 | 140 |
| 5 | 164 | **188** |
| 8 | 236 | 260 |

`check_connections`, `check_versioning` and `check_generated_pins` all pass.

## Two things found on the way, not fixed here

- **`layer` v4 was authored without minting a new uid.** VERSIONING.md says a new version is a new uid, and `stamp_versions.py` assumes it — run it today and it writes `uid: jog7` onto v3, which is already v2's uid, and `generate.py` then refuses the file with `duplicate uid(s): ['jog7']`. Worth a fresh uid on that entry.
- **`check_connections.py` cannot see this class of bug at all.** Four edges carry `method: "tnut"` and, until `layer` v4, none of their assemblies had a `tnut-m5-2020` line, so the assembly page rendered a "T-NUT" label over a nut nothing counted. An anchor-kind edge (`tnut`, `nut`, `insert`) whose assembly has no line for that anchor is always a bug and could be a check. That file is the PR gate, so it is not mine to change.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -8112,8 +8112,8 @@
     },
     {
       "id": "bottom-layer",
-      "uid": "sqsi",
-      "version": "1",
+      "uid": "7dq2",
+      "version": "2",
       "name": "Bottom distribution layer",
       "description": "The lowest bin layer. Identical to a distribution layer except that its frame carries foot covers instead of bracket covers and bottom verticals, and the bottom interface hangs underneath it.",
       "docs": "/hardware/assembly/distribution/bin-frame/bottom-two-layers/",
@@ -8137,6 +8137,10 @@
         {
           "part": "scr-m5-16-shcs",
           "qty": 36
+        },
+        {
+          "part": "tnut-m5-2020",
+          "qty": 24
         }
       ],
       "connections": [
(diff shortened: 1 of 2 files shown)


basically-balloon:
Closing as a duplicate of #561, which already contains this exact change (`bottom-layer` v1 → v2, `tnut-m5-2020` qty 24, 164 → 188 at 5 layers) alongside the extrusion work from the same deferred list.

I opened this without checking for an existing PR first, which is my mistake. Nothing here is lost: #561 has the same fix with the same reasoning.

The two findings in the body above are not in #561, so recording them here rather than losing them with the branch:

- **`layer` v4 was authored without minting a new uid.** `stamp_versions.py` then writes `uid: jog7` onto v3, which is already v2's uid, and `generate.py` refuses the file with `duplicate uid(s): ['jog7']`. Anyone running the stamper on `main` will hit this.
- **`check_connections.py` cannot see an uncounted anchor.** An edge with `method: "tnut"`, `"nut"` or `"insert"` whose assembly has no line for that anchor is always a bug, and the assembly page still renders the label as though the part were counted. That file is the PR gate, so not mine to change.

Issue #570, opened 2026-09-06
Catalog: retire one of the two chute bearing holders and print the survivor twice


basically-balloon:
Follow-through in the parts catalog for #569, which establishes that `bearing-holder-left` and `bearing-holder-right` are the same geometry (a proper rotation of exactly 180 degrees about the flange face normal maps one onto the other, max deviation 0.065 mm, determinant +1).

**Blocked on the decision in #569.** File this now so the catalog side is not lost once that decision is made.

## What the change is

- Drop one of the two ids from the `chute-bearing` assembly's `lines` and set the survivor's qty to 2.
- Rename the survivor. `bearing-holder-left` printed twice is a confusing row; it wants to be `bearing-holder` / "Bearing holder".
- Update `door-module.md`, whose `parts_needed` lists both ids at qty 1 each.

## Why it is not just an edit

- Changing an assembly's `lines` is a stamped revision under `scripts/check_versioning.py`: `chute-bearing` needs a `version` bump and a `versions[]` entry carrying `date`, `message` and the `breaking` bit. The bit is arguably `false` here, since an assembled unit built from one of each is physically identical to one built from two of the survivor, but that is a call for whoever stamps it.
- An id change is a second, larger question. Every one of the four existing "delete the unused X" entries in `changes` (#453, #454, #478, #536) is parked on the same point: a minted part id has to stay resolvable once it exists.
- `scripts/check_docs_refs.py` will warn (W1/W2) on the docs page until it is updated in the same change. Warnings, not failures.

## Meanwhile

Both ids are marked in the parts calculator as one change entry, `merge-bearing-holders`, so the calculator says so on either part page. Nothing is removed and nothing renders differently for a builder.

Asked by barthel in Discord: https://discord.com/channels/1430279849171222682/1546045981294985287/1546054980639785061

basically-balloon:
The marking half is #571 (draft). This issue stays for the actual retirement.

Issue #569, opened 2026-09-06
Chute door module: the two bearing holders are the same part, not a handed pair


basically-balloon:
`bearing-holder-left` and `bearing-holder-right` in the chute door module are the same geometry. Either one, printed twice, builds the whole bearing assembly, so the machine's printed-part count can drop by one.

## Measurement

Taken from the two master STLs pinned in `parts-calculator/catalog/parts.json`.

- A **proper rotation** of exactly **180.000 degrees** about the axis `(0, 0.90631, 0.42261)` maps the left holder onto the right one. That axis is the **flange face normal** `(0, 0.90592, 0.42345)` to within 0.05 degrees, so in the hand it is "lay the part flange down and spin it half a turn".
- The transform's **determinant is +1**. A reflection would be -1, so no handedness is involved.
- Max deviation after the rotation **0.065 mm**, mean **0.004 mm**, over all 3013 unique vertices, measured point-to-surface in both directions.
- The holder is **self-mirror-symmetric** about the plane containing the bore axis and the flange direction (plane normal `(0, -0.4226, 0.9063)` through the bore centre). An object with an internal mirror plane is achiral, which is why its mirror image is reachable by a rotation rather than only by a reflection.
- The only asymmetry anywhere in the part is **0.05 mm**: two of the three M3 insert pockets sit at 17.35 mm and 17.40 mm from the bore centre. That residual is the entire 0.065 mm figure above.

![Four renders of the bearing holder from one camera. Top left the left holder as printed, top right the same part turned 180 degrees about its flange, bottom left the right holder as printed which is the same image again, bottom right both holders with the bearing race between them showing the two bearing seats facing outward](https://assets.basically.website/sorter-parts/bearing-holders-one-part-full-2d451d042546.png)

Top right and bottom left of that figure are the same image: the left holder turned 180 degrees about the flange, and the right holder untouched, from one camera.

## What it changes and what it does not

- One printed part fewer in the machine, printed twice instead of two parts once each.
- **Nothing changes physically.** The two bearing seats still face outward, away from each other, with both flanges pointing inward to the race. That is an orientation at assembly, not a difference between the parts.
- Anyone who has already printed both loses nothing; the two are interchangeable today.

## What needs doing

- [ ] **CAD:** the part studio should carry one holder, not two.
- [ ] **Catalog:** retire one id and print the survivor twice. Tracked in #570.
[...]

PR #571, opened 2026-09-06
parts-calculator: mark the two chute bearing holders as one part

basically-balloon:
Asked by **barthel** in Discord: [could we use it as one part and reduce the printed-part count by one?](https://discord.com/channels/1430279849171222682/1546045981294985287/1546054980639785061)

Yes. `bearing-holder-left` and `bearing-holder-right` in the chute door module are the same geometry, so the machine can print one part twice instead of two parts once each.

**What this PR does is mark that, and nothing else.** No part is removed and no quantity changes; retiring an id is a stamped catalog revision plus a rename, which is #570, and the CAD side is #569.

## The measurement

From the two master STLs pinned in `catalog/parts.json`:

- A **proper rotation** of exactly **180.000 degrees** about the axis `(0, 0.90631, 0.42261)` maps the left holder onto the right one. That axis is the **flange face normal** to within 0.05 degrees, so in the hand it is "lay it flange down and spin it half a turn".
- **Determinant +1.** A reflection would be -1, so there is no handedness.
- Max deviation **0.065 mm**, mean **0.004 mm**, over all 3013 unique vertices, point-to-surface in both directions.
- The only asymmetry in either part is **0.05 mm**, between two of the three M3 insert pockets, at 17.35 mm and 17.40 mm from the bore centre.

Four renders of the bearing holder from one camera. Top left the left holder as printed, top right the same part turned 180 degrees about its flange, bottom left the right holder as printed which is the same image again, bottom right both holders with the bearing race between them showing the two bearing seats facing outward

Top right and bottom left are the same image: the left holder turned, and the right holder untouched, from one camera.

## The changes

- A **P1 `changes` entry with `condition: "broken"`**, `merge-bearing-holders`, targeting both parts, so the calculator shows the warning on either part page. Opened as P3 with no condition; barthel asked for P1 + broken, on the reading that the defect is the catalog asking for two different printed parts where one printed twice does the job. The description leads with that.
- An **`Interchangeable with`** attribute on each part.
- Both **descriptions** say it, since that is the line a builder actually reads.
- `internal_notes` on both carry the numbers and the reason the earlier "handed pair" claim was wrong.

`catalog.generated.json` regenerated with `generate.py --metadata-only` in the same commit. `check_generated_pins`, `check_versioning`, `check_connections` all pass locally; `check_docs_refs` reports its 23 pre-existing warnings and none of them is a bearing holder.

Closes nothing. Tracked in #569 and #570.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -413,8 +413,9 @@
     {
       "id": "merge-bearing-holders",
       "name": "Merge the two chute bearing holders into one part",
-      "priority": "P3",
-      "description": "The chute door module's two bearing holders are the same geometry, so the machine can print one part twice instead of two parts once each. Measured off the two master STLs: a proper rotation of exactly 180.000 degrees about the flange face normal (0, 0.906, 0.423) maps one onto the other with a maximum deviation of 0.065 mm and a mean of 0.004 mm, and the transform's determinant is +1, so no handedness is involved. The only asymmetry in either part is 0.05 mm, between two of the three M3 insert pockets, at 17.35 mm and 17.40 mm from the bore centre. The two ids are interchangeable as they stand and nothing about the assembly changes. Which id survives, and what a single holder should be called, are open questions. Tracked as sorter-v2 issue #569, with the catalog side in #570.",
+      "priority": "P1",
+      "condition": "broken",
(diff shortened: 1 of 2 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #520, opened 2026-09-02
parts-calculator: count the interface frame's run in the C quantity

basically-balloon:
Closes #519 if the count is accepted.

**Piece C goes from `6 × (layers − 2)` to `6 × (layers − 1)`.**

The cut list counted one vertical set per bin layer, with D covering two of them at the bottom and F covering the top. **Daddy-O's Bricks - Bill** assembled a 5-layer machine to the docs and came up 6 pieces short: the list called for 18 C and 24 were needed.

His marked-up photo walks the machine from the caster up, naming every segment and circling every break point:

Bill's 5-layer machine, each extrusion segment marked C or D with the break points circled

> Above this is another vertical layer support then the hexagon where the electronics go and the interface vertical supports drop down into the electronics hexagon's brackets

So the interface frame, the hex frame the PSU box, control board housing and Orange Pi mount bolt onto, is **its own ring above the top bin layer**, with a run of C between the two. Piece F only spans from that frame up into the Interface brackets on the top plate. The old formula assumed F did both jobs, so that run was never cut.

Counting rings for N = 5: five bin layers plus the interface frame is 6 rings and 5 runs between them. D takes the bottom run plus the foot, leaving 4 runs of C, which is 24.

## What changed

- `framing.ts`: C's `qtyFor` is now `6 * Math.max(0, n - 1)`, with a comment giving the ring/gap reasoning. Its `from` cell and `zeroNote` say where the extra set goes.
- `framing/+page.svelte`: a new Notes bullet, "The top layer needs a C too", explaining the interface-frame run alongside the existing D bullet.

Nothing about D changed. Bill's photo confirms it does exactly what `bottom-two-layers.md` says at the bottom, and Spencer's 1.5 × C length stands.

`npm run check` is 0 errors, 0 warnings. Built and read out of the prerendered page: C is ×12 at the default 3 layers, was ×6.

## Deliberately not in this PR

- **The 1-layer case.** At N = 1 C is still 0 and D's `n >= 2` guard still leaves a single-layer build with no feet, so it gets no vertical at all. That predates this change ([mollydrippy, 2026-07-25](https://discord.com/channels/1430279849171222682/1503528472818094220/1530497775551254699)) and is a design call.
- **The docs.** `top-interface.md` step 14 describes piece F running into "the top layer's bracket" as though the interface's hex frame and the top bin layer were one ring, and `hex-frame.md` says "N + 1" frames in its body while its lede enumerates N + 2. Both need the same ruling this PR is asking for, so they are left to #519.
[...]

PR #574, opened 2026-09-07
Framing cut list: piece J, drive it off the catalog, and retire pieces G, H and I

basically-balloon:
Three commits, all from one thread: the framing cut list had its own model of the machine, and it drifted.

## 1. Piece J: C-channel 1's support legs

`ext-2020-c1`, the 228 mm 2020 leg C-channel 1 stands on, has been in the parts catalog and the machine tree since 2026-09-02 but was never added to `framing.ts`, so the framing tab neither listed the three legs nor packed them into bars. Added as piece **J**, 3 per machine.

C2 and C3 use printed legs (148 mm and 68 mm) and C4 has none, so J is the only 2020 extrusion outside the frame.

## 2. The cut list now reads the catalog

`framing.ts` used to hold nine hardcoded pieces with a `qtyFor(n)` function each. That is a second model of the machine, and it is why C was 6 short at the top for weeks (#520) and why J was missing at all. Now:

- **The row set comes from the catalog** — every `kind: cots` part with `cots.type === 'extrusion-2020'`. A new extrusion piece appears in the cut plan automatically, which is the bug class this fixes.
- **The quantities come from the machine tree**, through the same `resolveHardwareTotals` resolver the Hardware tab counts screws with. `qtyFor` is deleted.
- **The lengths come from `cots.length_mm`**, and `framing.ts` keeps only presentation: marker colour, frame/feeder group, and the sentence saying where a quantity comes from. A part with no style entry still renders.
- **Two new fields ride in the existing `cots` block**, so `catalog/generate.py` needed no change: `cots.letters` and `cots.cad_length_mm` (only on C, D, E and F, the four trimmed 6 mm short of CAD).

**Piece I is gone.** The 267 mm "Bulk bucket support" had no catalog part and belonged to the prototype foot/leg/leg-extension stand that the layout guide, support legs and dovetail adapters replaced on 2026-09-02. zed0 confirmed none of it was ever used on a machine on the website. Those three parts and their three assemblies already carry no quantities and are unreachable from `machine`, so retirement was already done; the missing step was the tracked delete entry, now filed as P4 `delete-retired-c-channel-supports` alongside the other nine. `VERSIONING.md` does not allow removing an entry outright.

## 3. Pieces G and H retired

A and G were the same 320 mm extrusion, B and H the same 158 mm one, split only by which ring they sat on. Every ring is the same hex frame built to the same guide, interface frame included, so the split named nothing real: they were already one catalog part with one uid, and `hex-frame.md` already opened "This guide builds one hexagonal frame".

- **`Hex outer horizontal (A)`** and **`Hex spoke (B)`**, replacing "Outer horizontal / Horizontal interface frame" and "Spoke / Interface spoke, short", which were double-barrelled only because of the split.
- **30 occurrences across 8 docs pages**, every one already written "A/G" or "B/H" rather than a bare G or H. **No page has ever said "piece G" or "piece H", and neither has anybody in the Discord**: A/G 31 mentions, B/H 33, piece F 53, piece G and piece H zero. Nothing anybody cites is stranded.
- **The letters keep their holes**: A, B, C, D, E, F, J. Renumbering to close the gap would invalidate every existing mention of the pieces after them, and F alone has 53 in the Discord.
- Nothing physical changes: same part, same uid, same length, same quantity. A rename, not a revision, so no version stamp. Version history `message` fields keep the old names, since they record what was true when written.
- Also corrects alt text shared by five pages that called the B spokes printed. They are 2020 extrusion; the printed parts in the inner ring are the crossbeams and the 90 degree brackets.

## Numbers

Quantities are unchanged at 2 layers and up, and unaffected by commit 3. At 3 layers: A 24, F 6, E 6, D 6, B 24, C 12, J 3, 20 bars, 18.46 m of material (19 bars before piece J existed).

**At 1 layer the tree gives 6 foot extensions where the old table gave 0.** C is still 0 there, and its note now says so honestly: the run from the bottom layer up to the interface frame is not modelled. That was the open gap mollydrippy hit in 2026-07 and it is now visible rather than silent.

Verified at 1, 2 and 3 layers by building and reading the prerendered page (the layer count lives in `localStorage`, so a preview only ever shows the default).

`npm run check` 0 errors on 4388 files, both sites build clean, and `check_generated_pins.py`, `check_versioning.py` and `check_connections.py` all pass. `validate_images.py` validates 202 assets; `validate_frontmatter.py` and `validate_credits.py` report only their known pre-existing findings on pages this branch does not touch.

## 4. The explainer image

The piece letters are drawn into the framing page's explainer render, so
retiring G and H left the picture arguing with the table under it. Requested by
zed0 (@zed0) [from Discord](https://discord.com/channels/1430279849171222682/1546451479613808772/1546527951628009522): "are you able to modify the image with the labels to update the G label to A?", and
followed up by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1546451479613808772/1546571488981815458).

The G label now reads A and the H label reads B, each glyph lifted pixel for
pixel out of the A and B labels already in the same render, so the type matches
exactly. Nothing else in the image changes: the left column reads F, B, A, B, A,
B, A top to bottom, which is the rename's whole point, the same two pieces on
every ring. D and J are still not in it (D is not in the CAD, as the caption
says).

The URL has `full` in it twice. The asset service publishes an image at
`<name>-full-<hash>`, and my first upload of this file was itself named
`...-full.png`, so the correctly-named upload's published address is occupied by
that first upload's private original and answers 403. The address in this commit
is the public, permanent copy of exactly these bytes.


basically-balloon:
zed0 marked this ready for review in Discord.

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #393, opened 2026-08-23
feat(order-files): add production and procurement files for Distribution Board V1.3 and Adapter Board V1.0

barthel:
Adds `electronics/order_files/` with Gerber files, BOMs, CPLs and centroid/pick-and-place files for two PCB assemblies:

- **Distribution Board V1.3** (schematic date 2026-06-01) — 10 units
- **Adapter Board V1.0** (schematic date 2026-04-01) — 50 units

Files are provided in four manufacturer-specific formats:

| Manufacturer | Region | Format |
|---|---|---|
| JLCPCB | China | Gerber ZIP, BOM with LCSC part numbers, CPL |
| PCBWay | China | Gerber ZIP, BOM, CPL |
| Beta LAYOUT | Europe (DE) | Gerber ZIP, BOM (CSV + Excel template), centroid, pick-and-place |
| Eurocircuits | Europe (BE/HU) | Gerber ZIP, BOM, centroid |

A README documents the directory structure, manufacturer overview, board revisions, and DNP component guidance for JLCPCB orders (D1, U3, U4, U5 have no LCSC equivalents and must be sourced separately).

PR #577, opened 2026-09-07
docs(order-files): D1 is DNP for all manufacturers, not just JLCPCB

barthel:
D1 (MBR120VLSFT1G Schottky diode) was assumed to be a JLCPCB-specific DNP workaround, with PCBWay, Beta LAYOUT and Eurocircuits populating it normally. Confirmed with the project's electrical engineer that D1 should be DNP everywhere.

- Updates the order-files README: D1 moved out of the JLCPCB-only DNP section into a manufacturer-agnostic DNP note; U3/U4/U5 remain JLCPCB-only DNP.
- Marks D1 as DNP in the PCBWay, Beta LAYOUT and Eurocircuits BOM files (matching the existing JLCPCB BOM, which already had D1 correctly marked).
- No centroid/CPL changes needed: D1 was never present in the PCBWay/Beta LAYOUT/Eurocircuits placement files, consistent with it now being DNP.


barthel:
Superseded: merged directly into feat/order-files-procurement, see #393.

PR #584, opened 2026-09-07
parts-calculator: cable cage bracket (cable mount) v2, barthel's fix for the ribbon clamp mismatch (#403)

### LINKED ISSUE #403: Ribbon cable clamp sits crooked: foot stands 5.00 mm proud vs 1.30 mm at the screw end, and the cage bracket has no slot for it
The ribbon cable clamp (`ribbon-cable-clamp`) cannot sit flat against the cable cage bracket it bolts to: the foot at the far end from the screw stands **5.00 mm** proud of the clamping face while the pad at the screw end is only **1.30 mm**, and the bracket's face is flat, so the clamp pivots on its single M3 and never closes on the cable.

Three builders hit this independently:

- **barthel** (`@uwe_barthel`), [2026-08-17](https://discord.com/channels/1430279849171222682/1503528472818094220/1538987500285071401): *"Shouldn't the spacer shown at the bottom in the illustration (actually the top end) be exactly the same size as the one at the top near the screw hole (actually the bottom end)? When I screw the clamp on like this, it doesn't clamp the cable, and the part sits crooked."*
- **zed0** (`@zed0`), [2026-08-17](https://discord.com/channels/1430279849171222682/1503528472818094220/1539012427847303268): *"This was the same for me. I cut the other end down to size with a knife"*
- **Jon ⚡** (`@effreek`), [2026-08-18](https://discord.com/channels/1430279849171222682/1503528472818094220/1539253997972754492): *"I made a slot in the cage bracket to accept the tab. I think that was the original design intent. A way to 'lock' the position since its only one screw."* and [here](https://discord.com/channels/1430279849171222682/1503528472818094220/1539257522957848677): *"I did what `@zed0` did... except I cut the bracket not the clamp. `@chriskiepfer0591` do you have the 3D file that meshes with the clamp? I think it needs merged into the main files so it can be referenced in the parts calculator."*

How it sits

## Measured

From the published STLs, both of which are positioned in the same assembly frame (`ribbon-cable-clamp-nf0n-39b7e6d0.stl`, `interface-cage-bracket-cable-cy58-081da7a9.stl`). Coordinates below are that frame, X across the joint, Z up.

The two parts really are a mating pair: the bracket's M3 heat-insert bore (Ø4.2 × 4.0 mm deep) is centred at Y 197.6 / Z 240.1, and the clamp's Ø3.5 clearance hole is centred at Y 197.6 / Z 239.7.

- **Bracket outer face:** one flat plane at X = −15.00 over Z 210.5 … 258.9. The insert bore is the only feature in it: no slot, no pocket.
- **Clamp clamping face:** X = −16.20, so a **1.20 mm** cable gap when the clamp seats.
- **Screw-end pad:** face at X = −14.90 → **1.30 mm** proud. Lands on the bracket face (0.10 mm interference, i.e. flush).
- **Far-end foot:** face at X = −11.20 → **5.00 mm** proud, which is **3.80 mm past** the bracket's outer face.
- The foot occupies Z 208.5 … 211.4, and the bracket's underside in that region is Z = 210.0, so foot and bracket overlap by **~1.4 mm** vertically. The foot can neither pass under the bracket nor lie against it, so it levers the clamp off the face and the whole part rotates about the single screw ~30 mm above.

Difference between the two standoffs: **3.70 mm**. That is the tilt.

## Two ways to close it, and they need a decision

1. **Slot the bracket** (Jon's fix, and probably the original intent). A notch in the bracket's bottom edge, ≥3.80 mm deep from the outer face, ≥1.5 mm tall, ~24 mm wide (the foot spans Y 182.9 … 206.6). Keeps the foot as an anti-rotation key, which is worth having when one M3 is the only fastener.
2. **Cut the foot back to 1.30 mm** so both standoffs match (barthel's proposal, zed0's knife fix). One dimension, but the clamp is then free to rotate on the screw.

My read is (1): the foot is too deliberate to be anything but a locating feature, and the lock is the reason it exists. Either way both parts change together, and the mating bracket has to land in the main CAD, not in one person's local copy. ChrisKiepfer (`@chriskiepfer0591`), Jon asked whether you already have that file.

## Also worth fixing while it's open

`ribbon-cable-clamp`, `interface-cage-bracket` and `interface-cage-bracket-cable` have no `onshape` link in `parts-calculator/catalog/parts.json`, so there is no way to get from the calculator to the source document. `cable-clamp-inner` / `cable-clamp-outer` do.

## Until it is fixed

Trim the foot down flush with a knife (zed0), or cut the matching notch in the bracket (Jon). Both work; the second one keeps the clamp from rotating.

---
_Filed by [balloon](https://discord.com/channels/1430279849171222682/1503528472818094220/1541293768735592489) at barthel's request._

spencerhhubert:
The new version of the cable-mount cage bracket barthel posted for #403, replacing v1 in the catalog.

- `interface-cage-bracket-cable` is now **v2**, uid `nu2n`, master `0dad6ad3b408` (already on the asset service). The underside near the heat-insert boss rises ~2 mm (Z 210 → 212) and the mounting foot widens from ±15 to ±22 mm in X, measured against the v1 master in its own export frame. Bolt pattern and insert bore are unchanged, so an old print still bolts in the same place. Not yet confirmed to close #403.
- v1 (uid `cy58`) is recorded as the superseded version with its pinned hash, so the uid on existing prints keeps resolving.
- `stamp_versions.py` also stamped two entries that were still pending on main (chute v2, chute-stepper-gearing v2).
- Regenerated with the local slicer: 16.0 g, four stamp faces, new bundles. Nothing else in the catalog moved.

This is the change balloon started this morning and could not finish; the reason is fixed in #582.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -1586,18 +1586,29 @@
     },
     {
       "id": "interface-cage-bracket-cable",
-      "uid": "cy58",
+      "uid": "nu2n",
       "name": "Cable cage bracket (cable mount)",
       "description": "Cable cage bracket with the cable mount. 1 needed.",
-      "internal_notes": "1 per interface; the cage bracket variant with a cable mount. Frame color.",
-      "version": "1",
+      "internal_notes": "1 per interface; the cage bracket variant with a cable mount. Frame color. v2 (2026-09-07): barthel's proposed geometry for the ribbon-cable-clamp mismatch (sorter-v2 #403) -- the underside near the heat-insert boss moved from Z 210.0 to Z 212.0 and the mounting foot widened from X +-15 to X +-22, measured against the v1 master STL in this part's own export frame. Not confirmed to close #403; that's still tracked separately in the P1 changes entry.",
+      "version": "2",
       "created_at": "2026-06-27",
-      "updated_at": "2026-06-27",
+      "updated_at": "2026-09-07",
+      "versions": [
+        {
+          "version": "2",
+          "date": "2026-09-07",
+          "message": "barthel's proposed fix for the ribbon-cable-clamp foot/lip mismatch (#403): the underside near the heat-insert boss raised ~2mm (Z 210.0 -> 212.0) and the mounting foot widened from +-15 to +-22 in X. Mounting bolt pattern and insert bore unchanged, so an old print still bolts in the same place.",
+          "breaking": false,
+          "commit": null,
+          "onshape_version": "https://cad.onshape.com/documents/36d65973fe08f685f790dc8b/v/b0b87904579c8becc296a460/e/5e3340c2e9170c356be9cfb9"
+        }
+      ],
       "quantities": {
         "interface": 1
       },
       "assembly": "cable-cage",
-      "stl_hash": "081da7a95e53340f4f27219a224c0015905f176707c84f82c2f69bd72da85530",
+      "stl_hash": "0dad6ad3b408d51771212d81c465973dee0d4f0f41be46ac8f0ba92fbc1e7567",
+      "onshape_version": "https://cad.onshape.com/documents/36d65973fe08f685f790dc8b/v/b0b87904579c8becc296a460/e/5e3340c2e9170c356be9cfb9",
       "color": {
         "role": "frame"
       },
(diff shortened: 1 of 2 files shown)


spencerhhubert:
Closing so balloon can finish this one itself in the original thread, now that #582, #585 and #587 are on main. Branch kept; reopen if its attempt goes sideways.

PR #590, opened 2026-09-07
feeder: B1 belt-feeder topology (cleated conveyor replaces C1+C2)

mneuhaus:
## What this is

The first of a series that brings the mature parts of `b1-belt-feeder` to main, lane by lane. This lane is the machine topology itself: a new setup `belt_feeder` pairs a cleated conveyor on the C1 rotor port with the existing C3 pulse-perception exit metering and the C4 classification channel. C2 has no motor and no camera in this topology.

## Contents

- `FeederMode.BELT_REV01` (`subsystems/feeder/belt/`): the belt runs at a continuous speed throttled by C3's perception fill level, stops on hold and while the chute moves, and raises its own `belt_feeder_stalled` incident when the jam window passes without a new piece reaching C3. C3's stuck watchdog treats the belt as its upstream.
- Setup registry entry, required steppers (`c_channel_1` as the belt, `c_channel_3`), `belt_stepper` alias, capability-based C4 rotor profile (`uses_classification_channel` instead of a key check, which had left `belt_feeder` at the slow profile), `[feeder_belt]` tuning section with `GET/POST /api/tuning/feeder-belt` and a belt status endpoint, `machine.example.toml` section.
- Frontend: "B1 Belt Feeder" hardware station, feeder-belt tuning page, dashboard without the C2 tile (C3 over C4), belt-stall incident card, setup camera roles without C2.
- Fixes carried by the same lane: OBSBOT auto-exposure detection for non-standard V4L2 menus, `[perception.conf_thresholds]` per-channel detector confidence, C4 rev01 forward-only converge with a no-detection grace on the walled platter (`precise_blind_grace_ms`), tap-to-open model activation menu for the touch tablet.

## What changes for machines without a belt

Nothing selects `belt_feeder` unless the setup does. The only defaults that change for every setup are the rev01 single-piece converge (forward-only, 1200 ms blind grace instead of hunting) and the model activation menu (tap instead of hover).

## Included after review

The second commit (after the Codex review, see the comment below) adds the stale-perception stop, the stall-latch stop with the watch-only StallGuard load monitor, the bounded stuck-watchdog nudge, the belt_feeder / belt_rev01 pairing and turns the rev01 blind grace off by default.

## Not included

Every later lane (Waveshare door safeguard, C4 two-piece hardening, C3 exit pacing, set extraction). Those follow as separate PRs on top of this one.

## How it was ported

Squash of `b243aa49 … 6de68166` from `b1-belt-feeder`, in production on the B1 machine since July. Commit cherry-picks no longer apply (main moved underneath since July), so 28 files were taken from the branch directly and 9 shared files (`irl/config.py`, `main.py`, `toml_config.py`, `vision_manager.py`, `rev01_config.py`, dashboard, setup page, header, models section) were hand-ported to main's current structure.

## Tests

- `tests/test_belt_feeder.py` (14) and the dashboard config test pass.
- Full backend suite: the same 20 failures as on current main, nothing new (those 20 are what #523 fixes).
- `pnpm check`: no errors in the touched files.

### DIFF
--- a/software/machine.example.toml
+++ b/software/machine.example.toml
@@ -153,3 +153,18 @@ c_channel_3 = 4
 
 [stepper_direction_inverts]
 carousel = true
+
+# B1 belt feeder (machine_setup = "belt_feeder" pairs [feeder] mode = "belt_rev01").
+# Every key is tunable live under Settings > Tuning > B1 Belt Feeder.
+# [feeder_belt]
+# belt_speed_usteps_per_s = 2000   # base speed while C3 wants pieces
+# forward_direction_sign = 1       # -1 if the belt motor is wired the other way
+# c3_full_speed_pieces = 1         # at/below this many pieces in C3: full speed
+# c3_stop_pieces = 3               # at/above: belt stops (linear ramp in between)
+# speed_update_interval_ms = 250
+# jam_timeout_s = 45.0             # no new C3 piece for this long -> incident; 0 = off
+# enable_belt = true
+
+# Per-channel detector confidence overrides (read once at backend start).
+# [perception.conf_thresholds]
+# 4 = 0.25
--- a/software/sorter/backend/global_config.py
+++ b/software/sorter/backend/global_config.py
@@ -111,6 +111,9 @@ class GlobalConfig:
         # main.py after camera startup. Not an env-toggle: the mode config
         # determines whether this is non-None.
         self.perception_service = None
+        # B1 belt feeder live status: one dict BeltFeeding updates in place,
+        # read by the tuning router. None until a BELT_REV01 feeder has run.
+        self.belt_feeder_status: "dict | None" = None
 
         # Standalone training-image grabber. Runs regardless of machine mode;
         # set in main.py after camera startup. See sample_collector.py.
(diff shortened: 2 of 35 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up** — verdict was "needs rework"; addressed in the second commit:

- **Frozen C3 camera leaves the belt running blind** (`_c3_piece_count` never checked the state age) — fixed: the C3 state must be younger than `[feeder_belt] perception_stale_s` (2 s), otherwise the belt stops with reason `stale_perception`.
- **No stop on a latched stall** — fixed: the stall-latch stop and the StallGuard load watch from `b1-belt-feeder` (6f50d2e6) are included; the load watch stays watch-only (threshold 0) until a value has been read off the machine.
- **Watchdog nudges unbounded and restarting the jam clock** — fixed: a nudge is one run of `nudge_run_ms` (1.5 s) at base speed, `nudge_max_attempts` (1) per stall, refused on a latched stall, jam epoch untouched; after that the watchdog escalates to the operator. This is the failure that sheared the belt shaft on 2026-09-06.
- **rev01 blind grace changed the default for every machine** — fixed: `precise_blind_grace_ms` defaults to 0 (legacy wait for the rotate timeout); the walled-platter machine sets 1200 explicitly.
- **`belt_feeder` / `belt_rev01` not enforced as a pair** — fixed at config load (`mkIRLConfig` refuses a mismatch), at the feeder-mode endpoint (409) and in the settings page (only the matching mode is offered).

Tests: 22 belt tests (stale perception, stall latch, bounded nudge, mode pairing), full suite unchanged against main (the same 20 pre-existing failures), `pnpm check` no new errors.

PR #591, opened 2026-09-07
servos: Waveshare bus stabilisation, door calibration and the door safeguard

mneuhaus:
## What this is

Second lane from `b1-belt-feeder`, stacked on #590 (belt-feeder topology). Everything here concerns the Waveshare serial-bus door servos and the distribution steps that depend on them. The PCA9685 backend is untouched.

## Contents

**Bus and servo driver**
- Stabilised bus stack (ported from the `sorthive` branch, 04813910): serial identity per board, read timeouts, an inventory that survives a busy bus; a board without `GET_OBSERVABILITY` no longer fails hardware init.
- Door torque is released after every move (the doors sat at 74 °C with torque never released); the supervisor no longer deadlocks on SIGTERM.
- Calibration finds the door's free travel instead of the crushed stop, with probes larger than the stiction band, settles until the horn stops creeping, and reports progress relative to the probe. 500 ms door moves (300 ms tripped the overload protection of a stiff door).
- `[servo] max_torque_percent`, `move_time_ms`; "Install position" moves a servo to mid travel and resets its calibration.

**Door safeguard in distribution**
- The target door is verified to have arrived before a piece is dispensed: a stalled or yielding flap reports "stopped" by time but sits far from its stop and lets pieces through to the wrong row. One re-close, then the chute-jam alert instead of a wrong-row drop. An unknown door position blocks dispensing.
- The target door stays powered from the drop until `[distribution] chute_settle_ms` elapsed, so the piece meets a held flap.
- `[chute] max_angle_deg` is the machine's travel limit.

**Settings UI**
- The Storage Layers settings configure the Waveshare bus again; the PCA-only storage components (`StorageLayerSettingsSection`, `ServoSettingsSection`, `storage/*`) are removed, nothing imports them any more. Bins per layer are editable there, and the page says why the inventory is empty while the machine runs.
- Servo settings save keeps the `[servo]` keys the form does not own.

## What changes for machines without Waveshare doors

The distribution door check and the door hold only act when the layer servo exposes `target_reached` / `hold`; PCA9685 servos do not, so those paths return early. The settings page restructuring applies to every machine.

## Origin

`187606b8 … f3a50b73` on `b1-belt-feeder` (2026-09-02/04), in use on the B1 machine since. Eleven test files cover the servo driver, calibration, inventory, door check, chute travel limit and the distribution flow.

## Tests

- Full backend suite: the same 20 pre-existing failures as on main (what #523 fixes), nothing new.
- `pnpm check`: no new errors (the one error in `ServoLayerCalibrator.svelte` is on main today, line 609).


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up** — verdict was "needs rework"; all five findings addressed in the second commit:

- **Door powered only after the drop** (the hold happened in SENDING, after classification had flung the piece) — fixed: positioning holds the verified door before READY opens the gate and hands it over via `shared.held_door`; sending adopts it and releases after the settle.
- **A hold could outlive its drop** (an incident skips distribution without cleanup) — fixed three ways: the servo's `hold()` carries a 30 s cap after which it releases torque on its own, positioning releases on an abort, and the distribution state machine releases on pause/incident cleanup.
- **A failed re-close counted as verified** — fixed: it raises the chute jam alert and never passes.
- **Blocking bus reads in the coordinator thread** — fixed: one read per tick, three unreadable ticks count as unknown, no sleeps.
- **Outer "Save servo setup" wiped PCA9685 speed overrides** — fixed: the outer save renders for the Waveshare form only; the PCA calibrator keeps its own save.

Tests: door check (stateful reads, failed re-close, hold-before-READY), servo hold cap, sending adopting a held door; full suite unchanged against main.

basically-balloon:
Heads up on merge order: this PR's base is `pr/belt-feeder-topology`, the head branch of #590, not `main`.

Because `sorter-v2` squash-merges, the moment #590 lands that branch is dead. GitHub normally retargets a stacked PR to `main` when its base merges, but that is asynchronous, and if this PR takes its second approval inside that window it merges into the dead branch instead: it will say "merged" on the board and reach nothing. That is exactly what happened to #696 and #698 this morning (90 seconds after their base landed), and #708 exists to put them back.

So after #590 merges, retarget this one to `main` before anyone approves it, and check `git log --oneline origin/main..origin/pr/waveshare-door-safeguard` afterwards: if it still lists #590's commits, the branch needs rebuilding on `main` and retargeting alone is not enough.

Same applies to #592, which is stacked on this one.

PR #592, opened 2026-09-07
sorter: robustness fixes from the B1 runs

mneuhaus:
## What this is

Third lane from `b1-belt-feeder`, stacked on #591. Small, independent fixes found during the September runs; none of them is belt-specific.

## Contents

- **incidents**: rows left active by a previous backend process are closed at startup instead of blocking the new run.
- **main**: the traceback of a main-loop exception is logged before shutdown.
- **hardware**: every stepper is stopped first at recovery start and at shutdown.
- **bus**: 250 ms read timeout on the MCU bus, so a silent board cannot hang the coordinator.
- **bins**: `/api/bins/move-to` closes the target layer's door like production does.
- **cameras**: zoom is an adjustable USB camera control; auto-exposure detection handles non-standard V4L2 menus (OBSBOT).
- **perception**: the region lookup table is keyed by zone sections, not by channel id; a piece overlapping the drop arc is a DROP piece for the handler.
- **c4**: the exit-release test parameters are range-checked in one shared validator (`exitReleaseTestParams`, up to 24000 µsteps/s and 200000 µsteps/s², the values the stall-clear ladder in #529 needs), used by both the review endpoint and the router.
- **ui**: no camera-drift dialog while the operator previews unsaved values.
- **tests**: `irl` is imported before `machine_platform` in the stepper safety test (import cycle).

## Origin

`d124d773 … 9fc547b0` on `b1-belt-feeder` (2026-09-04/05).

## Tests

Full backend suite: the same 20 pre-existing failures as on main (what #523 fixes), nothing new. `pnpm check`: no new errors.

### DIFF
--- a/software/sorter/backend/hardware/bus.py
+++ b/software/sorter/backend/hardware/bus.py
@@ -89,13 +89,15 @@ class MCUBusError(Exception):
 class MCUBus:
     """Class for communicating with the MCU over a serial bus using a custom protocol."""
 
-    def __init__(self, port: str, baudrate: int = 576000, timeout: float = 0.1):
+    def __init__(self, port: str, baudrate: int = 576000, timeout: float = 0.25):
         """Initialize the MCUBus with the given serial port parameters.
 
         Args:
             port: The serial port to use (e.g. "/dev/ttyUSB0")
             baudrate: The baud rate for the serial communication (default 576000)
-            timeout: The read timeout in seconds (default 0.01s = 10ms)
+            timeout: The read timeout in seconds (default 0.25 s — a reply
+                that has started may stall while the board steps a pulse;
+                100 ms cut GET_STALL_STATUS replies mid-frame ~100x/hour).
         """
 
         self._serial = serial.Serial(port, baudrate=baudrate, timeout=timeout)
--- a/software/sorter/backend/incident_records.py
+++ b/software/sorter/backend/incident_records.py
@@ -86,6 +86,17 @@ def _ensureInitialized() -> None:
             )
             conn.execute("CREATE INDEX IF NOT EXISTS idx_incidents_kind ON incidents(kind)")
             conn.execute("CREATE INDEX IF NOT EXISTS idx_incidents_status ON incidents(status)")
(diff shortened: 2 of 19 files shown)


mneuhaus:
**Codex review (gpt-5.6, 2026-09-07) and follow-up** — verdict was "needs rework"; addressed in the second commit, two remarks left as notes:

- **Recovery stop claimed success without a stop** (`stopStepper` ignored the result, called a `halt()` the stepper does not have, and was suppressed on a software-disabled motor) — fixed: forced past the software disable, an unacknowledged stop cuts the driver, and only a confirmed stop is reported.
- **Drop gate and handler used different drop predicates** — fixed: both use the interior-grid overlap; a test asserts they agree for straddling and interior boxes.
- **Stale incidents closed lazily** — fixed: `main()` closes them first thing.
- **`/api/bins/move-to` moved the chute despite a failed door command** — fixed: 409, chute stays.
- **Zoom `-1` sentinel** — fixed: zoom joins the filtered properties.
- **Drift dialog race with an in-flight poll** — fixed: the pause is re-read after the fetch.
- *Note:* the recovery stop runs after `mkIRLInterface()` because the stepper handles only exist after discovery; stopping earlier is not possible with this interface. *Note:* the raised exit-release test bounds (24000 µsteps/s, 200000 µsteps/s²) are the values the operator-proven stall-clear kick uses on the B1 platter; the lower maintenance caps in the April notes were for the older rotor.

basically-balloon:
Same merge-order note as #591 (see the comment there): this PR's base is `pr/waveshare-door-safeguard`, the head branch of #591, not `main`.

`sorter-v2` squash-merges, so the instant #591 lands that branch is dead, and a stacked PR approved inside GitHub's retarget window merges into the dead branch and reaches `main` with nothing. #696 and #698 did precisely that this morning; #708 is the cleanup.

Retarget this one to `main` once #591 has landed, before it picks up its second approval, and check `git log --oneline origin/main..origin/pr/robustness-fixes` afterwards: if #590's or #591's commits are still listed, the branch needs rebuilding on `main` rather than just retargeting.

PR #602, opened 2026-09-08
popover: one floating layer for every popover and tooltip, in both apps

basically-balloon:
**Requested by Spencer (@spencerhubert) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1546693223143972904):** *"these popovers are insanely annoying. it seems like if you dont move your mouse just the perfect way they dont persist whiel your mouse is hovering on the popover itself. it seems they also ahve some z-index nonsense going on where things can overlap them. see the pic where it clips and turns this component into a scrollable thing. can you just make like a way nicer family of popover components that dont suck and use them everywhere across the parts calc and the docs."*

**Follow-up from barthel (@uwe_barthel) [in the same thread](https://discord.com/channels/1430279849171222682/1546693223143972904/1546729604088336414):** *"The new pop-ups cannot be scrolled (Firefox on macOS)"*. Fixed in `c2dc58f`, see the last section.

## What was wrong

Three separate bugs, all from the same decision: the panel was absolutely positioned inside its trigger's own box.

**It got clipped.** The parts table scrolls horizontally (`.pl-scroll`), so a panel opened from a row was cut off at the table's edge and turned the table into a scroller. That is the second screenshot in the request. `ChangeStatus.svelte` already carried a comment working around this by forcing every panel to open leftwards.

**Things overlapped it.** Any ancestor that starts a stacking context painted over the panel however high its `z-index` went.

**It closed under the cursor.** The panel sat `top-6` (24px) below the trigger with nothing in between, and open/closed was driven by `mouseenter`/`mouseleave` on the trigger's wrapper. Moving the pointer down towards the panel took it through that 24px of nothing, which is a `mouseleave`, so it shut. There was no close delay either, so a diagonal route out of the badge closed it instantly.

**And the browser's own tooltip landed on top.** In the first screenshot, `View Bin (third, center) details` over the open panel is a native `title` on the thumbnail button. Native tooltips render above every layer on the page, on their own schedule, and cannot be styled or dismissed.

## What replaces it

`src/lib/popover/`, one implementation:

| | |
|---|---|
| `floating.ts` | placement, flip, shift, clipping detection, one-open-at-a-time |
| `Popover.svelte` | hover to peek, click to pin, any content and any trigger |
| `Anchored.svelte` | the same layer with no chrome, for a menu or a colour picker |
| `tip` / `tipScope` | a short label in place of a native `title` |

- **Everything renders into a layer appended to `<body>`**, or to the open `<dialog>` when the trigger is inside one, because a modal dialog is painted in the browser's top layer above every `z-index` on the page. Nothing upstream can clip it and nothing can paint over it.
- **Panels flip and slide to stay on screen, and cap their own height.** A long notice scrolls inside the popover rather than running off the bottom of whatever was clipping it. The panel's outer 8px are transparent and sit flush against the trigger, and closing is delayed by 220ms, so there is no gap to fall through on the way to it and no need to move in a straight line.
- **Keyboard and touch**: focus opens it, Escape closes it and returns focus, a tap pins it, clicking outside dismisses it. `aria-expanded` and `aria-describedby` are wired through to whatever trigger a caller supplies.
- **It hides itself** when its row scrolls out of a scroll container, instead of hanging in the viewport pointing at nothing.
- **One at a time.** Opening a popover closes the previous one and any open tip, unless the new trigger is inside the open panel.

**Every native `title` is gone**: 79 in the calculator, 4 in the docs, plus the copy-link buttons the docs inject into headings and the affiliate-link star inside rendered markdown. They are `use:tip` now, drawn in the same layer, and a tip will not open while a popover is open. `use:tipScope` handles `data-tip` inside `{@html}` content, where there is no element to hang an action on. A tip on something wide (a table row) anchors to the cursor and waits longer, since the pointer crosses those while reading rather than to ask a question.

Two things went past the popovers themselves because they are the same bug in a different costume: the **colour picker** and **`DropdownMenu`** were both absolutely positioned panels that could be clipped, and are now `Anchored`. The **fastener chip** in the docs lost its `title`, which only repeated the label printed next to it.

## Kept in sync by hand

`docs/` and `parts-calculator/` are separate npm packages with nothing linking them at build time (the one existing cross-import, `docs/src/lib/server/content.ts`, reaches for a JSON file, not a component). So the family is a byte-identical copy in each, and `parts-calculator/scripts/check_popover_sync.py` fails if they drift.

**It is not wired into `.github/workflows/check-parts.yml`**: this account's token has no `workflow` scope, so a PR from it is rejected outright if it touches that file. The one line that adds it, next to the other checks:

`- run: python parts-calculator/scripts/check_popover_sync.py`

## How it was checked

Both production builds served locally and driven with a headless browser, since a still screenshot proves nothing about a hover behaviour. **17 checks on the calculator, 5 on the docs, all passing**: opens on hover, survives the pointer walking from the badge onto the panel, stays while the pointer rests on it, is not inside `.pl-scroll` any more, sits inside the viewport, scrolls its own overflow instead of overflowing (`clientHeight` 421 against `scrollHeight` 486), click pins, Escape closes, no `title` attribute left on either page, a tip never opens over an open popover, and the colour picker opens in the layer and closes on Escape.

## Follow-up: a wheel inside the panel did not scroll it

[Reported by barthel](https://discord.com/channels/1430279849171222682/1546693223143972904/1546729604088336414) against the preview. Not Firefox-specific, every browser, and my own bug: the panel repositions itself on scroll events, and a wheel inside the panel is a scroll event that reaches the capture listener before anything else. The reposition clears the height cap to measure the card's natural size, the card briefly does not overflow, and a browser clamps a non-overflowing element's `scrollTop` to 0. It scrolled 65px and snapped back to the top on every tick.

- `autoUpdate` ignores scroll events whose target is the card or inside it. The anchor has not moved, so there was nothing to reposition.
- `applyPlacement` saves `scrollTop` before it lifts the cap and puts it back after, for the repositions that are real (a row above expanding, a resize).

The other half of "cannot be scrolled" is that it did not look scrollable: `scrollbar-width` / `scrollbar-color` replace the `::-webkit-scrollbar` rules, which do nothing in Firefox, and there is now a shadow at whichever edge has more content past it, since macOS hides its overlay scrollbars until you are already scrolling.

Both suites now run in **Chromium and Firefox**: 19 checks on the calculator (two new ones covering exactly this), 5 on the docs, all passing in both.

## Merged `main`, 2026-09-17

[Asked for by barthel](https://discord.com/channels/1430279849171222682/1546704975994818675/1550061305023561779) once the branch went conflicted. 45 commits had landed since it was cut.

Two files conflicted, both because `main` added something in the style this branch replaces:

- **`ColorPicker`** — `main` named the LEGO counterpart of each swatch in its `title`; that string is now the `tip`.
- **The parts table** — `main` swapped the inline Optional badge for the new `OptionalBadge` component in two places. Kept, with this branch's `tip` conversions around it.

Three more merged clean and would have broken at runtime, so they are fixed in the merge commit:

- **`OptionalBadge` is new on `main`** and imports `$lib/components/Popover.svelte`, which this branch deletes. It is on `$lib/popover` now, with a CSS length for `width` and the trigger spreading `props`, the same shape as its sibling `AlternativeBadge`.
- **`PartRef` is new on `main`** and carried a native `title`. `use:tip`, matching `HardwareRef`.
- **The STL download tip on the parts table was my own bug**, from this branch's first commit: Svelte's `{x}` interpolation left inside a JS template literal, so it printed `Download {p.name}.stl{engraveIds && …}` to the screen and lost the `stamped` guard with it. Same trap as `View {p.name} in 3D` further up the diff.

`npm run check` passes in both apps and `check_popover_sync.py` is green.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

basically-balloon:

PR #601, opened 2026-09-08
Take the output guide off C1 and C4, fix the camera clasp screw count, and clear eight stale changes entries

basically-balloon:
**Requested by Spencer (@spencerhubert) [from Discord](https://discord.com/channels/1430279849171222682/1545477314471657474/1546694513433645137): "this hsould have jusat been a thing to remove the part from the assembly not make more issues"**, and, on which assemblies: [the output guide should not be used in the assembly for the classificaiton channel or the bulk bucket](https://discord.com/channels/1430279849171222682/1545477314471657474/1546694860277293116), ["that was just a mistake"](https://discord.com/channels/1430279849171222682/1545477314471657474/1546694887418892308).

## Everything asked for in the thread, in this one PR

1. Output guide off `bulk-bucket` (C1) and `classification-chamber` (C4), version-stamped.
2. Count back to 2, on C2 and C3: the part's `quantities` and description follow.
3. `output-guide-clocking` deleted, not completed. Nothing was fixed; it was never a defect.
4. `output-guide-finned-rotor-clash` closed, resolved by the C4 removal.
5. `fix-output-gear-rotor-fastener` closed: the catalog has said M3 x 12 since #515.
6. `improve-camera-extension` and `retain-camera-led-strips` deleted, parts unused.
7. `lock-overhead-camera-mount` and `secure-classification-chamber-rotation` deleted, same reason: the Overhead camera mount and the Classification dome both went when the Camera lamp landed.
8. `camera-clasp-screw-count` solved rather than recorded: `adapted-camera-module` v1 to v2, 4 screws to 2.
9. Docs pages corrected to two guides, with the handover framing removed.
10. Issues #541, #543 and #580 closed by this PR; #540 closed as superseded.

The standing instruction behind all of it ("solve the problem, do not file it") is recorded in my own notes, not here, since it is not a repo change.

The two P0 entries #542 landed yesterday described a part that should not have been there. This takes it out instead, and clears four other entries he pointed at in the same thread.

## The output guide comes off C1 and C4

`bulk-bucket` v3 to v4 and `classification-chamber` v3 to v4, each losing its `output-guide` line and the friction joint that went with it. Both were given a guide by `e22fce30` (#534, 2026-09-02) when the four channels were consolidated onto one shape, and neither had one before that, so this restores the earlier state.

[...]

PR #603, opened 2026-09-08
docs: the fan arm and the Orange Pi's fan, and the six inserts the mount takes

basically-balloon:
**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1546944765709389834):** "Please add the assembly instructions for the fan arm and the Orange Pi fan on https://docs.basically.website/hardware/electronics/installation/orange-pi-mount/. Check whether we need heat inserts on the side of the Orange Pi mount and what kind of screws should be used to attach the fan arm to the Orange Pi mount."

**Also from barthel [in the same thread](https://discord.com/channels/1430279849171222682/1546944765709389834/1546955228216172635):** "Could you perhaps find some photos of the fan arm with the fan attached, an Orange Pi mounted on the mount, and a complete mount with the Orange Pi attached to the hex frame?" There are none, anywhere in the server: no message has ever mentioned the fan arm, and the only photo of the printed mount is [zed0's](https://discord.com/channels/1430279849171222682/1437436041685631129/1542842638015922206), the plate with its four top-face inserts beside an Orange Pi 5 Pro that does not fit it. So the renders stay and the `img-placeholder` in the frame step stays.

That search did turn up one correction, which is the last commit here. The page asserted the fan's lead "has to reach a switched 24 V port on the control board". It does not: the Orange Pi's own manual runs a **5 V** fan off two pins of the 26-pin header, Basically named that route for the Pi's fan on 2026-08-24: ["It also supports this 40mm fan, which can go onto the 5v pins on the pi"](https://discord.com/channels/1430279849171222682/1437436041685631129/1541498838680997979), and ["it's the same form factor of fan as I used on the other enclosure except the one on the basically board is 24v"](https://discord.com/channels/1430279849171222682/1437436041685631129/1541499174556668024), and the same 40 mm frame is sold in both voltages. The page now says which fan this is has not been settled, instead of naming a joint that may not exist.

Both of those questions are answered by the geometry, so everything below is measured off the two published STLs (`orange-pi-extrusion-mount`, hash `4832b406`, and `fan-bracket-40mm`, hash `d4757b35`), not from a build. The page's front-matter warning still stands.

### Heat inserts: yes, two more, and they are in the front edge face

The mount has **six** M3 insert bosses, not four. Four open upward on the top face at the corners of the Pi's 95 x 56 mm pattern, which is what the page already documented. Two more lie on their sides in the plate's front edge, on the centreline, 32 mm apart, and they are the same size as the other four: a 4.00 mm bore, 6.00 mm deep, flat bottomed, which is a ruthex `RX-M3x5.7` boss. They are what the fan arm bolts back into, which is why nothing on the page used them.

These are the same two features [#421](https://github.com/basicallysource/sorter-v2/pull/421) picked up as four holes in a "32 x 6 mm grid" on the plate's middle rail and then rejected as the wrong ones once the real standoff pattern turned up in the corners. They were the two ends of two holes, and the rejection was right about the standoffs but never said what they actually were.

### Screws: M3 x 16, both joints

**Arm to mount, 2 off.** The arm's foot is 12.0 mm thick with a 3.50 mm clearance hole through it, chamfered at the head end, no counterbore. 12.0 mm of travel, then 5.7 mm of insert. An `scr-m3-16-shcs` leaves about 4 mm in the insert. An M3 x 12 is exactly flush with the insert's mouth and grips nothing; an M3 x 20 bottoms out on the 6.00 mm pocket floor at 18 mm and jacks the arm off the plate.

**Fan to arm, 4 off.** The arm's top plate is 12.0 mm thick and flat, with four 2.50 mm pilots on the standard 32 x 32 mm fan pattern around a 38 mm opening, chamfered on the top face. Self-tapping, no inserts, same as the fan on the [control board housing cover](https://docs.basically.website/hardware/electronics/installation/control-board-housing/). The same M3 x 16 gives 6 mm of thread; the housing's M3 x 12 is sized for a thin cover and would only reach 2 mm into this one.

The fan seats on the flat top face at z=70 and blows down through the opening, which flares out underneath from 38 mm to 46 mm. That is the only flat 40 mm seat on the part.

### What changed

**Docs** (`orange-pi-mount.md`)

- Step 1 now presses 6 inserts, with a render of all six ringed, and a note that the two in the edge go in with the plate stood on end.
- New step 3, the fan onto the arm while it is loose, with the lead taken down through the 18 x 10 mm slot at the near end of the arm.
- New step 4, the arm onto the mount, with the screw-length arithmetic in one line.
- Step 6 says where the fan's power has to come from, and says plainly that the port has not been recorded from a build. The Orange Pi has no 24 V rail and no fan header (Orange Pi 5 user manual v2.1.1 recommends a 5 V fan off the 26-pin header), the catalog specifies a 24 V WINSINN 4010, so the lead has to reach one of the control board's LED ports with its bypass jumper bridged. **This one wants a person**, and it is the only thing on the page that is not settled by geometry.

**Catalog** (`orange-pi-mount`, v1 to v2)

- `hsi-m3` 4 to 6, and `scr-m3-16-shcs` x 6 added. The assembly had the arm and the fan as members with nothing holding either on.
- Both joints recorded in `connections` with their measured `through_mm` / `thread_mm`, marked `draft` until somebody builds one.
- `fan-bracket-40mm`'s description said "two per machine", which its own `internal_notes` and its `quantities` have contradicted since the control board got its printed housing on 2026-08-23.

Renders are meshfig, from the STLs, on the [asset service](https://assets.basically.website/sorter-parts/opi-mount-six-inserts-full-09f831150f82.png).

`orange-pi-mount` v2 is deliberately left with `"commit": null`. The board squash-merges, so any hash written on a branch stops existing the moment it lands, and `version-snapshots.mjs` fails the build on a commit it cannot read. The v1 line snapshot the stamp wrote is kept; the hash is for whoever runs `stamp_versions.py` after the merge.

Unrelated, noticed while stamping: `classification-chamber` v4, `bulk-bucket` v4 and `adapted-camera-module` v2 are all sitting on `main` with `"commit": null` from [#601](https://github.com/basicallysource/sorter-v2/pull/601). `stamp_versions.py` picks them up on any run, so they are left alone here rather than dragged into this diff.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #604, opened 2026-09-08
C1 takes no camera lamp: three per machine, not four

basically-balloon:
**Asked by Jon (@effreek)** in [#balloon-electronics](https://discord.com/channels/1430279849171222682/1546665692022050847/1546956188976156675), on a thread where I had just told [iubiubiub](https://discord.com/channels/1430279849171222682/1535802933768364052/1546665692022050847) the machine takes four camera lamps because that is what the catalog says:

> this is incorrect. it should only be 3 lamps. the bulk bucket doesn't get a lamp.

## What changed

**The catalog.** `bulk-bucket` (C-channel 1) loses its `camera-lamp` line, v4 to v5. `e22fce30` (#534) gave C1 one when all four channels were consolidated onto one shape; nothing since has used it. The lamp itself is untouched, as are C2, C3 and the classification channel.

The software agrees with Jon: there is no C1 camera role anywhere in `software/sorter/backend`, the crop zones are the second channel, the third channel and the classification channel, and the bulk station ignores vision. BrickCycleAlice raised exactly this on 2026-09-05 and it had gone unanswered until now.

Knock-on counts for anybody printing or buying: **3** lamps, so 3 each of the mount, arm, brackets A and B, ring, both clasp halves, reflector and cover, **18** LED hooks rather than 24, and **2** OV9732 rather than 3.

**The docs.** `camera-lamp.md`, `c-channel.md` and the feeder index all said every channel carries one. The open-question warning on `camera-lamp.md` is now a plain note for people who already printed four.

**Also on `camera-lamp.md`, the LED strip.** The page said the length was not recorded, which is what started the thread. It is measurable off the published reflector STL: the strip sits against the inner face of the reflector's outer skirt, a cylinder 147.0 mm across (r 73.5, outer face r 75.0) and about 18 mm tall (z 158.41 to 176.52), so **one turn is 462 mm**. [Spencer's photograph of a lit lamp](https://assets.basically.website/sorter-docs/camera-lamp-lit-from-below-w1600-bb9d54f4f43e.jpg) shows two turns side by side in that skirt, each with its own LED row and its own printing, so **about 0.92 m per lamp and 2.8 m for a machine**. The catalog's `led-strip-24v` is a 24 V daylight-white 6000 K COB strip on a 5 m roll, one roll per machine, which covers all three. Which strip is actually on the lamps in those photographs, and how it is wired, is still not recorded, and the page still says so.

## Checks

`check_versioning.py --base origin/main`, `check_generated_pins.py`, `check_connections.py` and `check_asset_urls.py` all pass. `generate.py --metadata-only` regenerated `catalog.generated.json` in the same commit. `bulk-bucket` v5 is left with `"commit": null` because the board squashes.

### DIFF
--- a/docs/src/content/hardware/assembly/feeder/c-channel.md
+++ b/docs/src/content/hardware/assembly/feeder/c-channel.md
@@ -52,7 +52,7 @@ The parts list above is **all four channels' worth** — the count is fixed for 
 
 - **Build 4 per machine**, three in the feeder and one for the classification channel. It is the same build four times over, and the rotor is the only thing that changes.
 - **One rotor per unit, and only one of the two.** The three feeder channels take the Rotor (faceted); the classification channel takes the Rotor (finned), the one with the fins in the photographs below. The list above already splits them the way a real machine needs them: three faceted, one finned. Colour does not change with the rotor: both print ash grey on all four channels, as do the stator and the output guide.
-- Every channel also carries a [camera lamp]({{ '/hardware/assembly/feeder/camera-lamp/' | relative_url }}), whose parts and screws belong to that page rather than to the list above.
+- C2, C3 and the classification channel also carry a [camera lamp]({{ '/hardware/assembly/feeder/camera-lamp/' | relative_url }}), whose parts and screws belong to that page rather than to the list above. C1, the bulk bucket, does not.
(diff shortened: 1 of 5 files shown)


basically-balloon:
Jon ⚡ marked this ready for review in Discord.

basically-balloon:

basically-balloon:

PR #616, opened 2026-09-09
catalog: specify the 16-pin ribbon cables by length range, not by listing

basically-balloon:
The 16-pin ribbon cable between the distribution board and the chute was pinned to one Amazon listing for a 1 m uxcell cable, with the length and the brand written into its name and its attributes. It is a stock part, so what matters is that the cable is 16-pin FC-to-FC at 2.54 mm pitch and long enough, not which shop it came from.

**Requested by Spencer H (@spencerhubert) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1547332898754011216):** *"i think it's trying to link to some specific cable listing right? i instead want to make it more generic and give that type of thing an ideal length and a length min and max set of params. the ideal is 1.2m. the min is 1.2m. the max is 1.5. do the same for the other 16 pin cable but just set the ideal length to whatever the current length is for that one, leave the min and max null."*

## The schema

`cots` gains two optional fields beside the existing `length_mm`:

- `length_mm` is the length the part **is**, or, for an item specified rather than pinned, the **ideal** length to buy.
- `length_min_mm` / `length_max_mm` bound what else still works. Null or absent on either side means the ideal is the only length that does, which is every screw and standoff in the catalog, so nothing else changes.

## The two cables

- **`cable-idc-2x8-long`** (distribution board to chute): ideal 1200 mm, range 1200 to 1500 mm. Renamed to *IDC ribbon cable, 16-pin (2x8), 1.2 to 1.5 m*, brand attribute dropped, and the old 1 m listing swapped for the [1.5 m MECCANIXITY 2-pack](https://www.amazon.com/MECCANIXITY-Connector-Ribbon-Industrial-Electronics/dp/B0GJ8PP3LH) Spencer linked (2 pcs, $11.39, checked 2026-09-09), with a note on the vendor saying it is the long end of the range and any cable inside the range works. The old listing is a 1 m cable, which is now below the minimum.
- **`cable-idc-2x8-short`** (layer to layer below): ideal 300 mm, min and max null, as asked. Its listing and photo are untouched.

Both product photos are kept.

## Where it shows up

- **Parts calculator, hardware list**: the corner badge now reads `1.2m` rather than `1200mm` (lengths past a metre read in metres, screws still read in mm), and a card with a range gets a spec chip: *Length 1.2 m, 1.2 m to 1.5 m works*. Cards with only an ideal length are unchanged.
- **Hardware detail**: a `Length` row with the ideal and the range, only where there is no authored `Length` attribute, so screws keep the attribute they already state.
- **Docs parts-needed cards** (`Requirements.svelte`): same badge formatting, and the popover under it explains the range instead of the generic "shared photo" line.
- **`hardware.csv`**: `length_min_mm` and `length_max_mm` columns, so a substitute can be judged from the file on its own.
- **Search**: both apps index `1.2m` and `1.5m` alongside `1200mm`.

## Rebased on main (2026-09-17)

Rebased onto `main` at #602 (the shared popover layer). The one conflict was the length badge in `Requirements.svelte`: #602 turned the corner badges into `use:tip` / `Popover`, this branch had rewritten the same badge to say the ideal length and explain the range. Kept #602's structure and put the range text back through it (`use:tip={{ text: lenNote(part), wide: true }}` around `lenText(part.length_mm)`), so there is one popover implementation and the badge still reads `1.2m` with *Ideal length. Anything 1.2 m to 1.5 m works here* under it. The no-range wording now matches main's rewritten sentence. Everything else merged clean, and `catalog/generate.py --metadata-only` reproduces the generated catalog byte for byte against the rebased `parts.json`.

Re-checked after the rebase: `npm run check` in `docs/` (810 files, 0 errors, 0 warnings, the a11y warnings this branch used to carry are gone with #602) and in `parts-calculator/` (4394 files, 0 errors), both proved real with a deliberate type error, plus `scripts/validate_images.py` (292 assets) and `catalog/generate.py --metadata-only`, which reports the generated catalog agrees with `parts.json`.


basically-balloon:
Spencer marked this ready for review in Discord.

basically-balloon:

basically-balloon:

PR #623, opened 2026-09-10
docs: the last assembly images move out of the catalog and onto their pages

basically-balloon:
**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1547571868389609552/1547576232885883011):** *"Check the other assemblies in the same way. If they contain images that haven't been added to the documentation yet, create a PR to add them there. If a documentation page is missing for an assembly, please create it with the AI Warning and add the images right away. Then remove the images from the parts calculator assembly tree in the same PR as the documentation update."*

Follows [his earlier ask](https://discord.com/channels/1430279849171222682/1536088194557419660/1547571868389609552) that produced #622, which did the camera lamp: *"I want to prevent a situation where we suddenly end up with multiple places where built elements appear that could potentially be out of sync."*

## The sweep

Five assemblies carried `images` after #622 took the camera lamp's out. Four of them at the top level, and **none of those eight images was on the docs page that builds the thing** (checked URL by URL against each page's assets). Every one has a docs page already, so no new page was needed.

| Assembly | Images | Where they went |
|---|---|---|
| Light post | 4 photos | `/hardware/assembly/feeder/light-post/`, overview + steps 2 and 3 |
| Cage bracket (cable mount) + ribbon clamp | 1 render | `/hardware/assembly/distribution/top-interface/`, step 11 |
| basically Embedded Control Board Housing | 2 renders | `/hardware/electronics/installation/control-board-housing/`, step 4 |
| Orange Pi mount | 1 render | `/hardware/electronics/installation/orange-pi-mount/`, overview |

### Light post

The page's own warning said *"every photograph are still missing"* and it had five `Image coming` placeholders. [Spencer's four bench photos](https://github.com/basicallysource/sorter-v2/commit/271ea214173484a33b0166e4cc4c98a4eaa29b69) fill three of them, and the warning is corrected.

[...]

PR #620, opened 2026-09-10
docs(top-interface): BrickCycleAlice's build photos on steps 3, 4, 5 and 8, and step 1's real credits

basically-balloon:
BrickCycleAlice went through the Top interface page against her own build and said where each of her numbered photos belongs. This places them, and fixes the credits on step 1.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1547526128942583838/1547527748074086510):** "Top interface Step 1, my photos have Grey parts, the black and white part photos have the wrong source listed, go back a few revisions. Step 2 this photo should be the last one,"

Then, in the same thread: ["step 3 add photos 6 and 8"](https://discord.com/channels/1430279849171222682/1547526128942583838/1547527943083786250), ["Step 4, photos 9 and 10, Then photos 14 and 15"](https://discord.com/channels/1430279849171222682/1547526128942583838/1547529114775195699), ["Step 5, photos 18 and 19, then 22 and 23"](https://discord.com/channels/1430279849171222682/1547526128942583838/1547529722140041286), ["step 8, photos 24 and 25"](https://discord.com/channels/1430279849171222682/1547526128942583838/1547530192996671508), and finally ["Ok that's all I've got for the top interface. I'm thinking we do a pr per page, to allow for ease of review and tracking."](https://discord.com/channels/1430279849171222682/1547526128942583838/1547531053433102336) The photo numbers are the ones agreed with her on 6 Sep, from her 30 Aug build thread.

## Step 1: four stills were credited to the wrong person

[...]

PR #626, opened 2026-09-10
docs(top-interface): remove the duplicate og_image key that broke the frontmatter

basically-balloon:
`top-interface.md` picked up a second `og_image:` line in #620. js-yaml throws on a duplicated mapping key, so `parseFrontmatter` in `docs/src/lib/server/content.ts` swallowed the error and fell back to its flat `key: value` reader, which turns every value into a string. On the live page that meant:

- **Parts needed is gone entirely.** `parts_needed:` parsed as the empty string, so all 54 entries disappeared and the section never rendered.
- **Tools needed renders one bullet per letter.** `tools_needed: [Hex key, Soldering iron or heat-set insert press]` became the literal string, and the template iterated its characters.
- **The byline** read `[barthel, brickcyclealice]` as a single contributor name.

This deletes the duplicate line. Nothing else changes, and the remaining `og_image:` is byte-identical to the one removed.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1547532333983465473/1547591718092939347):** "https://docs.basically.website/hardware/assembly/distribution/top-interface/ is broken"

The duplicate came in with #620, whose merged diff re-added the line at its old position while `main` already carried it at line 10 after #623 moved it.

## Checks

`npm run check` 0 errors (same 4 pre-existing a11y warnings in `Requirements.svelte`), `docs/scripts/validate_images.py` clean at 245 assets. Against `npm run dev` locally the page now renders **Parts needed** with 54 rows and **Tools needed** as two items.

Worth someone's judgement separately: that `catch` in `parseFrontmatter` turns a YAML error into a silently half-rendered page, and no check caught it. Failing the build, or at least warning, would have.

### DIFF
--- a/docs/src/content/hardware/assembly/distribution/top-interface.md
+++ b/docs/src/content/hardware/assembly/distribution/top-interface.md
@@ -10,7 +10,6 @@ permalink: /hardware/assembly/distribution/top-interface/
 og_image: https://assets.basically.website/sorter-docs/assembly-top-interface-framing-4-full-b9ae16940954.jpg
 author: zed0
 contributors: [barthel, brickcyclealice]
-og_image: https://assets.basically.website/sorter-docs/assembly-top-interface-framing-4-full-b9ae16940954.jpg
 parts_needed:
   - part: interface-upper-fixed-section
     qty: 1


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #628, opened 2026-09-10
docs: the top interface's lead figure repeated its own result photo

basically-balloon:
**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1547315250515157095/1547602127243710644): "Remove the first instance of the image `assembly-top-interface-framing-4-full-b9ae16940954.jpg` from the top interface".**

That picture was on `top-interface.md` **three times**: as the lead figure under the intro, in step 13 under a second name (`assembly-top-interface-framing-3-full-b9ae16940954.jpg`, byte-identical, captioned "A hex frame lowered onto the interface assembly"), and in the finished result at the foot of the page. This removes the first of the three, the lead figure.

`og_image` already points at the result copy, so the page's link preview is unchanged.

**Still duplicated, not touched here:** step 13 and the finished result are the same bytes under two names. Flagged to barthel [in the thread](https://discord.com/channels/1430279849171222682/1547315250515157095/1547602127243710644); one of the two wants a different photo rather than a deletion, which is a decision rather than a fix.

This follows sorter-v2#612, which merged at 09:14 UTC while the thread was still running.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #638, opened 2026-09-11
parts-calculator: the Orange Pi WiFi module is optional, and a COTS part can say so

basically-balloon:
**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1477417230235861104/1547986139305218222): "Mark wifi-module-opi5 as optional in bom parts catalog."**

`wifi-module-opi5` now carries `"optional": true` in `catalog/parts.json`, and off-the-shelf parts can carry that flag at all, which they could not before.

**What changed**

- `catalog/parts.json`: `"optional": true` inside `cots` on `wifi-module-opi5` (see below for why in there). Its description already said "Optional."; the flag is what a surface can read.
- `src/lib/filament.ts`: `optional?: boolean` on a `Hardware`'s `cots`, plus `hardwareOptional(h)` as the one reader. Same meaning as `Part.optional`: it marks the item, it does not take it out of the quantities or the cost.
- `/hardware` row and `HardwareDetail` (so the modal, the assembly tab and `/part/wifi-module-opi5` all get it): the amber **Optional** badge, the same `Badge variant="warning"` printed parts use.
- `hardware-csv.ts`: an `optional` column, next to `qty_source`, matching the printed-parts CSV which has had one all along.
- `OptionalBadge.svelte`: the marking itself, one component, so an optional printed part and an optional bought one carry the same tag and the same explanation. It sits in the cluster next to the name, beside the `ChangeStatus` priority chip and the green "A", and like both of those the tag is the trigger for the sentence explaining it. Asked for [in the thread](https://discord.com/channels/1430279849171222682/1547986139305218222/1547993688292982896): "an aditional tag/marking for optional items like this. would be great."
- It renders wherever an item is named: the `/hardware` row, `HardwareDetail` (the modal, the assembly tab and `/part/<id>`), both printed-parts tables on `/`, and the hardware rows in the assembly tree and `AssemblyDetail`. The two printed-parts tables already had a plain amber "Optional" chip; they now use the shared one, so the wording is in a single place.

The flag marks the item, it does not remove it from anything: quantities, weights and costs all still include an optional part, and the popover says so.

`npm run check`: 0 errors, 0 warnings. `scripts/check_generated_pins.py` still agrees.

**Where the flag lives, and why it is not beside `alternative`**

`catalog/generate.py`'s `build_hardware()` copies a fixed list of keys onto each COTS record, so a top-level `optional` is silently dropped on the way to `catalog.generated.json` and never reaches the app. That file is maintainer-owned and not one I open PRs against, so the flag rides inside `cots` instead, which the same function passes through whole:

```json
"cots": { "type": "wifi", "optional": true }
```

That is a real fix rather than a patch: it survives regeneration, because the generator reproduces it. The type carries the reason and what to do about it, so the day the whitelist gains `"optional": p.get("optional", False)` the flag moves up a level and that paragraph gets deleted. `hardwareOptional(h)` is the single reader, so the move is one line in `filament.ts`.

`src/lib/data/catalog.generated.json` is regenerated for it. `generate.py` cannot run in my container (it imports `engrave`, which needs `scipy`), so I extracted `build_hardware` from the generator source with `ast`, ran that one function against the edited manifest, and confirmed its output is identical to the committed `hardware` array. The one difference it produced is the two-line `cots` change, which is what is committed. `scripts/check_generated_pins.py` agrees.

The docs site's parts-needed cards (`docs/src/lib/server/content.ts` into `Requirements.svelte`) also read hardware off the same generated record and have no optional marker; that is a separate change and worth doing once the field actually arrives.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #640, opened 2026-09-12
parts+docs: the control board housing fan screw length (ON HOLD, see banner)

basically-balloon:
> [!WARNING]
> **On hold, do not mark ready.** The 16 mm figure assumes the fan bears on the four nub end faces and stands 2.5 mm off the lid. Daddy-O's Bricks - Bill, who has the parts in front of him, reports that the nubs instead recess into the recesses in the fan's corners so the fan sits flush. If that is right there is only 7.5 mm of fan above the thread, the longest usable screw is 14 mm, **M3 x 16 would bottom out** and the existing M3 x 12 is correct with 4.5 mm of bite. Waiting on two bench measurements: whether a dry-fitted fan sits flush or stands off, and how deep a screw actually goes into a bare nub hole on a real print.
>
> The cover geometry in this PR body is verified three ways against the pinned STL and is not in question. What is in question is how the fan sits on it.

The control board housing's fan is specified as four self-tapping M3 × 12 mm button heads. It should be M3 × 16 mm.

Asked for by **Daddy-O's Bricks - Bill** (@daddyosbricksbill), who [snapped one of the cover's fan nubs off with a 12 mm screw](https://discord.com/channels/1430279849171222682/1487518366020276397/1548126898620731524) and asked whether the joint takes 12 or 10.

### What the geometry says

Measured off the current `ctrl-board-housing-cover` STL (`stl_hash` 462b0290…, pulled from the asset service). The four fan bosses sit on a 32.0 × 32.0 mm pattern, standard for a 40 mm fan, and each one is:

- a **2.5 mm tall nub**, ⌀6.0 mm, standing off the inner face of the lid
- with a **⌀2.5 mm pilot hole** running **6.5 mm deep in total**: 2.5 mm through the nub, then 4 mm up into the 5 mm lid plate
- blind, closing flat at 1 mm below the outer face

The nub wall is therefore 1.75 mm.

### Why 12 mm is wrong

The WINSINN 4010 is 40 × 40 × 10 mm ([manufacturer page](https://winsinn.com/40mm-fan-4010/)), so the screw does nothing for its first 10 mm.

- **M3 × 10** leaves 0 mm of thread in plastic.
- **M3 × 12** leaves 2 mm, and all 2 mm of it lands inside the 1.75 mm-walled nub. That is the joint that failed.
- **M3 × 16** leaves 6 mm: through the nub and 3.5 mm into the solid lid plate, stopping 0.5 mm short of the blind end. That 0.5 mm margin reads as the design intent.

It does not actually bottom out at 12 mm, as reported; there is 6.5 mm of hole. The likely cause of the hard stop is the tapered lead of the self-tapper binding in a pilot that printed undersize.

### Corroboration elsewhere on the machine

The machine's other 40 mm fan, on `fan-bracket-40mm`, has an 11 mm pilot and the catalog already calls for **M3 × 16 self-tapping** into it, for the same 6 mm of engagement. Measured off that STL too.

### What changed

- `ctrl-board-mount` stamped **v2 → v3** (new uid `60ni`, `breaking: false`, `commit: null` for the post-merge stamp): the fan line goes `scr-m3-12-bhcs` → `scr-m3-16-shcs`, qty 4 unchanged, and the `joining` note follows it. Nothing else in the assembly moves.
- Descriptions updated on `ctrl-board-housing-cover`, `fan-40mm-24v` and `scr-m3-16-shcs`.
- Control board housing page: `parts_needed` and step 3's fastener chip.
- `catalog.generated.json` regenerated with `generate.py --metadata-only`; its own pin, versioning and connection checks pass.

Nothing about the cover's geometry is wrong, so there is no `changes` entry and no `cad` issue here. The catalog could say the fix, so it says it.

### DIFF
--- a/docs/src/content/hardware/electronics/installation/control-board-housing.md
+++ b/docs/src/content/hardware/electronics/installation/control-board-housing.md
@@ -26,7 +26,7 @@ parts_needed:
     qty: 8
   - part: scr-m3-6-bhcs
     qty: 4
-  - part: scr-m3-12-bhcs
+  - part: scr-m3-16-shcs
     qty: 4
   - part: scr-m3-12-cs
     qty: 4
@@ -78,7 +78,7 @@ Sit the board on the four inner bosses and fix it with 4 {% include fastener.htm
 
 {% include step.html n="3" title="Fit the fan inside the cover" %}
 
-Fan on the inside of the cover, over the vent, label facing into the enclosure so it blows inwards. 4 {% include fastener.html size="M3" variant="button" length="12" %} screws, self-tapping into the plastic. Route the lead to the corner cutout.
+Fan on the inside of the cover, over the vent, label facing into the enclosure so it blows inwards. 4 {% include fastener.html size="M3" variant="button" length="16" %} screws, self-tapping into the cover's four bosses. Route the lead to the corner cutout.
 
 <div class="img-row">
   <figure>
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -2817,11 +2817,11 @@
       "id": "ctrl-board-housing-cover",
       "uid": "ksh2",
       "name": "Control Board Housing — Cover",
-      "description": "Cover of the basically Embedded Control Board housing. Screws to the base with four M3 × 12 mm countersunk screws, carries the 40 mm 24 V fan on four self-tapping M3 × 12 mm button head screws, and the plunger passes through it.",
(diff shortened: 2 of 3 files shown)


basically-balloon:
Closing at the requester's ask. Daddy-O's Bricks - Bill confirmed at the bench that the nubs bury in the fan's corner recesses and the fan frame sits flat on the lid, which is the seating case this PR did not account for. Under that seating there is only 7.5 mm of fan above the thread, so M3 x 16 would bottom out and the catalog's existing M3 x 12 is correct. No change is needed and nothing here should land.

Separately, and not a reason to reopen this: on his print an M3 x 6 bottomed out and snapped the nub on three separate bosses, against a 6.5 mm bore in the file. That is a printability question about the boss, not a fastener spec question, and it is not mine to file over his objection.

PR #619, opened 2026-09-10
The camera lamp's LED strip: a catalog line, and a page for wiring it

basically-balloon:
The LED strip in a camera lamp was in no parts list and its wiring was written down nowhere, which is where this build stopped.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1547465864184463402/1547474283423473726): "Ok, can you double check this against any similar items already in hardware. Then make a pr for this, as at present it is stopping a build. I'll get Jon or someone else who knows connections to double check."** The thread started with ["help me out with the led strip for the lamp, it's not in the parts needed, how long should it be how do I join the ends? Do I use the sticky backed tape to help secure it?"](https://discord.com/channels/1430279849171222682/1536088194557419660/1547465864184463402)

## The catalog: the strip is a member now

`led-strip-24v` was a transitional `sheet_qty` hand count sitting outside the tree, so it appeared in no assembly and on no page's parts list. It is now a line of `reflector-with-led-hooks` (v1 to v2), which is the node whose six hooks exist to retain it.

The similar item already in the catalog is **`rod-steel-3-8`**, the steel rod the retired overhead camera arms were cut from: a `stock` part, bought as a 4 ft length, counted in pieces, with the app dividing pieces into lengths to buy. `parts.json`'s own schema comment says that is the pattern ("A cots part that is cut from a bought length carries `stock {...}`; its lines count PIECES and the app divides to get lengths to buy"), so the strip now carries `stock` too: a 5 m roll, cut into 950 mm rings, 5 per roll. Three lamps therefore ask for one roll, and the count comes from the tree rather than a hand number.

`breaking` is `false`: the strip was always physically in this assembly, it was just never recorded.

## Where 950 mm comes from

The reflector's outer skirt is 147 mm across, so one turn is 462 mm and the two turns that fit side by side are 924 mm. That measurement is off the published `lamp-inner-reflector` STL and has been on the camera lamp page since #548. The catalog's roll is [cuttable every 1.5 in](https://www.amazon.com/dp/B0DHV8ZF4R), so the first mark past 924 mm is 25 segments, about 950 mm. [BrickCycleAlice cut 950 mm and it fits](https://discord.com/channels/1430279849171222682/1547465864184463402/1547468694681034762), which is the first time that figure has been checked against hardware rather than geometry; her photo of it hooked in is the new figure on step 5.

## The new page

[...]

PR #642, opened 2026-09-12
parts-calculator: a product photo for dupont-lead-2p-1m

basically-balloon:
`dupont-lead-2p-1m` has rendered a grey "no image" tile since it went into the catalog with #619, because the Amazon listing would not fetch from the box the generator runs on. It fetches now, so the part has a photo.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1477417230235861104/1548228816428863531): "dupont-lead-2p-1m has no image. Please find an image and add it to the parts catalog."**

## The image

`https://assets.basically.website/sorter-parts/dupont-lead-2p-1m-full-1d6495f8bb1b.jpg`, 1500 x 1500, published with `scripts/publish_assets.py --upload` under the part id like every other COTS photo.

It is the listing's own photo for **the variant the catalog sources**, not the brand's family shot: red and black 22 AWG, a male end and a female end, and two loose 2-pin shells. The generic Kidisoii image that the search surfaces first shows blue, yellow, white and brown leads as well, which is a different variant and would have been misleading on a card that says "red and black".

**Where it came from.** Plain `curl` on `amazon.com/dp/B0CCTZJZZ3` still gets the CAPTCHA page, which is what `scripts/fetch_product_images.py` runs into, so this was done by hand: rendered the listing to confirm the variant (`5Pcs MF 2P-100CM`, Kidisoii, currently unavailable), read the gallery's image ids off the page through a reader proxy, and pulled the hi-res file from `m.media-amazon.com` directly. That CDN is not blocked from here, only the listing HTML is. `fetch_product_images.py` is untouched; making it work against Amazon's bot wall is a separate job.

The `updated_at` stamp moves to 2026-09-12. Nothing else on the part changes, and `catalog/generate.py --metadata-only` wrote the generated file (two lines: the stamp and the image).

## One thing the photo raises, not changed here

The catalog description says the lead arrives "with a 2-pin 2.54 mm Dupont housing crimped on". The listing's own text says the pack is 10 pre-crimped wires plus 10 separate 2-pin shells and that you "insert the pins into the shell", and the photo shows two empty shells alongside the assembled leads. If that is right, the builder pushes the terminals into the housing themselves (no tool, but it is a step), and the description and `alternative` both want a word changing. I have not touched either in this PR, since it is a reading of a listing rather than something anybody has had in their hands.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -5508,7 +5508,8 @@
       "note": "Sold male-to-female: cut the male plug off and discard it. Any 22 AWG pair plus a 2-pin Dupont housing does the same job if you own a crimp tool.",
       "alternative": "Your own 22 AWG pair plus a 2-pin 2.54 mm Dupont housing and two crimps",
       "created_at": "2026-09-10",
-      "updated_at": "2026-09-10",
+      "updated_at": "2026-09-12",
+      "image_url": "https://assets.basically.website/sorter-parts/dupont-lead-2p-1m-full-1d6495f8bb1b.jpg",
       "attributes": [
         {
           "label": "Pitch",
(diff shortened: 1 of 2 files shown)


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

Issue #645, opened 2026-09-12
Control board housing cover: the four fan bosses have a sharp 90 degree root, no fillet, and are snapping off


basically-balloon:
The four fan bosses on `ctrl-board-housing-cover` meet the lid with a **sharp 90° internal corner and no fillet anywhere on the boss**. They are breaking off in the field.

Reported by **ReveryX** (`@reveryx`), [2026-09-12](https://discord.com/channels/1430279849171222682/1548126898620731524/1548468985375363163): *"the nubs have a 90 degree corner, which adds stress to the part and tends to break. adding fillets should solve this."*

## Measured

From the pinned STL, `ctrl-board-housing-cover` uid `ksh2`, `stl_hash` 462b0290…, still v1 (2026-08-22), so nothing has changed since the part shipped. Coordinates are the STL's own frame; the lid plate is z 30 to 35.

Each of the four bosses, on a 32.0 × 32.0 mm pattern centred at (98, −95):

- a **⌀6.00 mm cylinder** running z 30.0 → 27.5, so **2.5 mm proud** of the lid's inner face
- a **⌀2.50 mm bore** through it, running z 27.5 → 34.0, blind, closing flat 1 mm below the outer face
- wall thickness **1.75 mm**

**There are no vertices anywhere strictly between z = 27.5 and z = 30 other than the two cylinders' own end rings.** The boss is a plain extruded cylinder: sharp 90° internal corner where it meets the lid, sharp 90° external corner at its free end, sharp corner at the bore mouth. ReveryX's read of it is exactly right.

## The failures

**Daddy-O's Bricks - Bill** (`@daddyosbricksbill`) broke **four bosses on one cover** over 2026-09-12:

- [the first one](https://discord.com/channels/1430279849171222682/1487518366020276397/1548126898620731524), fitting the fan with an M3 × 12
- [two more](https://discord.com/channels/1430279849171222682/1548126898620731524/1548138479735808100) testing with an **M3 × 6**, which against a 6.5 mm bore should have gone in with room to spare: *"both bottomed out and snapped the nubs"*

So they are not only failing at a torqued-down joint, they are failing on a screw that meets resistance early. A 1.75 mm wall on a sharp-rooted 2.5 mm stub has very little in it before the root gives.

## The fix

A fillet at the root, where the ⌀6.00 cylinder meets the lid face. That is the corner carrying the bending load when a screw binds.

[...]

PR #646, opened 2026-09-12
parts: record the cover's sharp-rooted fan bosses as a P0 broken change

basically-balloon:
Adds one `changes` entry so the cover's fan bosses carry a warning everywhere the calculator draws them, and records the interim handling until the geometry is revised.

This is the board half of [#645](https://github.com/basicallysource/sorter-v2/issues/645), filed at **ReveryX**'s (@reveryx) [request](https://discord.com/channels/1430279849171222682/1548126898620731524/1548468985375363163). The issue has the measurements and the fix; this is what puts it in front of a builder.

### The entry

`fillet-fan-boss-roots`, **P0 / broken**, targeting the part `ctrl-board-housing-cover` and the assembly `ctrl-board-mount`.

The bosses are plain ⌀6.00 mm cylinders standing 2.5 mm proud of the lid, bored ⌀2.50, so the wall is 1.75 mm and the root is a sharp 90° internal corner with no fillet anywhere on the boss. Measured off the pinned STL (`stl_hash` 462b0290…, still v1). **Daddy-O's Bricks - Bill** (@daddyosbricksbill) broke four of them on one cover, [two of those on an M3 × 6](https://discord.com/channels/1430279849171222682/1548126898620731524/1548138479735808100) that should have gone in with room to spare against a 6.5 mm bore.

`broken` rather than a lower priority because it stops a build: there is no way to finish the housing once a boss is off, and it is coming off on screws that are not even long enough to bottom.

### Why this is an entry and not a fix

No catalog edit resolves it. The fastener spec on the page is correct and untouched, the assembly is correct, and the defect is in the geometry, which is a `cad` issue plus the entry that goes with it. The interim advice is already on the assembly page as a step 3 warning in #641.

### Checks

`generate.py --metadata-only` run and committed with the source; its pin, versioning and connection checks pass.

### DIFF
--- a/parts-calculator/catalog/parts.json
+++ b/parts-calculator/catalog/parts.json
@@ -491,6 +491,21 @@
           "bulk-bucket-support"
         ]
       }
+    },
+    {
+      "id": "fillet-fan-boss-roots",
+      "name": "The cover's fan bosses have a sharp root and snap off",
+      "priority": "P0",
+      "condition": "broken",
+      "description": "The four bosses the 40 mm fan screws into on the Control Board Housing cover are plain ⌀6.00 mm cylinders standing 2.5 mm proud of the lid, bored ⌀2.50 mm, so the wall is 1.75 mm and the root where the boss meets the lid is a sharp 90° internal corner with no fillet. One builder broke four of them on a single cover, two of those on an M3 × 6 that should have gone in with room to spare against the 6.5 mm bore. Until the root is filleted, run the four screws by hand, stop as soon as the fan pulls down onto the cover, and back a screw out and clear the hole rather than forcing one that stops early. sorter-v2 issue #645.",
+      "targets": {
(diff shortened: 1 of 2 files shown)


basically-balloon:
Closing at Daddy-O's Bricks - Bill's ask. He is the builder the evidence in this entry came from, and he has since [reprinted the cover and fitted the fan with M3 x 10 socket heads with no trouble](https://discord.com/channels/1430279849171222682/1548126898620731524/1548490705268117597), which puts the original breakages down to that first print rather than to the part.

A P0 'broken' badge shows on the part everywhere the calculator draws it, so it is not something to land on two approvals while the premise is disputed. If a maintainer decides the sharp boss root is worth tracking on the board, this entry is in the branch history and is one commit to restore.

The underlying question is still open as #645, which is an issue and not on any merge path.

PR #649, opened 2026-09-13
docs: connecting the components

basically-balloon:
Requested by Daddy-O's Bricks - Bill in `#balloon-electronics`: a page under Electronics for connecting the components, between mounting the three enclosures on the machine and software setup. Assumes the wires are made and the PSU, board and Pi are prepared. Written for the Mira persona: short sentences, one instruction each, every socket named.

### The new page

`/hardware/electronics/installation/connecting/`, step 5 of the installation sequence.

- A "every cable on this page" table: what the cable is, and the connector at each end.
- Three to-scale inline SVG layouts: basically board v1.3, the Orange Pi 5 ports, and the PSU terminal block with its three DC jacks.
- Eight steps: 24 V out of the PSU box, 24 V into the board, the five steppers, the chute limit switch, the three camera lamps, the ribbon down to the layer boards, USB (Pico to hub to Pi, plus the three cameras), then the pre-power check.

The board diagram is generated from `electronics/KiCad/1_Distribution_Board/1_Distribution_Board.kicad_pcb`: connector positions and the board outline are the real footprint coordinates, so the socket positions are right. The Orange Pi drawing follows the manufacturer photo already on `/hardware/orange-pi-5/`.

### Facts read out of the board, not guessed

Everything below was taken from the v1.3 netlist and `software/firmware/sorter_interface_firmware/hwcfg_basically_v1_2.h`, and none of it is checked against a built machine yet.

- **Which stepper socket is which motor.** Each output connector is wired to the driver socket beside it, and each driver socket is wired to fixed STEP/DIR pins on the Pico, so the firmware channel is fixed by the board. `J23/J24` is `Stepper Module` (GP28/GP27) = channel 0 `chute_stepper`; `J27/J28` = channel 1 `c_channel_1_rotor`; `J31/J32` = channel 2 `c_channel_3_rotor`; `J35/J36` = channel 3 `carousel`; `J39/J40` = `Stepper Module4` (GP8/GP7, the second UART bus) = channel 4 `c_channel_2_rotor`. The silkscreen names each one `Stepper_A2` to `Stepper_A6` in that order, which is what the pages now tell a builder to read.
  - `steppers.md` said this was "still to confirm" and put the chute on `J40`. Corrected to `J24`, with the mapping table added.
  - `control-board-prep.md` gave the address jumpers per channel with no way to tell which physical driver that is. It now names the driver as printed.
- **LED port polarity and grouping.** On every LED port pin 1 is +24 V through the 180 Ω resistor (R21 on J8, R22 on J9, R27 on J10, R28 on J11) and pin 2 is switched to ground by an AO3400A. `J8`/`J9` are both on GPIO 1 and `J10`/`J11` both on GPIO 6, so four sockets are two switched channels and one lamp always comes on with the housing fan. The board prints `+V` and `GND` beside each port, so the page tells the reader to go by the print.
- **The chute limit switch goes on `J5`.** `chute_home` is input alias 0, which is `digital_input_pins[0]` = GPIO3, which the board prints `HALL_SW_0`. Pin 1 is GND, pin 2 is SIG (10 kΩ pull-up to 3V3), pin 3 is +3V3, which matches the harness schedule's "position 3 empty". `ChuteCalibrationConfig.endstop_active_high` defaults to `True`, so the page says to use `COM` and `NC` and notes the setting is in software if homing reads backwards.
- **`J1` pin 1 is +24 V** (through F1) and pin 2 is GND (through F2), as the harness page already says.
- **The ribbon chain** is board `J17` to `J3` on the first layer board, then `J4` to `J3` down the stack. `J3` in, `J4` out, confirmed from the Adapter Board netlist in a previous thread with Bill and Jon (`PWM_0` never reaches `J4`). Not previously on any page.
- **The Orange Pi has two USB-C sockets** and only the one printed `PWR IN` is a power input, per the Orange Pi 5 manual. It also has its own 2-pin 5 V `FAN` connector, which is why the page's open item says the 24 V fan in the mount's parts list cannot run from it.

### Follow-up, 2026-09-14

BrickCycleAlice asked in `#balloon-electronics` what actually links the PSU to the control board and the USB hub, since only the buck converter shows up in Parts needed: https://discord.com/channels/1430279849171222682/1535802933768364052/1548985579860009040

The page named the connector at each end of `W1` and `W2` but never said where those leads come from. Added a short "The three 24 V leads" section under the cable table: `W1` is a lead you make (male barrel plug to a VHR-2 housing, no supplier sells that pair of ends), `W2` is a male-to-male barrel lead you buy, `W3` is a 6 in barrel lead onto the buck's input wires. Plus a warning that every PSU jack is 5.5 x 2.1 mm centre-positive and a 2.5 mm plug will not mate, which is read off the [PSU pigtail](https://docs.basically.website/hardware/electronics/psu-pigtail/) page. Credited her in `contributors:`.

### Follow-up 2, 2026-09-14

Same thread, https://discord.com/channels/1430279849171222682/1548985579860009040/1548989251851526265 : how many barrel plugs the whole build needs, what size wire they go onto, and how much of it.

Answering that surfaced a contradiction between two pages. The harness schedule's section 4.2 said **gauge TBD** for `W1`, `W2` and `W3`, while the order spec has carried 18 AWG / 22 AWG / 22 AWG since it was written, off Jon's 2026-08-08 reference drawing (and `electronics/wire_harness/power.yml` agrees). Filled the three cells in, flagging `W2` and `W3` as guesses the way `power.yml` marks them. Open item 5 (gauge per segment needs current per load) is untouched and still open.

Also added gauge and length to each of the three leads on the connecting page, plus the whole-build count (seven male barrel plugs: `W1`, both ends of `W2`, `W3`, and the three LED pigtails) and a link to the order spec for everything else.

### Follow-up 3, 2026-09-14

BrickCycleAlice, same thread: https://discord.com/channels/1430279849171222682/1548985579860009040/1548990053148262412 -- the page now tells you to crimp a VHR-2 and does not list a crimp tool or wire strippers.

- **`connecting.md`**: `tools_needed` gains `"Only if you make your own W1 lead: wire strippers and a crimp tool"`, following the conditional convention already on `led-drop.md` (`"Only if you make your own lead: crimp tool"`). Conditional because the page's premise is that the cables are already made.
- **`psu-pigtail.md`**: had **no `tools_needed` at all** despite being a strip-and-crimp build guide with four crimping photos on it. Given `[Side cutters, Wire strippers, Ratcheting crimp tool, Multimeter]`, matching what its own steps call for (the body already specifies the 22 to 16 AWG die and metering tip vs sleeve). Credited her in `contributors:` there too.

"Wire strippers" is a new name in `tools_needed`; nothing in the docs named the tool before, only the verb. Worth watching under the naming-drift tracker if a different name shows up elsewhere.

**Adding that list blanked the PSU pigtail page, and the cause is worth knowing.** `psu-pigtail.md`'s `lede` contains a bare `: ` (`...the PSU box: a panel-mount barrel jack...`), so `yaml.load` throws and `parseFrontmatter` falls back to its flat `key: value` line reader, which hands every value back as a **raw string**. `tools_needed` arrived as the string `[Side cutters, Wire strippers, ...]`, `Requirements.svelte` iterated it character by character into `{#each tools as tool (tool)}`, the repeated characters are duplicate keys, hydration crashed, and the page rendered blank with a 200 and correct-looking SSR HTML. Two fixes:

- Reworded the lede here, and on `lab/operation-target-10ppm-2026-04-25.md`, which a scan showed is the **only other page in `src/content` whose frontmatter throws**. Rewording rather than quoting is how the same fault was fixed when it hit the camera lamp pages on 2026-09-13. Both now parse; nothing else in the tree is running on the fallback.
- Dropped the key from the tools `{#each}` in `Requirements.svelte`. It is a plain list of strings that needs no key, two pages could legitimately list the same tool, and the key is the only reason a bad value blanks a page rather than just looking wrong. Comment on the line records why.

Verified by screenshot: the page was blank on the preview before this and renders with the tools list after.

### Follow-up 4, 2026-09-14: the stepper table is confirmed against a real board

Bill posted close-ups of his assembled board: https://discord.com/channels/1430279849171222682/1548985579860009040/1549099098488053884

The silkscreen is legible, and it **confirms the driver-to-designator table on this page against hardware for the first time**. It was read off the v1.3 netlist and had never been checked. Left to right along the bottom edge: `Stepper_A5` with the 2.54 mm header `J36` and the white socket `J35`, `Stepper_A4` with `J32`/`J31`, `Stepper_A3` with `J28`/`J27`, `Stepper_A2` with `J24`/`J23`. So the odd designator is the socket and the even one is the pin row beside it, exactly as the page says, and `Stepper_A6` is indeed on its own elsewhere.

It does **not** settle which motor each driver actually drives; that is firmware, still read from `hwcfg_basically_v1_2.h`, and still the thing `steppers.yml` disagrees with.

One clause added from the same photo: **each pin's coil name is printed beside the 2.54 mm row**, so you read them off the board instead of counting positions.

It started longer and Bill was right to push back on it. The coil rule, the pin-by-pin footprint and the multimeter check are already a callout two lines below and on [stepper connectors](https://docs.basically.website/hardware/electronics/steppers/); the driver names and the channel table are already on preparing the control board with Spencer's photos. The only thing nowhere in the docs was that the board prints the names, so that is all it says now. I also dropped the `B2 B1 A1 A2` order I had quoted: that is one photo of one board and the sequence reads either way depending which end you start from.

### Checks

`validate_frontmatter.py`, `validate_images.py` and `npm run check` pass (the four frontmatter errors it prints are pre-existing on other pages). `validate_credits.py` is unaffected: the page has no photos, only inline SVG. Rendered on `npm run dev` and read back.

### Known gaps, marked on the page

- The page carries the AI-generated first draft banner: the sockets are real, the order of the steps is not build-checked.
- Whether each socket is reachable with the housing cover on is not recorded, so the page says to take the cover off if it will not reach.
- Where the Orange Pi's fan lead lands is still open, as on the mount page.


basically-balloon:

basically-balloon:

basically-balloon:
**Daddy-O's Bricks - Bill** approved this in Discord.

That is 2 of 2 approvals.

basically-balloon:
**Merged.** Approved in Discord by barthel, Daddy-O's Bricks - Bill, and it only changes files this repo allows to merge that way, so the board merged it.

PR #654, opened 2026-09-14
parts: state 24V on the LED strip and warn about the 12V variant

basically-balloon:
**Requested by ReveryX (@reveryx) [from Discord](https://discord.com/channels/1430279849171222682/1535802933768364052/1549080694175830117):** *"at least in germany, the amazon US link for the COB LED redirects to a very expensive (55€) 12V variant on amazon.de - add the expected voltage to the details section in the docs, like the "Buy One 5m" and "Colour temperature 6000K daylight white"..."*

**Then [asked to check the density](https://discord.com/channels/1430279849171222682/1549080694175830117/1549082701209018399):** *"most offers seem to have 480 LEDs/m, check this and also add it to the details if correct.. 9W/m or similar could also be interesting, but it is not mentioned everywhere"*

**Then [cut the note back](https://discord.com/channels/1430279849171222682/1549080694175830117/1549085541360865364):** *"the yellow note contains way to much information, "Buy a 24V roll (5m, 6000K)" or similar should be enough"*

**Then [four more](https://discord.com/channels/1430279849171222682/1549080694175830117/1549091518541991949):** *""power: about 12W/m" is enough, no details per lamp needed here. and CRI 95+ is IMO not a hard requirement, I ordered a CRI 92 version, it will probably make no noticeable difference? please also add lumen per meter, this is likely (next to the 6000k) the second most important fact for the camera image quality... also change the "cuttable every..." to "should be cuttable every..., check the marks printed on the strip"."*

**And [barthel](https://discord.com/channels/1430279849171222682/1549080694175830117/1549523485066534913) supplied the EU source:** *"I ordered these. https://www.amazon.de/dp/B0BVVSLPDN ... These should actually fit."*

The part is called `led-strip-24v` and named "White/daylight LED strip - 24v", but **the voltage was nowhere in the details block** a builder actually reads on the part card or in the modal. Only the id and the name carried it, and neither is what you check before clicking Buy. Density, power and brightness were not there either.

## The card now

| Attribute | Value |
|---|---|
| Voltage | 24 V DC |
| Power | About 10 W/m; the listing says 12 |
| Buy | One 5 m (16.4 ft) roll, the 24 V size option |
| Colour temperature | 6000K daylight white |
| Brightness | About 1100 lm/m, not stated by the seller |
| CRI | 95+ claimed. No minimum is set for the machine, and 90+ is normal for this type |
| LED density | 320 LEDs/m |
| Width | 8 mm, COB (dotless) |
| Cut pitch | Check the marks printed on the strip. The listing says every 1.5 in (38 mm); 24 V COB is usually 50 mm |

And the whole `note`, which renders in the warning colour:

> Buy the 24 V roll: 5 m, 6000 K, 8 mm COB. The same listing also sells a 12 V one, and outside the US the link may not land on this product at all.

## Why the link cannot carry the spec

- **`B0DHV8ZF4R` does not exist on amazon.de at all.** `https://www.amazon.de/dp/B0DHV8ZF4R` returns Amazon's "The Web address you entered is not a functioning page on our site". The 12 V roll a German builder lands on is Amazon bouncing a non-US visitor to a lookalike of its own choosing. There is no URL fix for that.
- **The US listing sells a 12 V roll too.** Its Size selector is `16.4ft 12V` / `16.4ft 24V` / `32.8ft 24V`, all $9.99. The wrong strip is buyable without leaving the linked page.

## The listing is not a reliable source for its own specs

It states LED density five times on one page and disagrees three ways:

| Where on the page | LED density | CRI |
|---|---|---|
| "About this item" bullets | **320 lamp beads per meter** | CRI 95+ |
| A+ body text, "Ultra Bright" | **320 lamp beads per meter** | 95+ |
| A+ comparison banner | 384 lamp beads per meter | — |
| "Specifications" box | 384 leds/M | 95+ |
| "Why buy this Strip Lights" box | 120 lamp beads per meter | 90+ |

320/m is the best supported: the bullets and the body text are the two product-specific places. The 384 and 120 figures sit in the block that also says "Working Voltage: 12V/24V" for a roll sold as one or the other, i.e. boilerplate reused across the seller's range. **480 LEDs/m rolls are a different product** (BTF-LIGHTING `16.4FT 480LED/m`, $14.99, from the also-viewed rail).

## Brightness and power come from the class datasheet, because VOEWT publish neither

**VOEWT state no lumen figure anywhere on the listing.** The manufacturer datasheet for exactly this construction — [ledsuntech ST-FCOB](https://ledsuntech.com/wp-content/uploads/2024/07/ST-FCOB-12V-24V-320leds-white-cob-led-strip-specification.pdf), model `FCOB-24V-D320-8mm`, 24 V / 8 mm / 320 LEDs per m / Ra>90 — gives **10 W/m ±10%**, and for the 6000 K variant **110 lm/W and 1100 lm/m**, with a **50 mm minimum cut**. ENER-J sell the same spec (24 V, 320/m, 6000 K, CRI 90-92) and also state 10 W/m and 1100 lm/m.

Three consequences, all in the diff:

1. **Brightness is "About 1100 lm/m, not stated by the seller"** — flagged so nobody reads an inferred figure as a measured one.
2. **Power is now "About 10 W/m; the listing says 12".** The previous revision of this PR said 12 W/m and argued that the page's own 100 W transformer recommendation for 16.4 ft corroborated it. **That argument was weak** — 100 W is a round number over 50 W just as easily as over 60 W — and two independent sources at this exact spec say 10. At 10 W/m a 950 mm ring is about 9.5 W and 0.40 A at 24 V. The requester asked for "about 12W/m"; this is the one place the change goes against the request, and it is called out in the thread for them to overrule.
3. **Cut pitch leads with the printed marks.** The datasheet gives 50 mm minimum cut for the 24 V 8 mm version and 25 mm for the 12 V one. VOEWT's 1.5 in (38 mm) matches neither. **950 mm divides exactly by 50 and not by 38**, so the lamp ring the docs specify only lands on a mark if the real pitch is 50.

## CRI is not a requirement and no longer reads as one

The requester ordered a CRI 92 roll and asked whether that matters. **Nothing in sorter-v2, in any meeting or call transcript, or in any note sets a CRI requirement.** Every CRI figure in this catalog came from this Amazon listing. So `CRI` is split off `Colour temperature` onto its own line and says so plainly. CRI 90-92 is the mainstream part at this spec.

Whether 92 versus 95 is visible to the camera is **untested and this PR does not claim it either way.** The only adjacent datapoint: a CRI 80 strip of this construction runs 125 lm/W against 110 for the CRI 90+ one, so lower CRI buys efficacy; between 92 and 95 that difference is negligible. Setting an actual CRI floor for the classification image would be a call for whoever owns the vision side, not something to bury in a catalog attribute.

## There is an EU source now

barthel ordered [`B0BVVSLPDN`](https://www.amazon.de/dp/B0BVVSLPDN) (BERIXDEEP) off amazon.de and asked whether it fits. It does, on every dimension that matters, so it is in as a second `sourcing.vendors` row (`region: EU`, `vendor: Amazon DE`, `currency: EUR`) rather than staying an open gap in this PR body.

**Spec-verified from the listing, not from a roll in anybody's hands.** Nobody has received one yet.

| | US roll (VOEWT) | EU roll (BERIXDEEP) |
|---|---|---|
| Voltage | 24 V | 24 V |
| Width | 8 mm | 8 mm |
| Type | COB | COB |
| Colour | 6000 K | 6000 K |
| IP | IP20 | IP20 |
| LEDs/m | 320 | 384 |
| Power | ~10 W/m | 9-10 W/m |
| CRI | 95+ claimed | 92 |
| Roll | 5 m | 10 m |

The three rows that differ are exactly the ones this PR already established are not matching requirements: density is not one, and no CRI requirement exists anywhere in the project.

**That listing contradicts itself as well**, in the same shape as the US one. Its A+ "Specifications" box claims **IP44** and 10 W/m; the title says 9 W/m; BERIXDEEP's own comparison table, the product-specific one, says **IP20, 8 mm, 6000 K, 384 LEDs/m, 10 m, no PSU, no remote, EUR 16.68**, which is the variant barthel linked. **IP20 is the one that had to go the right way**: a coated strip fights a solderless clamp-on connector, and a coating is also what the FCOB datasheet warns changes colour temperature and flux.

Three things carried in the vendor `note`, which renders as small print under the row rather than in the warning box:

- **The size selector offers 10, 12, 15 and 30 mm widths** beside the plain lengths. Only 8 mm goes under the reflector hooks and into the connector. The default is the 8 mm one.
- **It is a 10 m roll**, so one covers two machines. The part's `stock` block still describes the US 5 m roll; `buyUnits()` returns 1 roll for a machine either way so nothing miscounts, but the `unit_label` is US-shaped and a reviewer may want that changed.
- **No free connector.** One bullet claims the pack contains a DC cable and a solderless connector; "Package includes" says strip only and the comparison table says no PSU. Not something to promise a builder.

At 95 lm/W and 9-10 W/m it is roughly 855-950 lm/m against about 1100 for the US roll, so 15 to 20 percent dimmer per metre. That should be immaterial: LED brightness is PWM per channel, and Spencer cut classification-chamber brightness substantially in June 2026 with a series resistor.

**Price recorded is the struck-through EUR 19.63**, not the EUR 16.68 it showed on a 15 percent limited-time deal when checked, so the row does not go stale next week.

## What is not in here

- **No docs body copy changed.** The [LED drop page](https://github.com/basicallysource/sorter-v2/blob/main/docs/src/content/hardware/electronics/led-drop.md) says "about 0.5A" per lamp, which was an estimate and is now closer to 0.40 A; it is within "about" and changing prose was not asked for, so it is flagged here rather than edited.
- **Provenance is in `internal_notes`, not the note.** That field is agent-only and not emitted to either site (`grep -c internal_notes catalog.generated.json` → 0), which is where the listing's contradictions, the datasheet figures and the CRI finding live now.

`python3 catalog/generate.py --metadata-only` run after each change and `catalog.generated.json` committed alongside; pin, versioning and connection checks pass.


basically-balloon:
ReveryX marked this ready for review in Discord.

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #613, opened 2026-09-09
catalog: the cable clamp's long screw is 5 mm too short (M3 x 35 -> M3 x 40)

basically-balloon:
**Requested by Daddy-O's Bricks - Bill (@daddyosbricksbill) [from Discord](https://discord.com/channels/1430279849171222682/1487518366020276397/1547318343403438102):** *"the docs say to put a m3 nut into the bottom of the outer cable clamp then install it onto the top interface chute mount. When I add the ribbon cable to the inner cable clamp and try to pick up the nut with the 35mm screw is doesn't reach the bottom to grab it. ... I am not sure if this should be flagged as a CAD issue, a hardware issue, or a process issue in the docs."*

**Also requested by barthel (@uwe_barthel)**, twice in the same thread. On the screw, [from Discord](https://discord.com/channels/1430279849171222682/1547318343403438102/1547321717771341994): *"The M3 x 35mm screw is only used here. To make assembly easier and more secure, we could replace this screw (which is used in only one place) with an M3 x 40mm screw or make a slight adjustment to the component. I'd prefer to replace the screws."* And on the nut, [from Discord](https://discord.com/channels/1430279849171222682/1547318343403438102/1547339749226713220): *"Add a warning about the loose m3 nut in top interface page step 11."*

The screw is too short. It is a hardware spec problem, not a defect in either clamp.

## Measured

Both clamp masters are exported in the same assembly frame, so the two halves stack directly. Pinned STLs, unchanged since the report:

- `cable-clamp-inner` (`3fue`): https://assets.basically.website/sorter-parts/cable-clamp-inner-16cfdfea28d0.stl
- `cable-clamp-outer` (`50dv`): https://assets.basically.website/sorter-parts/cable-clamp-outer-dd778421784b.stl
- `top-interface-split-spur-gear-3` (`Top interface chute mount`): https://assets.basically.website/sorter-parts/top-interface-split-spur-gear-3-11f50ca23d80.stl

Along the screw axis, at X −76.60 / Y +15.85:

| Z | feature |
| --- | --- |
| 204.62 | flat underside of the inner clamp, where the head bears. Plain Ø3.50 bore opening flush, no countersink, no counterbore. |
| 219.37 | inner/outer mating plane. Ø3.50 through the inner clamp for 14.75 mm. |
| 236.62 | floor of the nut pocket in the outer clamp. Ø3.50 through the outer clamp for 17.25 mm. |
[...]

PR #660, opened 2026-09-16
catalog: say why the cable clamp screw is 40 mm (it spans the nut pocket)

basically-balloon:
The catalog records *that* the Cable clamp (outer)'s nut pocket was enlarged in v2, and separately *that* the screw went from M3 x 35 to M3 x 40 in #613. It does not record that the second follows from the first, so nobody reading the catalog later can tell why 40 mm is the number.

**Reported by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1549635852035031051/1549647654303305849):** *"The M3x40mm was added because the clamp was modified and the hole for the nut was wider and the nut doesn't stay in place. The 40mm goes to the bottom of the hole and catch the nut. 35mm was too short."*

One edit, to `cable-clamp-outer`'s **`internal_notes`**: its own v2 pocket change is what made the longer screw necessary, credited and dated. No geometry, no lines, no builder-facing copy, so there is nothing to stamp.

## Why 8 mm and 40 mm

Measured off the two masters along the screw axis at X -76.60 / Y +15.90, with the head bearing face as zero:

- **32.00 mm** floor of the nut pocket, where the screw arrives
- **34.40 mm** far face of an M3 nut (2.40 mm) lying on that floor
- **40.00 mm** mouth of the pocket, against the chute mount

So the nut has 5.6 mm of depth to wander in, a 40 spans the whole pocket and catches it wherever it sits, and the M3 x 35 reached only 3.00 mm in. That matches barthel's account exactly. The stack numbers are from [#613](https://github.com/basicallysource/sorter-v2/pull/613), which is where the screw was changed, following [Daddy-O's Bricks - Bill's report](https://discord.com/channels/1430279849171222682/1487518366020276397/1547318343403438102) that the 35 would not pick up the nut.

## The screw's own row is deliberately left alone

The first version of this branch also put the reasoning into `scr-m3-40-fhcs`'s `description`, where builders read it. **barthel objected and he is right** ([from Discord](https://discord.com/channels/1430279849171222682/1549635852035031051/1549650193073442877)): *"Why is this relevant in the BOM and to someone who doesn't know that a different screw length was used here previously? Isn't the BOM supposed to list the current inventory, not tell stories?"* A catalog row says what the part is and where it goes; the reasoning lives in the internal note, the version changelog and here. Reverted, so the diff is now the internal note only.

The existing warning callout in [step 11](https://docs.basically.website/hardware/assembly/distribution/top-interface/#step-11) already tells the builder the nut is not captive, which is the part they can act on.

`catalog/generate.py --metadata-only` run and its three self-checks pass (115 pinned parts, versioning discipline, 60 connections). `npm run check` is clean, 0 errors 0 warnings.


basically-balloon:
Closing on barthel's call: the history is of no interest even in `internal_notes`, and there are few people left who know about the old screw. The diff was one line in a field neither site renders, so nothing builder-facing is lost. Branch `balloon/nut-pocket-why-40mm` kept.

The reasoning itself is not lost either: it is in [#613](https://github.com/basicallysource/sorter-v2/pull/613), in `cable-clamp-outer`'s own v2 version message, and in [the thread](https://discord.com/channels/1430279849171222682/1549635852035031051/1549647654303305849) where barthel gave it.

Issue #648, opened 2026-09-13
Deepen the M3 insert boss in the Cable cage bracket (cable mount) from 4.0 to 6.0 mm, so the standard M3 insert seats and the M3 x 10 screw can be dropped


basically-balloon:
`interface-cage-bracket-cable` (NU2N) carries one M3 heat-insert boss, and it is **Ø4.2 × 4.00 mm deep, blind**. Every other M3 insert boss on the machine is **Ø4.0 × 6.00 mm deep**, which is the depth for the ruthex **RX-M3×5.7** that both the docs and `parts.json` call the default M3 insert. This boss is 1.7 mm shallower than the insert is long, and 4.00 mm is exactly the length of the short **RX-M3S×4.0**, which was retired from the catalog on 2026-09-11 as no longer used anywhere.

Nothing is broken today: builders press the standard 5.7 mm insert in and it melts its own way past the modelled floor. But the shallow boss caps the screw length, and that is what keeps a one-off M3 × 10 in the BOM.

## Reported

**barthel** (`@uwe_barthel`), 2026-09-13 in `#balloon-general` ([message](https://discord.com/channels/1430279849171222682/1487518366020276397/1548577612123738173)):

> It's possible to use a M3 x 12mm screw to mount Cable cage bracket (cable mount)  and Ribbon cable clamp  together and remove the M3 x 10mm (only one used here) and reduce the amount of different screw sizes?

Measuring that turned up the boss depth. barthel then confirmed the insert side from his own build, 2026-09-13 ([message](https://discord.com/channels/1430279849171222682/1548577612123738173/1548581410279919637)):

> The documentation specifies an M3 heat insert, not an M3S. I personally used a standard M3 heat insert as well, and it worked fine. There seems to be enough space to make this adjustment. To minimize the variation between M3 heat inserts and M3 screws, I would recommend adjusting the 3D-printed part. Create a CAD issue and describe the modification and the rationale behind it.

He is right about the documentation: `hardware/assembly/distribution/top-interface` lists `hsi-m3` × 16 for the section and no `hsi-m3s` anywhere, and the prep step says plainly "the Cable cage bracket (cable mount) takes 1 × M3 insert".

## Measured

Masters pinned in the catalog today, both exported in the same assembly frame so their coordinates compare directly:

- `interface-cage-bracket-cable-0dad6ad3b408.stl` (NU2N, v2)
[...]

Issue #618, opened 2026-09-10
parts-calculator: delete the unused scr-m3-35-bhcs catalog part


basically-balloon:
`scr-m3-35-bhcs` ("M3 × 35 mm flat head") in `parts-calculator/catalog/parts.json` (uid `38vn`) reaches no bill of materials once [#613](https://github.com/basicallysource/sorter-v2/pull/613) lands. It was the long screw clamping the two cable-clamp halves together at the top interface, and it was 5 mm too short for that joint: the stack under the screw head is 32.00 mm from the bearing face on the Cable clamp (inner) to the floor of the nut pocket in the Cable clamp (outer), and 34.40 mm to come through a 2.4 mm nut sitting on that floor, so 35 mm carried 0.60 mm of margin and builders could not pick the nut up. `cable-clamp` v2 replaces it with `scr-m3-40-fhcs` ("M3 × 40 mm flat head", uid `ucgw`).

That was its only live use. `chute-stepper-gearing` v1 also listed it, but v2 replaced it with `scr-m3-20-cs` some time ago, so the only remaining references after #613 are historical version snapshots of superseded assemblies, which are expected and correct to leave alone. It has no `sheet_qty`, so it contributes nothing to any BOM.

I can't delete the part object myself: `parts-calculator`'s rule is that a minted uid stays resolvable forever, and `check_generated_pins.py` refuses a change that drops one from `parts.json` ("uid(s) removed from parts.json ... mark them superseded/rejected instead"). That is a CI-enforced invariant, not a call to make from a Discord thread. Same situation as #453, #454, #478, #536, #563, #566, #567 and #568.

So this needs someone who can decide whether to override or adjust that check for this uid, then delete the part object from `catalog/parts.json` and re-run `generate.py`. Worth noting that `38vn` is pinned inside `chute-stepper-gearing` v1's line snapshot, so whoever does it should check the snapshot still resolves, or accept that it will not.

In the meantime the part's `description` and `note` say it is retired and unused, and a P4 `changes` entry (`delete-unused-m3-35-flat-head-screw`) puts a visible marker on it wherever the calculator renders it. Both land in #613.

[...]

PR #661, opened 2026-09-16
catalog+docs: the ribbon clamp and cable clamp M3s are 12 mm, not 10

basically-balloon:
The single M3 holding the Ribbon cable clamp to the Cable cage bracket (cable mount) is an M3 x 10 in the catalog and the docs. It is too short, because the clamp is drawn standing off the bracket face so the ribbon is not crushed. Two builders have now hit it.

**Reported by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1549635852035031051/1549648206676361256), with a photo of the joint:** *"The cable cage bracket cable mount needs an m3x12 not 10, As the bracket stays away to prevent the cable from being crushed."*

**Asked for first by barthel (@uwe_barthel) [on 2026-09-13](https://discord.com/channels/1430279849171222682/1487518366020276397/1548577612123738173): *"It's possible to use a M3 x 12mm screw to mount Cable cage bracket (cable mount) and Ribbon cable clamp together and remove the M3 x 10mm (only one used here) and reduce the amount of different screw sizes?"* He wanted the size dropped from the BOM. That round measured the joint and stopped short of changing it, because as drawn the boss is only 4.00 mm deep and a 12 bottoms out exactly as the head touches. barthel then confirmed the docs specify the standard ruthex M3 x 5.7 insert, not the short one, and that a standard insert presses in fine past the modelled floor, which is what makes the 12 seat. The shallow boss is tracked as [#648](https://github.com/basicallysource/sorter-v2/issues/648) and is not touched here. Alice's build is the confirmation that the 12 works in practice.

## Measured, both joints

Off the masters in the shared export frame.

**Ribbon clamp to bracket** (`ribbon-cable-clamp` NF0N, `interface-cage-bracket-cable` NU2N), axis Y 197.6 / Z 239.7: 6.64 mm of clamp under the head, plus the 1.30 mm the clamp stands proud, so **7.94 mm to the bracket face**. An M3 x 10 leaves **2.06 mm** in the insert, under 1 x D.

**Cable clamp (outer) to chute mount**, the other joint on the same page: the clamp's two lobe holes are Ø3.44 through 7.10 mm of tab (Z 237.52 to 244.62), and the chute mount has a **Ø4.20 bore 5.77 mm deep** at each (Z 244.62 to 250.39, measured this turn off `top-interface-split-spur-gear-3`), which is a standard-length insert. So **12.87 mm** is usable: an M3 x 10 gets 2.90 mm of thread, an M3 x 12 gets 4.90 mm and stops 0.87 mm short of bottoming.

The catalog already says M3 x 12 for that second joint (`interface`, 2 off). **Only the docs page still said 10**, so this aligns it rather than changing it.

## The change

- `cage-bracket-cable-clamp` line and connection: `scr-m3-10-shcs` to `scr-m3-12-bhcs`. Stamped **v2** (uid `frol`, superseding `s4we`, `breaking: false`, `commit: null` because the board squashes). The v1 snapshot pins `scr-m3-10-shcs` as `ufvb`.
- `scr-m3-10-shcs` **retired in place**: nothing live uses it now, it only survives inside that v1 snapshot, so the uid has to stay resolvable per VERSIONING.md. Its stale `sheet_qty: {per_machine: 3}` and the note explaining that 3 are deleted; both had been wrong since the cable clamp moved to `scr-m3-40-fhcs`.
- `top-interface.md` step 11: both M3 x 10 references become M3 x 12, and the ribbon-clamp line says why the screw has to cross the gap. `parts_needed` swaps `scr-m3-10-shcs` (3) for `scr-m3-12-bhcs` (3).

No geometry changed and no STL moved, so nothing to re-slice.

**`scr-m3-10-shcs` is marked to be removed.** Asked for by barthel [in the PR thread](https://discord.com/channels/1430279849171222682/1549651256866242591/1549652408529850379): a P4 `changes` entry `delete-unused-m3-10-shcs-screw` with `condition: "retired"` and `targets.hardware` puts the "To Be Removed" badge on it wherever the calculator renders it, its `note` points at the issue, and [#662](https://github.com/basicallysource/sorter-v2/issues/662) (`bom`) tracks the actual deletion, which needs `check_generated_pins.py` overridden by a person because `cage-bracket-cable-clamp` v1 still pins uid `ufvb`. Same shape as [#618](https://github.com/basicallysource/sorter-v2/issues/618) for the M3 x 35.

`generate.py --metadata-only` passes its three self-checks (115 pinned parts, versioning discipline, 60 connections), `check_versioning.py --base origin/main` passes, `npm run check` is clean on both apps (docs 0 errors with the 4 pre-existing `Requirements.svelte` a11y warnings; calculator 0/0) and `validate_images.py` passes.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #671, opened 2026-09-17
docs(top-interface): her build photos on step 8 and the cable cage steps

basically-balloon:
**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550060075350229053):** *"27 top interface step 8. The hammer needs to strike the limit switch from both directions but not run on the dowel pin."*

Her photo 27 (`IMG_4327`) is a third figure on step 8: straight down on the ring gear with the hammer at the roller lever, which is the angle that shows the gap to the dowel pin.

The step already described both-directions-and-not-the-pin, buried in the middle of a long paragraph. It is now its own bolded sentence, in her words, since that is the whole point of the adjustment.

**And [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550067098720542721):** *"91-99 top interface installing the cable cage. 98 shows how the cable need to be with a chute core sitting on top, about 8cm, the basically adapter board isn't fitted but mounts on those 4 heat inserts shown. 97 is how the cable goes around but the cable is actually facing the wrong way, you should be able to see the pins, not have the pins facing the table."*

**Step 10** gains the cable mount bracket at its corner (93, 92 — the M3 insert in the hanging face is the one the clamp uses) and the cage top slotted over the chute mount from three angles (94, 95, 96).

**And [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550073990645682327), after barthel spotted it:** *"loose this photo."* Both cable mount bracket photos (her 93 and 92) are out of step 10. They show **`interface-cage-bracket-cable` v1**; v2 landed on 2026-09-07 in [#403](https://github.com/basicallysource/sorter-v2/pull/403) and raised the underside near the heat-insert boss by about 2 mm. She named one, I dropped both, because 92 is the same bracket from the other side and would be just as stale.

**Step 11** gains the clamp on the bracket's post before the ribbon (91), the ribbon clamped and folding back (99), and the tail left standing for the chute (98), plus two things from her message that the step did not say:

- **"Leave a significant tail" is now about 80 mm**, which is what it takes to reach the layer adapter board on the chute core above.
- **The connector goes on pins up.** If the pins face the bench, the cable is on the wrong way round.

**And [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550067321500991511):** *"when mounting the cage, every single screw check the chute still turns, it can jam easily especially if the cage is handcut or the screws are over tightened."* That is a warning callout on step 10, and step 12's existing "should still rotate" line now says to check per screw and back one off rather than force the chute.

**And [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550071686055854122):** *"Can you make a rendering for 97, using my image as the example. But the plug facing the other way. It's a critical step and if you get it wrong to have to undo 3 steps to fix it up."* Step 11 now carries a render of the free end of the ribbon with the socket's opening facing up, in place of her 97, and the prose says what undoing it costs.

**97 is deliberately not in the page.** It is the photo she says has the cable the wrong way round, and a caption does not rescue a photo of a build that is wrong — her own rule, in `docs/AGENTS.md`. The information in it is in the prose instead. Say the word if you would rather have it in with an explicit "this is what wrong looks like" caption.

I have also taken 26 and 28 off the outstanding list. **27 stays on step 8** below, since she asked for it there at 08:25, half an hour before "26-28 not needed"; if that was meant to retract it too, this PR is where to say so.

Photo uploaded to `sorter-docs` with the EXIF rotation baked in. `npm run check` 0 errors, `validate_images.py` clean.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #624, opened 2026-09-10
docs: bottom interface gains the layer connectors, and C-channel gets its gear photos

basically-balloon:
Two more pages get photographs from BrickCycleAlice's build, and the C-channel page loses a step that was in the wrong place.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1547526128942583838/1547573759299420183):** "Let's call this photo 89. This will be a new pr, the pages it will cover bottom interface and c channel", and closed [here](https://discord.com/channels/1430279849171222682/1547526128942583838/1547579703576100976): "that's it for that pr for the two pages."

## Bottom interface

- **A new step 4, "Add the two layer connectors"**, [asked for here](https://discord.com/channels/1430279849171222682/1547526128942583838/1547575999384784916): "use image 89 to show how two layer connectors are added. They will be a step after step 3. This completes the bearing stack." The old steps 4 to 8 become 5 to 9, and the sentence about the parts list moved with them.
- The page's parts list gains the two connectors and their screws, which it did not carry. One {% raw %}M3 × 8{% endraw %} countersunk per connector, because the chute mount has one M3 insert per side face where the chute core has two per connector: [confirmed](https://discord.com/channels/1430279849171222682/1547526128942583838/1547578435331948566), "Screw length is correct".
- **Photo 90** joins the M3 pocket pair in step 1, [her words](https://discord.com/channels/1430279849171222682/1547526128942583838/1547575553681391687): "it is a the heat inserts installed on the chute mount". The page already showed that tab with an empty pocket and a render of it; this is the same tab with the insert in.

## C-channel

[Asked for here](https://discord.com/channels/1430279849171222682/1547526128942583838/1547578435331948566) and [here](https://discord.com/channels/1430279849171222682/1547526128942583838/1547579381516476426).

- **Step 1's two `Image coming` placeholders are filled**: the 130T output gear with its 6806 pressed in, and the 24T idler with its 608.
- **Step 4** gains a close view of the NEMA bracket. I could not tell the part from the frame and captioned it plainly; she [named it](https://discord.com/channels/1430279849171222682/1547526128942583838/1547579526249193542) and the caption now says so.
- **Step 5** gains the NEMA bracket fastened under the stator, and its closing photo is replaced by a newer finished stage.
[...]

PR #517, opened 2026-09-01
bottom interface: add the 2 M3 heat inserts the layer connector needs

basically-balloon:
The Lazy Susan chute mount takes **two M3 heat inserts** that were in neither its parts list nor the page's Preparation step.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1544432803293438012):** *"Wir werden die layer connectors an der Lazy Susan — chute mount an der Seite montiert? Werden sie verschraubt und braucht es da auf der chute Seite heat inserts?"*, then [asked for the fix](https://discord.com/channels/1430279849171222682/1544432803293438012/1544434803703549962): *"Add the heat inserts to the required parts and, during the preparation step, to the bottom interface side."*

## What the geometry says

Measured off the published `ls-mount-to-chute` STL. It has two blind Ø4.20 × 5.70 mm pockets on its side faces, one per side, at **(13.22, ±69.15, 11.42)** in the assembly frame, opening outward on a face at y = ±72.0. That is the ruthex RX-M3x5.7 signature (`hsi-m3`), and it is exactly the same x/y as the [chute core](https://docs.basically.website/hardware/assembly/distribution/chute/chute-core/)'s own layer-connector pockets, which sit at (13.22, ±69.15, 42.77) and (13.22, ±69.15, 171.49) on the same axis.

So the joint is the one [Layer connectors](https://docs.basically.website/hardware/assembly/distribution/chute/layer-connectors/) already describes, with the chute mount playing the part of the core below: the connector's two Ø3.40 clearance holes are 31.858 mm apart, one screw goes into the bottommost chute core's lower pocket and the other into the mount. Screws are the chute's own, 2 of its 8 × `scr-m3-8-cs`, so no screw count changes here.

## Changes

- **`parts-calculator/catalog/parts.json`** — `ls-mount-to-chute` now requires 2 × `hsi-m3` alongside its 4 × `hsi-m4`, with the measurement in `internal_notes`. `catalog.generated.json` regenerated (`generate.py --metadata-only`); `check_generated_pins.py` and `check_asset_urls.py` both pass.
- **`bottom-interface.md`** — `hsi-m3` × 2 in `parts_needed`, and a Preparation item saying they are on a different face from the four M4, with two figures: a render locating the pocket, and a photo of it on a printed part.

[...]

PR #672, opened 2026-09-17
catalog+docs(layer connectors): N+1 pairs, and the top interface carries none of its own

basically-balloon:
**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550061375576080495):** *"64 goes into layer connectors, for now. We are only putting one set on each chute core at the top. One set on the top of the bottom interface. Then as the build occurs they slot on and get screwed down."*

**And [from Discord](https://discord.com/channels/1430279849171222682/1550036662069886987/1550064654661976086), answering the count question this PR originally left open:** *"The top interface connects to the connector mounted on the very top chute, so n+1."*

### The page

Her photo 64 (`IMG_4526`) fills the `img-placeholder` that has been on step 1 asking for exactly this, and it shows the thing her sentence describes: the connector stands proud of the top edge of the core, screwed through its **lower** hole, with the **upper hole empty** for the part that slots on next.

- **Where they go.** One pair on the top of each chute core, one pair on the top of the bottom interface, nothing on any underside.
- **How many screws go in now.** The page said each connector "takes 2 M3 × 8 screws through the flat strap", which reads as both into the core. One goes in now; the other hole is what the part above is screwed down into. That matches the STL geometry as well: a connector spans 31.86 mm between its two holes, and the core's own pockets are 128.71 mm apart, so one strap cannot take both screws into the same core.
- **N + 1, not N + 2.** One pair per layer plus one for the bottom interface. The top interface carries none of its own.

### The catalog, which had it the other way round

This is the question [#517](https://github.com/basicallysource/sorter-v2/pull/517) left open on 2026-09-01, and her answer settles it, so the count moves out of the docs and into `parts.json` where the quantities come from:

- **`interface` v4 → v5**: `layer-connector × 1` removed. The top interface lands on the topmost chute's pair, and `top-interface-split-spur-gear-3` measures no M3 pockets on any side face, so there was nothing for one to screw into.
- **`bottom-interface` v2 → v3**: `layer-connector × 1` added back. v2 dropped it in September on the reasoning that "the spare one belongs to the top interface", which is the assumption she has just corrected. `ls-mount-to-chute` carries the two M3 inserts it screws into, measured and photographed in #517, and `bottom-interface.md` step 4 has documented fitting it since #624.

The machine total is unchanged at N + 1; it was being counted against the wrong assembly. Neither version is `breaking`: nothing physical about either node changed. Both new version entries are left with `"commit": null` for whoever stamps after the squash.

Screws are deliberately untouched, same as #517: the pair's own screws come out of the parts they bolt to.

`npm run check` 0 errors, `validate_images.py` clean, and `check_versioning.py`, `check_generated_pins.py`, `check_connections.py` and `check_asset_urls.py` all pass.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

Issue #675, opened 2026-09-17
Image customizer fails for everyone: "config too large: N bytes, capacity 2"


basically-balloon:
## Summary

The browser image customizer at `setup.basically.website` fails for every user on the current release with:

```
Error: config too large: N bytes, capacity 2
```

`N` varies with what the user typed; **`capacity` is always 2**, so nobody can shrink their config past it. Two users have now hit it independently and both spent time assuming it was their download, their browser, or their SSH/Tailscale keys.

Only Wi-Fi-only installs are actually blocked: flashing the `.zip` unmodified and coming up over Ethernet works, so there is a workaround, but it is not discoverable from the error.

## Cause

`sorteros-setup/src/lib/img-patch.ts` scans the `.img` from byte 0, takes the **first** `__SORTEROS_CFG_START__` it finds, then the first `__SORTEROS_CFG_END__` after it, and sets `capacity` to the bytes between them.

In the released image that marker text occurs four times, in two pairs. Verified 2026-08-09 against `sorteros-v3.4.7-2026-05-30.zip` (1272317756 bytes):

| offset | what it is | capacity |
|---|---|---|
| `2374752408` | a `.pyc` constants table: `\xda\x16__SORTEROS_CFG_START__\xda\x14__SORTEROS_CFG_END__` | **2** |
| `3393847298` | the real placeholder, `# __SORTEROS_CFG_START__` + `\n` padding | 8147 |

The scanner hits the `.pyc` about a gigabyte before the real placeholder. Deterministic, not a race.

The `.pyc` is a byte-compiled `sorteros-firstboot.py`. That file deliberately writes the marker split across a concatenation so the literal never appears in the source:

```python
CFG_START_MARKER = "__SORTEROS_CFG" + "_START__"
```

That was the fix in `7c88aaf7` ("sorteros-v3.2.1: fix patcher finding markers in firstboot.py instead of config.toml"). **CPython constant-folds the concatenation at compile time**, so the full literal reappears in the `.pyc`, and marshal stores the two short interned ASCII constants adjacently: 1 type byte + 1 length byte = exactly the 2-byte gap. Reproduced by `py_compile`-ing the file and finding `gap=2` (`\xda\x14`). The source-level workaround does not survive byte-compilation.

## Suggested fix

[...]

PR #676, opened 2026-09-17
docs(sorteros): give the real UI address for each install path

basically-balloon:
**Requested by Daddy-O's Bricks - Bill (@daddyosbricksbill) [from Discord](https://discord.com/channels/1430279849171222682/1550174361464606760/1550285942059696290): "I just booted hard wired to the router and http://sorter.local:5173/ did not pull up the UI. I went in through my router and found the OrangePi and used its IP address to pull up the UI."**

He then [corrected the port too](https://discord.com/channels/1430279849171222682/1550174361464606760/1550292398914142309): *"on my set up I only used the ip address no additional :5173 at the end, with the port, I could not pull it up"*. He is right, and that is the second fix below. He also [reported the customizer failure](https://discord.com/channels/1430279849171222682/1437436132815274155/1550174361464606760) that sent him down the direct-flash path in the first place: *"I have extracted the file and filled out all of the fields and I get this error."*

Step 4 of this page had **both halves of its address wrong**. Neither the hostname nor the port was right for a reader following the page.

## The port is 80, not 5173

Both SorterOS units bind port 80:

```
overlay/etc/systemd/system/sorter-ui.service
ExecStart=__PNPM_BIN__ preview --host 0.0.0.0 --port 80

overlay/etc/systemd/system/sorter-ui-dev.service
ExecStart=__PNPM_BIN__ dev --host 0.0.0.0 --port 80 --strictPort
```

The dev unit even carries the comment *"Port 80 lets users hit `http://<host>.local` without a port suffix"*. `5173` is Vite's default and reaches the UI only under `./dev.sh` on your own machine, which is the [by hand](https://docs.basically.website/sorter/installation/by-hand/) and [generic Linux](https://docs.basically.website/sorter/installation/linux-generic/) paths. Those two pages use `5173` correctly and are untouched here; the new text says which port belongs to which path so the next reader does not carry `5173` across from one page to the other.

## `sorter.local` only exists if you used the customizer

`sorter` is not a default anywhere in the image. It is the initial value of the hostname field in the SorterOS Setup form (`sorteros-setup/src/routes/+page.svelte`, `let hostname = $state('sorter')`), and it only reaches the Pi through the config file that tool writes. `sorteros-firstboot.py` calls `hostnamectl set-hostname` only when that config carries a hostname:

```python
hostname = cfg.get("hostname")
if isinstance(hostname, str) and hostname.strip():
    log.info("setting hostname: %s", hostname)
    sh(["hostnamectl", "set-hostname", hostname])
```

Nothing in `build.py`, `chroot_apt.sh` or the overlay sets one otherwise, so a `.zip` flashed directly keeps the base Orange Pi Ubuntu image's hostname. The backend's own tests take that to be `orangepi5` (`backend/tests/test_server_security.py` uses `hostname="orangepi5"` and asserts `http://orangepi5.local:5173` is an allowed origin), which is where the address in the new text comes from.

**Flag for review:** I have not booted an unconfigured image to confirm `orangepi5` first-hand. If the base image's hostname is something else, that one word needs changing. The router-IP fallback below it covers the reader either way, and that is the route that actually worked for Bill.

Step 4 now splits by which path the reader took in Step 2, keeps the mDNS caveat, adds that Windows generally needs Bonjour, and ends with the router-IP fallback.

## Step 2 now says the tool is broken

Independent of the above, and **droppable on its own** if you would rather the docs not carry it: the page sends every reader to SorterOS Setup with no warning, and that tool currently fails for everyone with `config too large: N bytes, capacity 2` on `sorteros/v3.4.7`. Root cause and suggested fix are in #675. Two users have hit it and both assumed the fault was theirs.

The notice points at the issue and gives the Ethernet workaround, noting WiFi and Tailscale can be set from the Sorter UI's Settings page afterwards (`frontend/src/routes/settings/+page.svelte` has both sections; the WiFi one drives `nmcli` through `backend/server/routers/wifi.py`).

## Not fixed here

- `sorteros` arguably should set a hostname of its own so `sorter.local` is true for everybody. That is a change under `software/`, so not mine to make: filed as #677.
- `sorter/dev-flow.md` passes `-H "Origin: http://sorter.local:5173"` to the supervisor. That still works, since the backend accepts any `.local` origin and it is only a header, never a URL anyone browses to. Left alone as out of scope, but worth a look by somebody who owns that page.

## Checks

`npm run check` clean (810 files, 0 errors). `validate_images.py` clean. `validate_frontmatter.py` and `validate_credits.py` report only pre-existing failures on other pages, none on this one. Step 2's notice and Step 4 both read correctly on the deploy preview.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #678, opened 2026-09-18
docs: cover the blank-Pico case behind "No MCU buses found"

basically-balloon:
**Requested by Daddy-O's Bricks - Bill (@daddyosbricksbill) [from Discord](https://discord.com/channels/1430279849171222682/1437436132815274155/1550303158213943409):** "The setup wizard's controller discovery showed "No MCU buses found". I had to manually download the firmware v0.8.0.uf2 from the releases, mounted the Pico's bootloader drive manually on the orangepi and copied the .uf2 onto it. There was nothing in the documentation or troubleshooting that I saw for this. Not sure if I was an edge case based on the way I installed/booted or if there is a real issue."

He is not an edge case: every machine with never-flashed Picos hits this, and the docs currently describe the one flash path that cannot work on a blank board.

### What actually happens

`MCUBus.enumerate_buses()` filters `serial.tools.list_ports.comports()` to `vid == 0x2E8A and pid == 0x000A`, the Pico SDK CDC UART ([`software/sorter/backend/hardware/bus.py`](https://github.com/basicallysource/sorter-v2/blob/main/software/sorter/backend/hardware/bus.py#L273-L285)). A Pico with empty flash boots into its own USB bootloader and enumerates as `RPI-RP2` mass storage, never as that serial device, so the list is empty and `discover_control_boards()` raises `RuntimeError("No MCU buses found.")` ([`machine_platform/control_board.py`](https://github.com/basicallysource/sorter-v2/blob/main/software/sorter/backend/machine_platform/control_board.py#L228-L230)), which the wizard surfaces as an issue banner on step 3.

The supported path for that state already exists and is undocumented: the **Recovery flash** checkbox in Settings → Control board ([`ControlBoardSection.svelte`](https://github.com/basicallysource/sorter-v2/blob/main/software/sorter/frontend/src/lib/components/settings/ControlBoardSection.svelte#L542-L544)) skips the reboot-to-bootloader step, and `waitForBootloaderMount()` finds the `RPI-RP2` block device by label and mounts it itself ([`hardware/firmware_flash.py`](https://github.com/basicallysource/sorter-v2/blob/main/software/sorter/backend/hardware/firmware_flash.py#L130-L157)). So the manual download and manual mount Bill did by hand were work the UI would have done.

### The changes

- **`docs/src/content/hardware/assembly/software-setup.md`** — the "Flash the control boards" section now says to do this before the wizard, explains why a blank board is invisible to Controller Discovery, and adds a "A board that has never been flashed" subsection with the recovery-flash steps. The existing text ("flash it over the Pico's USB serial connection") is the path that only works once firmware is already on the board, so it stays but is no longer the only one described.
- **`docs/src/content/sorter/troubleshooting.md`** — a symptom-led entry for `No MCU buses found`, matching the page's what you see / cause / fix / verify shape, placed next to the existing `permission denied on /dev/ttyACM*` entry.

One board at a time is called out in both: a recovery flash writes to whichever board is in the bootloader and cannot disambiguate two of them.

### Not changed

Nothing under `software/`. Two things there are arguably worth a follow-up and are a person's call, not mine:

- `DiscoveryStep.svelte`'s empty state says "Check power and USB connections, then rescan", which sends people looking at cabling. It could say a board in `RPI-RP2` is present when `bootloaderPresent()` is true, the way the Control board settings section already does.
- The wizard has no link to the firmware flashing page from step 3.

`npm run check` clean (810 files, 0 errors), `validate_images.py` clean, both pages checked rendered on a local dev server.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #682, opened 2026-09-18
docs: helper page for fitting the Orange Pi 5 WiFi module

basically-balloon:
Adds a Helpers page for fitting the Orange Pi 5's M.2 WiFi module, and says at both ends of the build that it goes in while the board is still loose on the bench.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550463721992753234): "can you turn this last instruction into a helper page within hardware, as doing this before mounting the board would have been easier."** She is mid-build with the board already mounted, and asked first [whether the module goes in before or after the flash](https://discord.com/channels/1430279849171222682/1550454875698237472/1550462820691345550).

**Build detail added after opening**, from the same person, [from Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550464246457049119): *"the washer needs to sit between the pi and the wifi module for even tension, when bolting on"*. The slot holds the contact end of the module up off the board, so the washer under the free end is what keeps it level; screwed down without one, the module is bent. She has the parts in hand. She then [confirmed the washer is in the box](https://discord.com/channels/1430279849171222682/1550454875698237472/1550468254026440745), so nothing needs buying. `wifi-module-opi5` recorded only `Includes: Antennas`, which reads as "find your own washer", so that attribute is now `Antennas, retention screw and washer` and the page says the same. Regenerated with `catalog/generate.py --metadata-only`; all three checks pass.

### The sequencing, which is the point of the page

The M.2 slot, its retention screw and the u.FL antenna leads are all on the **underside** of the board, and `orange-pi-extrusion-mount` sits directly under it once the Pi is on its standoffs. Second reason, independent of the mount: the UI's WiFi section can only offer networks for an adapter that was present at boot (`software/sorter/backend/server/routers/wifi.py` driving `nmcli`, fronted by the Settings page), so a module fitted afterwards costs another shutdown and another boot.

Nothing said either of those. The procedure sat in a subsection of the Orange Pi 5 selection page, which is a page you read before buying, and the mount page, which is where a builder actually is when it matters, never mentioned the module at all.

### What is in the PR

- **New** `docs/src/content/hardware/helpers/wifi-module.md` at `/hardware/helpers/wifi-module/`: when to do it and why, checking the keying notch against the board variant before pushing, seating it, the antennas, and confirming the OS sees it.
- `hardware/orange-pi-5.md` "Installing the module" was the only copy of the procedure. It is now a pointer to the helper, so the two cannot drift. The module selection, the 5 vs 5 Plus connector difference and the driver notice all stay there, since those are buying decisions.
- `hardware/electronics/installation/orange-pi-mount.md` step 1 now names it alongside the Pi's heatsink fan, which is the other thing that has to go on while both faces of the board are reachable. This is the edit that would have saved the rework.
- `nav.yml` and the helpers index.

### Caveats

**Not photographed on a build.** The module seated in its slot is not in any build photo, so the page carries a `warning:` saying so and an `img-placeholder` at that step. The only image is the supplier listing photo already in the catalog for `wifi-module-opi5`.

**The procedure text is carried over unchanged** from `hardware/orange-pi-5.md` (author: spencer, who is credited as a contributor on the new page). Only the sequencing claim is new, and it comes from the build report above rather than from a bench check here. The page is authored to the requester per the usual new-page rule.

I have not claimed the module is impossible to fit with the Pi mounted, only that it is the wrong order. Whether the mount's central cutout leaves the slot reachable is not something I can tell from the STL alone.

### Is this redundant with `orange-pi-5.md#wifi`?

barthel [asked](https://discord.com/channels/1430279849171222682/1550454875698237472/1550477694532198460) whether the helper is redundant. Partly: step 1 restated the notch-by-eye explanation already on that page, so that is now a link to the one copy rather than a second one. The six install steps are **moved** out of `#wifi`, not duplicated.

The split stands on three things: `orange-pi-5.md` is `type: reference` under Parts in the nav, read before buying, and the install steps were the only how-to on it; the failure being fixed is sequencing, which needs a target for `orange-pi-mount.md` step 1 to point at that is not a buying page; and the washer, the adapter-present-at-boot requirement and the in-the-box fasteners are not on `#wifi` at all.

The person who hit this [put it better than I did](https://discord.com/channels/1430279849171222682/1550454875698237472/1550490187782426685): *"That wifi info is at the very end of the menu, I would not and did not look there before mounting the pi. Mira can't mount something if she hasn't bought it yet."* `Orange Pi 5` is the last entry in the whole Hardware sidebar, under Parts, and it is a `reference` page: the install steps sat at the one point in the docs a builder passes **before** they own the parts and never passes again.

**Still open, for a reviewer rather than me:** that fixes discoverability at mounting time, not at buying time. `wifi-module-opi5` could go in `orange-pi-mount.md`'s own `parts_needed` card, marked optional, so it is visible at the top of the page before any bolting starts. Not done here, because that card is supposed to mirror the calculator's assembly and the module is not in that assembly (it is a hand count carried from the BOM sheet). Say the word and it is one line.

Against it: the helpers index says helpers are for parts used in multiple places, and this one is used in one place. `pulley-gear-mod` has a single inbound link too, so there is precedent, but the index wording undersells what the section has become. Happy to fold it all back onto `#wifi` and close this if that is the preference.

### Checks

`npm run check` clean. `validate_frontmatter.py`, `validate_credits.py` and `validate_images.py` pass on every touched page; the 4 frontmatter and 8 credit failures they report are pre-existing on `main` and untouched here. Page rendered and inspected on `npm run dev`: the parts card resolves `wifi-module-opi5`, the step headings and the warning link render, and the nav and helpers-index entries appear.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #683, opened 2026-09-18
docs(orange pi mount): cut the redundant plug-it-in step and point at the connecting page

basically-balloon:
`orange-pi-mount.md` step 6 "Plug it in" repeated what [connecting the components](https://docs.basically.website/hardware/electronics/installation/connecting/) already covers, and covers better:

- **Pi power** (its own 24 V to USB-C buck off the PSU, not the control board, plus the 5 V and polarity check) is on the connecting page three times over: the cable table, the PSU terminal diagram, and step 7 item 1. Step 7 also warns that the Pi has two identical-looking USB-C sockets and only one is `PWR IN`, which the mount page did not say.
- **The USB hub and the cameras** were one sentence pointing at the wire harness page. All of step 7 on connecting.
- **Where the arm fan's lead lands** was the same open item as connecting's "What is not recorded yet", nearly word for word.

The mount page then closed with "Flashing and configuring the Pi is software setup", skipping the connecting page, which the installation index has as step 5 immediately after this one's step 4. A reader following the mount page to its end was sent two pages too far, and step 6 read as a plug-it-in step at a point where nothing can be plugged in yet: the PSU leads and the hub are not on the machine until the next page.

**What this changes**

- Step 6 is gone. The closing paragraph says the mount is complete, that nothing plugs into the Pi here and why (the power comes from a buck off the PSU, not the control board), and points at connecting as the next page.
- The fan's identity (24 V WINSINN 4010, XH2.54 2-pin, the same fan the control board housing takes, and not the Pi's 5 V heatsink fan) moves up to step 3, where the fan is screwed to the arm.
- The open item about where that lead lands moves to step 3 too, stated in full, next to the instruction to leave the lead loose.

No other page changes. `npm run check` and `scripts/validate_images.py` are clean.

Redundancy spotted by uwe_barthel, who then asked for the edit.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #684, opened 2026-09-18
docs(electronics): connecting the components moves up beside installing the electronics

basically-balloon:
Asked for in `#balloon-does-docs` by **barthel** (@uwe_barthel): move "Connecting the components" to the same level as "Installing the electronics".

The four pages under *Installing the electronics* each build one enclosure and bolt it to the frame. *Connecting the components* does something different: it wires the whole machine together and is the last thing before software setup. It now sits beside the installation group instead of inside it.

**What changed**

- `docs/src/content/hardware/electronics/installation/connecting.md` → `docs/src/content/hardware/electronics/connecting.md`, permalink `/hardware/electronics/connecting/`.
- `static/_redirects`: 301 from the old URL, so existing links keep working.
- `nav.yml`: the entry moves out of the *Installing the electronics* children and becomes a sibling of it.
- `installation/index.md`: the numbered list is the four enclosure builds; connecting follows as a sentence after them.
- Inbound links updated on `camera-lamp/index.md`, `lamp-arm.md` and `led-strip.md` (the "step 5 of connecting the components" references, which still point at the right steps).

No prose on the connecting page itself changed.

**Note on overlap:** #679 also touches `connecting.md`, `installation/index.md`, `nav.yml` and `_redirects`. Whichever of the two merges second will need a rebase.

Verified with `npm run check` (0 errors), `validate_images.py`, `validate_frontmatter.py`, `validate_credits.py` (no new findings), and the rendered sidebar on `npm run dev`.

PR #679, opened 2026-09-18
catalog+docs(psu box): the page follows a real build, and the pigtail page moves to helpers

basically-balloon:
## For a reviewer

Three docs pages, the installation section index and one catalog assembly, all
about the PSU box. It started as "move the pigtail page into helpers" and grew,
over twenty-five rounds with barthel in
[#balloon-does-docs](https://discord.com/channels/1430279849171222682/1536088194557419660/1550385167116279808),
into a rework of the PSU box page, which **barthel has now built from**. The
round-by-round history is folded up at the bottom; this is what the diff does.

**1. The pigtail page moves to helpers.**
`/hardware/electronics/psu-pigtail/` → `/hardware/helpers/psu-pigtail/`, with a
301 in `static/_redirects`, and `nav.yml`, the helpers index and `harness.yml`'s
`guide` link repointed. The page is now only about making the cable: start from
a panel-mount jack that already has leads, crimp a fork terminal on each, build
three. Landing and mounting them left this page for the PSU box page, and the
"not validated against a built machine" banner came off (Jon wrote the page,
built the pigtails and took its photos). Since #653 landed it carries
`parts_needed: [dc-jack-5521-panel, terminal-fork-m35]` instead of a
hand-written spec list, and the 12 mm panel hole, the one fact only this page
had, moved onto the jack's own entry.

**2. The PSU box page is rebuilt, in the order somebody actually builds it.**
barthel built the box while this PR was open and the page follows what he did:
**1 fit and wire the three jacks, 2 fit and wire the mains inlet, 3 fasten the
plates to the supply, 4 close the box**. The connections plate is populated and
wired in your hand and goes onto the supply last, which is the point of the
reorder: the old page had you landing nine terminals on a plate already fixed a
hand's width from the block. The first-draft banner is off and
`last_verified: 2026-09-18` is on, because he worked through it. Bolting to the
frame left the page for the installation overview, which already carried every
fact it stated, and the 2 M5 × 12 and 2 T-nuts left the page's parts list with
it. Three drawings replace the two `Image coming` placeholders, and there is a
`## The finished result` section holding a described placeholder.

**3. `connecting.md` loses its step 1.** It described the inside of the PSU box
and carried a second, thinner inline SVG of the same terminal block. Its two
real facts moved up into the page's own "three 24 V leads" section, steps 2 to 8
became 1 to 7, and the references on `lamp-arm.md` and `led-strip.md` moved with
them. It is also the only first draft left in the section, which
[...]

PR #685, opened 2026-09-18
docs(helpers): a page each for the two cables the builder makes

basically-balloon:
Asked for in `#balloon-does-docs` by **barthel** (@uwe_barthel): making the wire harness moves into Helpers, split one page per cable.

The PSU output pigtail (#679) and the prepared LED strip already have helper pages. `W1` and the chute stepper lead are the other two cables a builder builds, and they had no page: `W1`'s whole instruction was one `<dd>` on the connecting page, and the chute lead was a line in order spec section 4 and half a sentence in connecting step 3.

**Two new helper pages**

- **Make the control board's 24 V lead** (`/hardware/helpers/board-24v-lead/`). `W1`: moulded barrel plug spliced to 914 mm of 18 AWG, two JST-VH contacts crimped into a VHR-2. Carries the traps that were scattered across catalog notes: 2.1 mm not 2.5 mm, VH is not PH, and the crimp kit in the parts list has no VH housings even though its crimper takes the contacts.
- **Make the chute stepper lead** (`/hardware/helpers/chute-stepper-lead/`). The only stepper cable that gets built, since the four channel motors take a bought PH-to-PH cable. Coil check first, then four contacts into a 4-pin housing with each coil kept on positions 1-2 and 3-4.

**Two things checked against the board files rather than copied**

- **`J23` and `J24` are the same four nets.** Exported the schematic netlist for `1_Distribution_Board`: both connectors land on `A_1`, `A_2`, `B_1`, `B_2` at driver `P3` pins 4, 3, 5, 6. So the chute lead can end in either, and the page says so. This is the "unsettled" note on `jst-phr-4` in the catalog (harness drawing says PHR-4, netlist says Dupont header): the answer is that both are correct.
- **`J1` pin 1 is the square pad.** `J1` is a `B2P-VH-FB-B`, a shrouded header, and pad 1 is roundrect where pad 2 is circle. So the housing can only go on one way and the polarity is decided by which position the red contact goes into, which is what the page tells the builder to check.

**What is flagged rather than answered**

- Both pages carry the standard first-draft banner. Nobody has built either lead from these steps.
- The chute lead's length is a guess copied from the channel steppers, and no WireViz drawing covers that cable. Whether the motor's own flying leads are long enough to crimp directly is not recorded, so the page gives both cases.
- **The chute motor is a NEMA 23** and the harness specs its cable at 24 AWG, the same as the NEMA 17 channel motors, because that is the heaviest wire a JST-PH contact takes. Worth a current check by somebody with the motor's datasheet before this stops being a draft. Not changed here.

**Follow-up not in this PR:** `jst-vhr-2` and `jst-svh-21t` exist only for `W1` and their `docs_page` still points at the wire harness page. `catalog/generate.py` cannot run in my container (no `scipy`), so I cannot regenerate the derived data that a `parts.json` edit has to ship with.

**Rebased on `main` 2026-09-18, after #684 merged.** The conflicts were the additive ones in `helpers/index.md` and `nav.yml` (both sides kept, WiFi module page first), and `connecting.md`, which #684 moved to `/hardware/electronics/connecting/`; the edits here applied to it at its new path. The two new pages now link that page directly instead of through the redirect. Steps 2 and 3 on it are still the ones these pages point at. #679 is still open and still touches `connecting.md`, `nav.yml` and `harness.yml`, so whichever of the two merges last needs a rebase.

Re-verified after the rebase with `npm run check` (0 errors), `validate_images.py`, `validate_frontmatter.py` and `validate_credits.py` (no new findings), and both pages plus the wireviz, order and connecting pages rendered on `npm run dev`.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #692, opened 2026-09-19
docs: the WireViz drawings move to Parts and become the only copy

basically-balloon:
Moves **WireViz drawings** out of Electronics and into **Parts**, formats every drawing block the same way, and leaves it as the only page on the site carrying a WireViz drawing.

### The move

`/hardware/electronics/wireviz/` → `/hardware/parts/wireviz/`, with a 301 in `static/_redirects` so old links and bookmarks keep working. Sidebar entry moves in `nav.yml`, the page is listed on the Parts landing page, and the six pages that pointed at the old URL now point at the new one.

### The formatting

Every drawing now renders the same block, in the same order: heading, photo (where there is one), drawing, build link, downloads, BOM.

- **One label, one place for the build link.** It read *How to make your own →* under the PSU pigtail, inside the photo's caption, and *How to make it →* under the board lead, as a standalone bold line. Both are now *How to make your own →* on its own line under the drawing. The per-drawing `guide_label` is gone from `harness.yml` so it cannot drift again.
- **A stable anchor per drawing**, taken from the drawing's source name (`#power`, `#psu-pigtail`, `#board-power`, `#steppers`, `#leds`), so another page can link one drawing instead of the whole page.
- **Contents list at the top**, with the two sub-harnesses nested under the overview they come out of, and a line under each sub-harness heading naming its parent.
- **Drawing captions align with the drawing.** `.harness-figure figcaption` was falling through to the generic `figure > figcaption` rule, which centres a 440px column under a full-width drawing.

### The single source of truth

`helpers/board-24v-lead.md` was embedding the `board-power` drawing itself, which is the one real duplicate. It is removed; the page opens by naming `W1` with a link to that drawing and closes with a **Reference** section, matching what `helpers/psu-pigtail.md` already did. The order spec, layer adapter board, LED strip and electronics overview pages already linked rather than copied, so they only needed the new URL (and deep links where one drawing was meant).

### Checks

`npm run check` clean (0 errors). `npm run build` prerenders the whole site with no broken link to the old URL. `validate_frontmatter.py` and `validate_credits.py` report only the failures already on `main` (`top-interface`, `arranging-c-channels`, `camera-calibration`, and the four non-hardware pages), none on the pages touched here.

PR #693, opened 2026-09-19
docs: the adapter board's connections on the chute page, the control board's on connecting

basically-balloon:
Asked for in `#balloon-does-docs` by **barthel** (@uwe_barthel), who settled where each end of the ribbon is described: *"Plug the ribbon cable in Adapter Board in chute page is fine, because it's hard to arrive later on. The connection on the Distribution Board should be described in the connecting components page."*

So the split is by which end of the cable, and by where the builder is standing when she makes it.

**Layer adapter board page** (`assembly/distribution/chute/pcb.md`) gets the connections she makes on the bench, which is step 3 there and was already the right home:

- `J3` is the ribbon coming in and `J4` is the one going on down to the next layer. They are identical 16-pin sockets and only the silkscreen tells them apart, so the warning that a board fed on `J4` moves every layer below it wrong moves here with the instruction.
- This layer's servo on `J5`, the 3-pin header: pin 1 signal, pin 2 servo power, pin 3 ground.
- The reason barthel gave is on the page: do it now, these are hard to reach once the chute is in the frame.

**Connecting the components** keeps the control board's end: the 1 m ribbon out of `J17` down to `J3` on the first layer board. It no longer restates the layer-to-layer chaining or the servo; it says those are already done, and why. The pre-power checklist still checks `J3` in and `J4` to `J3` down the stack, which is a check rather than an instruction and is the last chance to catch it.

**Worth an eye on a real board.** `pcb.md` is builder-verified (`last_verified: 2026-09-07`) and the designators added to it are read off `electronics/KiCad/0_Adapter_Board` rather than from a build: `J3` and `J4` are the two `IDC-Header_2x08`, and `J5` is the `1x03` whose nets are `/PWM_Signal`, `+ServoV`, `GND` on pins 1, 2 and 3. No banner added, but somebody with a layer board in hand should confirm the silkscreen reads the same way round.

**Rebased onto `main` after #692 merged.** #692 moved the WireViz drawings to `/hardware/parts/wireviz/` and rewrote the sentence in `pcb.md` that step 3 here replaces; the conflict was resolved by keeping step 3's connection instructions and carrying #692's harness-routing pointer at its new path onto the end of the step.

`npm run check` clean, `validate_images.py` and `validate_frontmatter.py` report only the failures already on `main`.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #694, opened 2026-09-19
docs(wifi module): real photos of the board on the bench for first setup

basically-balloon:
Two real photos on `/hardware/helpers/wifi-module/`, which shipped in #682 with no picture of what the bench setup actually looks like.

**Photographed and supplied by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550804014273134653): "two images of orange pi, with wifi fitted, power plugged in, Ethernet plugged in for first setup. It doesn't have the orange pi official fan. Off the machine and on a desk to set wifi".** Same build as the washer report in #682.

### What they add

A new **On the bench for first setup** section after step 4. Setting WiFi means the board boots on its own before it is on the machine, and until now the page said that without showing it: USB-C in the socket marked `PWR IN`, Ethernet to a router for that first boot alone, antennas on.

Her own caveat is in the page rather than dropped: **neither photo has the Pi's heatsink fan on the chip.** That fan is fitted in `orange-pi-mount.md` step 1, so the caption says the board is running without it on a desk and that this is not a reason to leave it off. Without that line a reader can reasonably infer the fan is optional.

The step 2 `img-placeholder` stays, and so does the page `warning:`. Neither photo shows the module seated in its slot, because the slot is on the underside and the board is the right way up in both.

### Checks

Both uploaded with `docs/scripts/upload_image.py` (namespace `sorter-docs`), so `validate_images.py` verifies the bytes behind both URLs: 310 assets pass. `validate_credits.py` and `validate_frontmatter.py` both clean for this page; their remaining failures are the pre-existing ones on `main`. `npm run check` clean. `last_verified` bumped to the date of the photos.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #653, opened 2026-09-14
catalog: add every harness connector and the hook-up wire (15 parts)

basically-balloon:
Requested by BrickCycleAlice in `#balloon-electronics`, reviewing the draft [Connecting the components](https://github.com/basicallysource/sorter-v2/pull/649) page: https://discord.com/channels/1430279849171222682/1548985579860009040/1549002619383455856

> can you make a seperate pr to update the catalog, with the 13 missing items. I'll wait to see what falls out of @dov2000 build before buying anything. And leave as draft for now as well.

Closes the gap recorded in #652. Her point, which is the whole reason this exists: the docs tell a builder to crimp a VHR-2, push a #187 receptacle onto the limit switch and cut 14 ft of 22 AWG, and not one of those had a catalog entry. They appeared in no build list and on nothing a person could buy from.

**It is 15 entries, not 13.** The table in #652 has 14 rows, and I wrote "thirteen" in the issue text and in Discord; that was my miscount. The 14th row bundles the stepper-cable wire with its sleeving, which are two separate purchases, so the catalog needs 15.

## What is in it

All `kind: "cots"`, `category: "Wire harness"`, each with a product photo published to the asset service and the per-machine count totalled from the WireViz BOMs.

| id | Name | Per machine |
| --- | --- | --- |
| `dc-plug-5521-male` | DC barrel plug, 5.5 x 2.1 mm male | 7 |
| `dc-jack-5521-panel` | DC barrel jack, 5.5 x 2.1 mm female, panel-mount | 3 |
| `dc-jack-5521-inline` | DC barrel jack, 5.5 x 2.1 mm female, inline | 3 |
| `jst-vhr-2` | JST VH housing, 2-pin (VHR-2) | 1 |
| `jst-svh-21t` | JST VH crimp contact (SVH-21T-P1.1) | 2 |
| `jst-phr-4` | JST PH housing, 4-pin (PHR-4) | 5 |
| `jst-phr-6` | JST PH housing, 6-pin (PHR-6) | 4 |
| `jst-sph-002t` | JST PH crimp contact (SPH-002T-P0.5S) | 36 |
| `dupont-housing-3p` | Dupont housing, 1x3 female, 2.54 mm | 1 |
| `terminal-fork-m35` | Insulated fork terminal, M3.5, 18 AWG | 6 |
| `terminal-qc-187` | Insulated quick-connect receptacle, #187 | 2 |
[...]

PR #696, opened 2026-09-19
docs: one page for the harness order pack, drawings and spec together

basically-balloon:
Asked for in `#balloon-does-docs` by **barthel** (@uwe_barthel): *"Should we combine the WireViz and order spec pages into one page or one own sub section?"* → one page.

**Stacked on #692.** Its base is `balloon/wireviz-single-source`, not `main`, because it builds on the move that PR makes. GitHub will retarget this to `main` when #692 lands.

### The page

`/hardware/parts/wireviz/` → `/hardware/parts/harness-order/`, titled **Ordering the wire harness**. "WireViz" is the tool that drew the pictures and means nothing to a builder or a vendor. Both old URLs and `/hardware/electronics/order/` 301 straight to the new one, with no redirect chain.

Order: the warning, the zip, **Global spec**, **How to order it**, **Ends the vendor cannot terminate**, **Stepper cable pin map**, the five drawings with their BOMs, **Guesses to verify before sending**.

### What came across from the order spec, and what did not

Kept: section 1 (global spec, with section 3.1's barrel-jack detail folded into it), section 4 (ends the vendor cannot terminate), section 6 (how to order), section 7 (guesses), and the two "not part of this order" facts that were buried in a callout in section 2.

Cut:
- **Section 2, the cable build list.** The five drawings are one per buildable cable already, which is why section 3.4 was only a pointer at them.
- **Section 5, the connector BOM.** The catalog has carried every one of these since #653, with a per-machine count and a photo the table did not have. Replaced by one line pointing there.
- **Section 3.3, the chute pin map.** That cable is built rather than ordered, and its helper page has the pin order.

**Kept against what I said I would do:** section 3.2, the channel stepper crossover map and its "for the vendor, explicitly" warning. It is the one instruction in the pack a vendor can get wrong in a way that costs a whole batch, and my reason for cutting it rested on a bad measurement (below).

### Correcting a number in the thread

I told barthel the two pages rendered at about 5,100 and 3,200 words and that concatenating them would be unreadable. Those counts were wrong: my extraction fell back to counting the whole HTML document, inline SVG coordinates and all. Measured properly, the order spec is **1,521** words of article text and the drawings page **409**. The merged and trimmed page is **1,419**, against 2,235 for the connecting page. The trim is still right, but the reason is the duplication, not the length.

### Follow-up I cannot do here

`wire-24awg` in the catalog says its colours match "the pin map on the order spec page". That page is gone; the pin map is on the new one. `catalog/generate.py` needs `scipy`, which is not in my container, so a `parts.json` edit cannot ship with its regenerated data from here.

**Overlap:** #693 edits the same sentence of `pcb.md` that this updates the URL in. Whichever lands second needs a rebase.

`npm run check` clean, `validate_images.py` and `validate_frontmatter.py` report only the failures already on `main`, and every touched page renders on `npm run dev`.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #698, opened 2026-09-19
catalog: the parts that name the stepper pin map follow it to its new page

basically-balloon:
Asked for in `#balloon-does-docs` by **barthel** (@uwe_barthel): *"Create a new PR for the wire-24awg follow up"*, the dangling reference #696 leaves behind.

**Stacked on #696**, base `balloon/harness-order-pack`, because the page these parts now point at is the one that PR creates. GitHub retargets this to `main` as the stack lands.

### What changed, five lines in each file

- **`wire-24awg`** said its colours match *"the pin map on the order spec page"*. That page merged into **Ordering the wire harness** in #696. It now says *"the stepper pin map on the docs page"* and its `docs_page` points there.
- **`jst-phr-4`** had the same dangling reference in its **Pin order** attribute, *"see the stepper pin map on the order spec page"*. Same fix, same `docs_page`.

  Both use "the docs page" rather than naming it, which is how `jst-phr-4`'s own description already phrases this (*"follow the mapping on the docs page rather than guessing"*). It stays true the next time a page moves.
- **`dc-jack-5521-inline`** still had `docs_page: /hardware/electronics/led-drop/`, a page retired into the LED strip helper on 2026-09-16. It worked, through a 301. It now points straight at `/hardware/helpers/led-strip/`.

`catalog.generated.json` is regenerated and committed alongside, and its diff is the same five lines.

### One thing I did not touch

**`/hardware/helpers/external-bracket/` is a 404**, and `ext-bracket-left`, `ext-bracket-bottom-vertical` and `ext-bracket-cover` all carry it as their `docs_page`. No page has that permalink and there is no redirect for it, so those three parts link a builder to nothing. It is unrelated to the page moves and I do not know what it was meant to point at, so it is flagged rather than guessed at.

### Checks

`python catalog/generate.py --metadata-only` exits 0: *catalog.generated.json agrees with parts.json (115 pinned parts)*, *versioning discipline holds*, *connections are consistent (60 edges)*. `npm run check` in `parts-calculator` is clean (0 errors).

I told barthel twice that I could not run `generate.py` here because the container has no `scipy`. That was wrong, and my own notes said so: `pip install --break-system-packages -r catalog/requirements.txt` takes under a minute.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #700, opened 2026-09-19
catalog+docs: the external bracket parts stop linking to a retired page

basically-balloon:
Asked for in `#balloon-does-docs` by **barthel** (@uwe_barthel): *"Create a new PR to fix the 404 for external bracket"*, found while doing #698.

**Based on `main`**, not on the #692 / #696 / #698 stack. It touches `parts.json` in a different region and appends to `_redirects`, so it merges in any order.

### What was wrong

`/hardware/helpers/external-bracket/` returns 404 on the live site, and `ext-bracket-left`, `ext-bracket-bottom-vertical` and `ext-bracket-cover` all carried it as their `docs_page`. So the parts calculator has been linking builders to nothing.

The page was retired deliberately on 2026-08-31 in `fbf8ed9b`, *"superseded by the pages that install it"*, because both joints it demonstrated are covered in context on regular layers, bottom layer and top interface. Nothing wrong with retiring it; the catalog references and the redirect were just never cleaned up.

### The fix, and why it is a removal rather than a new target

**`docs_page` cannot be correct on these parts.** `ext-bracket-left` belongs to three assemblies across two different pages:

- `external-bracket` → `/hardware/assembly/distribution/bin-frame/hex-frame/`
- `external-bracket-with-support` → the same hex frame page
- `external-bracket-with-foot` → `/hardware/assembly/distribution/bin-frame/bottom-layer/`

Those assembly-level `docs` routes are already right and already live, so the per-part field has nothing true to say. `ext-bracket-foot-cover`, the newest member of the same family, already has no `docs_page`; the other three now match it. `filament.ts` describes the field as *"its docs-site detail page, when one exists"*, and one does not.

**The old URL now 301s to the hex frame**, which is where the catalog says the bracket is built, so a link made before August still lands somewhere useful.

### Checks

`python catalog/generate.py --metadata-only` exits 0: *catalog.generated.json agrees with parts.json (115 pinned parts)*, *versioning discipline holds*, *connections are consistent (60 edges)*. The generated diff is the same three parts, `docs_page` going to `null`. `npm run check` in `parts-calculator` is clean (0 errors). `_redirects` is a static file, so there is nothing to compile; the redirect is verified on this PR's preview.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #704, opened 2026-09-19
docs(sorteros): first boot is slow, and it has a progress page nobody knew about

basically-balloon:
Step 4 of the SorterOS install told people first boot "takes less than 5 minutes" and never mentioned that the Pi serves a live progress page the whole time.

**Triggered by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550826007114551457): "the light is still flashing, i plugged it in at 7:37 its now 9:08".** She spent about an hour and forty minutes watching an LED, with no way to tell a working first boot from a stuck one, and [reached the UI](https://discord.com/channels/1430279849171222682/1550454875698237472/1550828319954575360) once she found the Pi's IP on her router. Everything below is what the page should have told her.

### Three things the page had wrong or missing

**"Less than 5 minutes" is not what happens.** `software/sorteros/build/overlay/usr/local/sbin/sorteros-firstboot.py` runs 15 stages, of which `clone-repo`, `uv-sync` and `pnpm-install` fetch the software and then build the backend's Python environment and the frontend's node modules **on the board**. The image is a bootstrap, not a finished install. The page says first boot is not a matter of a minute or two and depends on the connection, and **quotes no figure**. She [corrected the assumption herself](https://discord.com/channels/1430279849171222682/1550454875698237472/1550829825516769311): *"I don't actually think it took over an hour, i just couldn't see it."* Her elapsed time measures when she found the UI, not when the boot finished, so it is not a number for a page. Nobody has timed one; the page points at the progress indicator instead of the clock.

**There is a progress page, on port 80, and no doc mentioned it.** `STATUS_PORT = 80`; the daemon renders every stage with its state and a 5 second meta refresh, and its own comment says it exists so that users pointing a browser at the device "see what's happening instead of an ERR_CONNECTION_REFUSED", then "hands port 80 over to sorter-ui-dev.service once all stages complete, same URL transitions from setup status to the UI". So the address the page already gives for the UI is also the address for watching setup, which is a free fix: no new address to explain.

**A network stage waits, it does not fail.** The daemon's docstring: "boot is NEVER blocked, errors are NEVER fatal". A stage needing internet skips and retries on the next 60s iteration. So Ethernet with a link light but no route out looks exactly like steady progress from outside, and the status page is the only thing that distinguishes them. Said outright now.

Also states that **the LEDs are not a progress indicator** (`sorter/safe-shutdown.md` has them blinking during normal running, stopping only at shutdown). I told her a blinking green meant it was still working, which is how she ended up waiting on it, so this line is there to stop the next person doing the same.

### Scope

`sorter/installation/sorter-os.md` step 4 only. The hostname and port-80 bullets from #676 are untouched; the progress-page paragraphs sit after them because you need the address first either way. `last_verified` bumped, since this is the first time step 4 has been checked against a real first boot.

No stage timings are quoted, and the earlier draft's "well over an hour" is gone for the reason above.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #706, opened 2026-09-19
docs(wifi module): the WiFi address handover, and shutting the board down

basically-balloon:
One paragraph in step 4 of `/hardware/helpers/wifi-module/`: read the machine's new WiFi address off the WiFi settings page before unplugging the Ethernet.

**Found and reported by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550830719113371679): "the wifi address number did change, but it's shown on the sorter page, beside wifi adapters, so garb that number, then unplug the ethernet".** She hit it doing exactly what the page tells you to do, on the bring-up in #682 and #694.

### Why it matters

Joining a network gets the machine a second address. The page told you to set WiFi and then take the board to the machine, and said nothing about the handover, so the obvious next move is to pull the cable, at which point the page you were reading goes with it and the machine is somewhere you have to go and find.

**The fix is free, because the address is already on screen.** `server/routers/wifi.py::_wifi_adapters` returns `"ip": _device_ip(device) if connected else None` per adapter, so the adapter row on that same WiFi page shows the new address the moment it connects. I had told her to go back to the router's device list; she found the better route on the hardware and corrected me.

Also adds half a sentence to the bench section saying the Ethernet comes out at the end of step 4, so the section states where the cable's life ends rather than leaving it implied.

### Second ask on the same branch: shutting the board down

[Asked outright](https://discord.com/channels/1430279849171222682/1550454875698237472/1550831153814970389): *"how do i turn it off again now i'm connected? press the button and wait. can that go in the page too please."* The page walked a builder up to a running board on a desk and then told them to take it to the machine, with nothing in between about powering it off, so it now has a **Shutting it down** section: never cut power to a running board, press the small black button once, shutdown takes about a minute and a half and has finished when the red and green LEDs stop blinking, then unplug. Says explicitly that the button is a shutdown button and not a power switch, since the board starts again the moment it has power.

Sourced from [Shutting down the machine](https://docs.basically.website/sorter/safe-shutdown/), which is linked as the UI route for when the board is in the machine and the button is harder to reach.

### Checks

`npm run check` clean; `validate_frontmatter.py`, `validate_credits.py` and `validate_images.py` all clean for this page. #694 merged first, so this is a fresh branch off `main` rather than a push to that one.

PR #707, opened 2026-09-19
docs: a Preparing the Orange Pi page, before the mount

basically-balloon:
The Orange Pi gets the same treatment as the control board: its own prep page in the electronics sequence, between **Control board housing** and **Orange Pi mount**.

**Requested by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550855039298572400): "The needed parts should move to the new page instead of copying. The link must be correct in parts catalog but the number of the BOM item is the same. #706 is merged. Create a PR"**, following [his proposal](https://discord.com/channels/1430279849171222682/1550454875698237472/1550853510403203133) to move the WiFi module and heatsink fan onto a new page "just as we did for the control board". [BrickCycleAlice (@brickcyclealice) asked for the same page by name](https://discord.com/channels/1430279849171222682/1550454875698237472/1550468374549499998) a day earlier, mid-build, before either of us had checked that it did not exist.

### What moved, and what deliberately did not

**Onto `/hardware/electronics/installation/orange-pi-prep/`:** the heatsink fan out of `orange-pi-mount` step 1, the WiFi module helper whole, and the bench tail that was added to it in #694 and #706 (first boot, setting the network, shutting down before the board moves). It runs bare board to ready-to-bolt-down in six steps.

**Stayed on `orange-pi-mount`:** the six M3 heat inserts. Those are prep of the *printed bracket*, not of the board, and that distinction is what keeps the new page coherent. The mount page keeps the 13 mm / 38 mm clearance line, because that is a fact about the arm, and gains a lead line naming the prep page as where the board comes from.

**Stayed on `orange-pi-5`:** the buying half. Variant, the two connector formats, the notch comparison, the driver notice. That page is read before you own anything, which is why the install steps should not have been on it in the first place. Its pointer now goes to the new page.

### Moved, not copied

- `docs/src/content/hardware/helpers/wifi-module.md` is **deleted**, not left behind, and `/hardware/helpers/wifi-module/` 301s to the new page in `static/_redirects` alongside the other moves this month. Removed from `nav.yml` and the helpers index; `orange-pi-5.md`'s link retargeted. Nothing else in `docs/` or `parts-calculator/` referenced it.
- `fan-orange-pi-5-heatsink` **leaves** `orange-pi-mount`'s `parts_needed` and appears on the new page instead. Worth noting this makes the mount page *more* accurate, not less: the `orange-pi-mount` assembly's `lines` never contained the heatsink fan, and does not contain `wifi-module-opi5` either. Both are BOM hand counts, so no assembly line, no version stamp and **no quantity changes anywhere**.
- Both parts now carry `docs_page: /hardware/electronics/installation/orange-pi-prep/`, which is the existing per-part link field. Regenerated with `catalog/generate.py --metadata-only`; the diff is those two keys.

### Page conventions

Ends in one `## The finished result` with the completed-board photo, and that same photo is the `og_image` and appears nowhere else on the page. Authored to barthel as the requester, with brickcyclealice and spencer as contributors so the bylines from the merged pages survive the move.

The caveat on that photo is kept from #694: the board in it has no heatsink fan on the chip, which is now step 1 of this very page, so the caption says so rather than letting it read as permission.

### Checks

`npm run check` clean in both `docs/` and `parts-calculator/`. `validate_images.py`, `validate_credits.py` and `validate_frontmatter.py` clean for every touched page; their remaining failures are the pre-existing ones on `main`. `generate.py --metadata-only` passes its three checks.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #709, opened 2026-09-19
docs: the Pico with headers is a prepared component, and its helper page ends in the result

basically-balloon:
Asked for in `#balloon-does-docs`: drop step 3 of the Pico headers helper, rework the page for the reader persona, give it a **The finished result** section with a photo of a Pico with its headers on, and make that photo the preview. Then take the Pico out of the control board page's parts list and have it arrive as a prepared component instead.

### `/hardware/helpers/pico-headers/`

- **Step 3 is gone.** Seating the Pico in the board is step 2 of [preparing the control board](https://docs.basically.website/hardware/electronics/installation/control-board-prep/), which the page now points at from the foot instead.
- **Step 2 keeps only the part that is a soldering decision.** The old text sent the builder off to dry-fit the bare Pico against the board's sockets to work out which face the pins go on. Which face is now stated: the plastic sits against the back, the pins point away from the chip side, you solder on the chip side. That comes off Spencer's own seating photos on the control board page, where the chip and USB face up and the pins run down into the sockets. The USB-end orientation went with step 3; it is a seating decision, not a soldering one, and the socket pinout forces it anyway.
- **Added:** a `tools_needed` list (the page had none), that a row is 20 holes down each long edge, and that the three holes in the short end are the debug pads and take nothing. Solder the two end pins first, check square, then the rest.
- **Removed:** "It is the only assembly in the parts tree recorded as soldered", which is what the catalog knows rather than what she does, and "Nothing here takes a heat insert" from the Preparation step, the shape you asked to have off the PSU box page.
- **Step 1's bare `Image coming` placeholder is gone** (the parts list above it already shows the strip). Step 2's is now a described placeholder saying what the missing photo is.

### The finished-result photo

barthel supplied it. A plain Raspberry Pi Pico with both rows soldered on, the joints visible on the chip side. Credited `Manufacturer photo (Raspberry Pi Pico with headers fitted; seller not recorded)` because the Amazon URL does not name a seller. It replaced a Raspberry Pi Pico H shot the first push used, which carried a DEBUG socket the bare Pico in the catalog does not have.

**The caption under it is now the credit alone.** The line above the photo already says what she is looking at, so everything else there was said twice.

### The Mira pass

The debug pads were in step 2, in the caption and in the alt text, three copies of an answer to a question she never asks. All three are gone. The rest:

- **An iron temperature**, which the step needed and did not have. The heat inserts page sets the precedent for giving one on this site.
- **One instruction per list item**, and "the board" no longer means the Pico in one sentence and the control board in the next.
- **Cut:** "which is the 40 the Pico takes" (arithmetic she does not do), "the pins are a separate purchase" (procurement, and the parts list above already has them), "so check it before the iron touches anything" (a sentence saying an instruction is an instruction), and the banner's clause about where the result photo came from, which the credit under the photo already carries.

**What would still stop her, and I cannot fix by writing:** step 2 has no photo of the soldering, because none exists. It is a described placeholder saying what the missing shot is. The AI-draft banner stays with it until somebody builds from the page.

- `mcu-rpi-pico` out of `parts_needed`, and a `prep-item` in above step 1 with the same photo, the shape the lamp arm uses for a prepared LED strip. The intro's "Solder the Pico headers first" clause came out with it, so the page does not say it twice.
- Step 2 still seats the Pico and is unchanged. Nothing moves in `parts.json`: `parts_needed` is the page's own list, and the `pico-headers` assembly already carries the Pico and the 40 pins, so buy quantities are untouched.

`npm run check`, `validate_images.py`, `validate_credits.py` and `validate_frontmatter.py` all pass on both pages.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

Issue #662, opened 2026-09-16
parts-calculator: delete the unused scr-m3-10-shcs catalog part


basically-balloon:
`scr-m3-10-shcs` ("M3 × 10 mm socket/button head", uid `ufvb`) in `parts-calculator/catalog/parts.json` reaches no bill of materials once [#661](https://github.com/basicallysource/sorter-v2/pull/661) lands.

Its only live use was the single screw holding the Ribbon cable clamp to the Cable cage bracket (cable mount) at the top interface, and 10 mm was too short for that joint: the clamp is drawn standing 1.30 mm off the bracket face so the ribbon is not crushed, so there is 7.94 mm from under the screw head to the bracket face and an M3 × 10 left only 2.06 mm of thread in the insert, under 1 × D. `cage-bracket-cable-clamp` v2 carries `scr-m3-12-bhcs` ("M3 × 12 mm socket/button head") instead. Two builders hit it: barthel (@uwe_barthel) on 2026-09-13 and BrickCycleAlice (@brickcyclealice) on 2026-09-16.

After that change the only remaining references are historical version snapshots of superseded assemblies, which are expected and correct to leave alone: `cage-bracket-cable-clamp` v1 (pinned by uid `ufvb`) and `cable-cage` v1. The part also had a stale `sheet_qty` of 3 per machine, which had been wrong since the cable clamp moved to `scr-m3-40-fhcs`; #661 deletes it, so it now contributes nothing to any BOM.

I can't delete the part object myself: `parts-calculator`'s rule is that a minted uid stays resolvable forever, and `check_generated_pins.py` refuses a change that drops one from `parts.json` ("uid(s) removed from parts.json ... mark them superseded/rejected instead"). That is a CI-enforced invariant, not a call to make from a Discord thread. Same situation as #453, #454, #478, #536, #563, #566, #567, #568, #618 and #636.

So this needs someone who can decide whether to override or adjust that check for this uid, then delete the part object from `catalog/parts.json` and re-run `generate.py`. Whoever does it should check that `cage-bracket-cable-clamp` v1's line snapshot still resolves, or accept that it will not.

In the meantime the part's `description` says it is retired and unused, and a P4 `changes` entry (`delete-unused-m3-10-shcs-screw`, condition `retired`) puts a "To Be Removed" marker on it wherever the calculator renders it. Both land in #661.

[...]

Issue #636, opened 2026-09-11
parts-calculator: delete the unused hsi-m3s catalog part


basically-balloon:
`hsi-m3s` ("Ruthex M3S heat insert (short)", uid `5nkl`) in `parts-calculator/catalog/parts.json` reaches no bill of materials. It was added on 2026-08-21 from the BOM sheet's Screws tab, whose row reads `M3 | Small | 4 per layer | 20 per machine | Ruthex M3x4 | Servo adapter only`, but it was never placed as a `requires` or an assembly line, because neither servo adapter half has an insert pocket:

- `servo-adapter-flap-side` (`lv7n`): four 4.0 mm bores straight through a 5.0 mm wall, 90 degree countersink on the back face.
- `servo-adapter-servo-side` (`fu2t`): four blind 2.8 mm holes 4.0 mm deep on the same bolt circle, which is a self-tap pilot diameter, not ruthex's recommended 4.0 mm insert hole.

That is a countersunk screw self-tapping into plastic, and it is what the docs have said since #416: the Door module page describes the adapter joint as 4 x M3 x 8 mm countersunk screws, "No heat inserts, the screws thread directly into the printed plastic." Neither STL has been revised since 2026-06-27. The short M3 is used nowhere else in the machine, so the ambiguity that had been tracked as the P2 `unconfirmed-servo-adapter-heat-insert` change is settled the other way: the insert is not part of the machine.

The accompanying PR retires the row (description and note say it is unused, `sheet_qty` dropped so it stops asking for 20 per machine on the Hardware tab, P4 `delete-unused-m3s-heat-insert` change entry with `condition: retired`, and the `parts_needed` reference removed from the Installing heat inserts helper page), but it cannot delete the part object: `parts-calculator`'s rule is that a minted uid stays resolvable forever, and `check_generated_pins.py` refuses a change that drops one from `parts.json` ("uid(s) removed from parts.json ... mark them superseded/rejected instead"). That is a CI-enforced invariant, not a call to make from a Discord thread. Same situation as #453, #454, #478, #536, #563, #566, #567, #568 and #618.

So this needs someone who can decide whether to override or adjust that check for this uid, then delete the part object from `catalog/parts.json` and re-run `generate.py`. **The BOM spreadsheet's own "M3 | Small" row should go with it**, and that needs someone with write access to the sheet.

[...]

Issue #567, opened 2026-09-06
parts-calculator: delete the unused led-cob-50mm-24v, with the retired light post assembly


basically-balloon:
`led-cob-50mm-24v` ("50mm LED COB - 24v", uid `s34o`) in `parts-calculator/catalog/parts.json` reaches no BOM: its only reference is `light-post-assembly`, which is not reachable from the `machine` assembly. It renders as **qty TBD** on the Hardware tab, which reads as a gap in the data rather than as a part that is out of the machine.

One per light post, lighting the feeder C-channels. The light post assembly has been marked unused since 2026-09-02, when the Camera lamp replaced it: the lamp carries its own shaded lamp over each channel, so there is no post and no COB plate on the current machine.

**The assembly goes with the part.** `light-post-assembly` still lists it, so deleting the part on its own would leave that assembly asking for something that no longer exists. Whoever picks this up should retire the two together, per `parts-calculator/VERSIONING.md` § Removing a part from the machine.

I can't delete the part object myself: a minted uid stays resolvable forever and `check_generated_pins.py` refuses a change that drops one from `parts.json`. That is a CI-enforced invariant, not a call to make from a Discord thread. Same situation as #453, #454, #536 and #565, and it should be handled with them.

Tracked in the calculator as the P4 `changes` entry `delete-unused-led-cob-50mm`.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1546026272176607232/1546035939254607942): "Tdb placeholders, can we move the three marked as unused to a p4 delete so that you're can all be handled directly."

Issue #566, opened 2026-09-06
parts-calculator: delete the unused scr-m2-8-shcs, with the retired camera extension assembly


basically-balloon:
`scr-m2-8-shcs` ("M2 x 8 mm socket/button head", uid `iwk6`) in `parts-calculator/catalog/parts.json` reaches no BOM: its only reference is `camera-extension-assembly`, which is not reachable from the `machine` assembly. It renders as **qty TBD** on the Hardware tab, which reads as a gap in the data rather than as a part that is out of the machine.

It self-taps the IMX415 classification camera board onto the camera extension's printed post, 4 per module. The camera extension assembly has been marked unused since 2026-09-02, when the Camera lamp replaced the light post, the overhead camera mount, the classification dome and the camera & LED insert. The part's description also floated it as a possible fit for the OV9732 overhead camera mount "once that gets a real bracket (#426)", which never happened and is itself part of the retired path.

**The assembly goes with the part.** `camera-extension-assembly` still lists it, so deleting the part on its own would leave that assembly asking for something that no longer exists. Whoever picks this up should retire the two together, per `parts-calculator/VERSIONING.md` § Removing a part from the machine.

I can't delete the part object myself: a minted uid stays resolvable forever and `check_generated_pins.py` refuses a change that drops one from `parts.json`. That is a CI-enforced invariant, not a call to make from a Discord thread. Same situation as #453, #454, #536 and #565, and it should be handled with them.

Tracked in the calculator as the P4 `changes` entry `delete-unused-m2-8-shcs-screw`.

Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1546026272176607232/1546035939254607942): "Tdb placeholders, can we move the three marked as unused to a p4 delete so that you're can all be handled directly."

PR #715, opened 2026-09-19
catalog+docs: both lazy Susans take M4 x 16, and the bearing is 10 mm

basically-balloon:
Both lazy Susans have been bolted with M4 × 12 countersunk since the first parts catalogue on 2026-07-11, and the screw is too short at every one of the four joints. The bearing is **10 mm thick, not the 9 mm the catalogue claimed**, which is where the error came from.

Requested by Extremetaz (@extremetaz) [from Discord](https://discord.com/channels/1430279849171222682/1550950146076835862/1550955113835659285): "at the point of attaching the lazy susan bearing to the top interface chute, the M4x12 specified is too short. This requires revision. M4x16 advised".

Confirmed by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1550950146076835862/1550957803680567398): "M4 x 12mm is definitely wrong. M4 x 16mm is the right one for the lazy susan mount.", and scoped by him [to both ends](https://discord.com/channels/1430279849171222682/1550950146076835862/1550958124981026857): "in both, top and bottom interface."

## The bearing is 10 mm

Extremetaz measured his [with a caliper](https://discord.com/channels/1430279849171222682/1550950146076835862/1550957251177488527): "The lazy susan bearing I have is 10mm thick". Checked against the sources:

- The [AliExpress listing](https://www.aliexpress.us/item/3256807110215562.html) the catalogue links **states no thickness at all**, only diameters.
- [SZSMARTER SMSP8-200](https://szsmarter.com/product-item/standard-low-profile-aluminum-lazy-susan-turntable-bearing-smsp8-200/), the manufacturer spec for this part: outside 203 mm, inside 145 mm, **material thickness 10 mm**, pitch circles 189.5 and 158 mm. That second figure is the confirmation it is the same bearing, since the printed parts are drilled to r 78.5 (top interface) and r 80.0 (bottom static), straddling r 79.0.
- [Amazon B0CDC8BFQB](https://www.amazon.com/dp/B0CDC8BFQB), same style: 8 inch, 203 × 145, **10 mm**, 450 g.
- An 8.5 mm low-profile variant is also sold, which is what the CAD appears to have been drawn around: it models the outer ring at exactly 8.500 mm seat to seat.

The 9 mm and 350 g in the catalogue were read off the Amazon measurements table in #465, on an ASIN that sells 4 inch through 20 inch from one page. **This PR corrects the thickness and leaves the weight alone**, because the only sources for a weight are vendor claims (450 and 460 g) and replacing one vendor claim with another is not an improvement. Worth a scale.

## The four joints, measured

All off the assembly-frame STLs, which for these parts are in one shared frame, so the seating is derived rather than assumed. Engagement is `L − through`, and `through` is the ring plus whatever sits under it.

**Top interface, inner ring** (step 6, 4 screws): Top interface lazy Susan washer 1.000 mm, chute mount pockets Ø5.80 × 8.200 mm blind. Inner ring at most 9.0 mm. M4 × 12 reaches **2.0 to 3.0 mm** of an 8.1 mm insert; M4 × 16 reaches 6.0 to 7.0 mm and stops 1.2 to 2.2 mm clear of the floor.

**Top interface, outer ring** (step 7, 4 screws): Interface big spacer 1.000 mm, Interface upper fixed section pockets Ø5.80 × 8.200 mm. M4 × 12 reaches **1.0 mm**; M4 × 16 reaches 5.0 mm, 3.2 mm clear.

**Bottom lazy Susan** (8 screws): 9.500 mm pockets both sides. M4 × 12 reaches **1.0 mm** on the rotating side and 3.0 mm on the static side; M4 × 16 reaches 5.0 and 7.0.

The 1 mm printed washers are not load spreaders, incidentally. They shim the inner ring 1.000 mm clear of the outer ring so the bearing does not bind, and `ls-bottom-static` does the same job with a machined step instead.

## The symptom was already written down

The original catalogue note on the M4 × 12, from 2026-07-11, reads: "These must be very tight — most electric screwdrivers won't tighten them enough. They're a hassle to reach if they come loose, and machine vibration can work them loose." That is what a screw with 1 mm of thread does, and the [bottom interface page](https://docs.basically.website/hardware/assembly/distribution/bin-frame/bottom-interface/) still carries it forward as a warning telling builders to try harder. The warnings are left in place here; they are good advice for a bearing joint regardless.

## What changed

- **Mints `scr-m4-16-cs`** and **retires `scr-m4-12-cs`**, which was used in exactly two assemblies and both were this bearing. The row is kept as an unused entry with the measurements in its `note`, since a minted uid has to stay resolvable. The catalogue ends with the same two M4 lengths it started with, which was Extremetaz's [ask](https://discord.com/channels/1430279849171222682/1550950146076835862/1550955113835659285): "Don't add new lengths to the catalogue."
- **`interface` v5 → v6** and **`lazy-susan` v2 → v3**, 8 screws each. `breaking: false`; an interface built before this is still an interface. Both new entries left with `"commit": null` for the squash.
- **`brg-lazy-susan`** thickness 9 mm → 10 mm.
- Seven fastener calls and two `parts_needed` ids across the two assembly pages; one sentence on the Lazy Susan page.
- A **P4 `retired` `changes` entry** (`delete-unused-m4-12-cs-screw`) marking the M4 x 12 for deletion, which puts the notice on it wherever the calculator draws it. Tracked as #716, labelled `bom`. Asked for by barthel (@uwe_barthel) [from Discord](https://discord.com/channels/1430279849171222682/1550950146076835862/1550961161338359889): "Create a BOM issue too for mark to be removed and delete of M4 x 12mm screw in part catalog." The part object itself cannot be removed from here: `check_generated_pins.py` refuses a change that drops a minted uid, which is why the deletion is an issue rather than a commit. Same shape as #453, #454, #478, #536, #563, #566, #567, #618, #636 and #662.

`generate.py --metadata-only` ran clean and its three self-checks passed (pins, versioning, connections). `npm run check` clean on both sites, `validate_images.py` clean.

## Not in here

The M3 × 12 holding the chute gear to the same mount, which is what [started the thread](https://discord.com/channels/1430279849171222682/1525928562971115601/1550950146076835862). That one has 2.5 mm of a 5.70 mm insert and the fix Extremetaz and I [agreed on](https://discord.com/channels/1430279849171222682/1550950146076835862/1550954931765256214) is to deepen the pocket in CAD so the existing `scr-m3-16-cs` fits, which is not a catalogue change and is not mine to make.


basically-balloon:

basically-balloon:

basically-balloon:
**Extremetaz** approved this in Discord.

That is 2 of 2 approvals.

basically-balloon:
**Merged.** Approved in Discord by barthel, Extremetaz, and it only changes files this repo allows to merge that way, so the board merged it.

PR #717, opened 2026-09-19
docs(electronics): rebuild the installation overview on the section-landing pattern

basically-balloon:
barthel asked for the installation overview to be formatted and rebuilt on the [Camera lamps](https://docs.basically.website/hardware/assembly/feeder/camera-lamp/) landing page as the example, then for the finished result to be the **bolted** enclosures rather than the sub pages' bench results.

**The shape now**

Intro, the sub pages as an `ol.numbered-steps` right under it, the bolting section in the middle, `## The finished result` as a row of the three enclosures on the frame, and a closing pointer.

**What changed**

- **The page list is an `ol.numbered-steps`** and moved to the top. It was a plain markdown list halfway down the page, under the fastener detail, and this site renders those with no numbers at all.
- **`## The finished result` is the three enclosures bolted on.** `assembly-control-board-housing-on-the-extrusion` moved out of the bolting section into it and is the page's `og_image`, so the result image and the preview image are the same picture. The bolting section keeps the top-down layout render and gains a link down to the row.
- **Preparing the Orange Pi is in the list.** That page has been in `nav.yml` since #707 and the overview's "build in this order" list never got it, so the list said four pages while the sidebar said five. It is item 4, before the mount, which is where `orange-pi-mount.md`'s prep block puts it.
- **Two sentences that went stale as pages moved**: "With all four bolted to the frame" (three enclosures bolt on; there are five pages) and "None of the three pages ends with a bolting step of its own" (five).
- The lede says what the section is rather than repeating the title.

**The two missing photos**

There is no photo anywhere of a bolted PSU box or a bolted Orange Pi mount. The control board housing on the extrusion is the only picture of any electronics enclosure on the frame in the whole corpus: every asset referenced by `docs/src` or `catalog/parts.json`, every image posted in the server since 2026-05-01, Drive and the video attachments were all searched on 2026-09-19, and re-checked against everything posted since. Both slots are described `img-placeholder`s, and the open-items section says which one exists and which two do not.

Asked in the channel for the shots. Anyone with the enclosures on a frame can take them: two clamp bosses down on the extrusion, same framing as the housing photo.

**Brought up to date with main (2026-09-20)**

#719 landed first and gave Electronics its own landing page, moving the wire harness onto `/hardware/electronics/wire-harness/`. `main` was merged in rather than rebased, so the approvals and the review history on this branch stay put. One conflicted file, `installation/index.md`, and it conflicted because this PR rewrites the whole page.

[...]

PR #719, opened 2026-09-19
docs(electronics): give Electronics a landing page and move the wire harness onto its own

basically-balloon:
barthel, after comparing `/hardware/electronics/` with `/hardware/electronics/installation/`: *"The website [electronics] looks very similar to [installation]."* Then *"Try"* on the fix below.

**The cause**

Electronics had no section landing page. The sidebar item `Electronics` opened `/hardware/electronics/`, whose own `<h1>` is **Wire harness**, a 325-line wiring reference, with `Installing the electronics` sitting directly under it. Every other section in that sidebar (Camera lamps, Bin frame, Chute, Feeder) opens on a short `type: landing` page. So you click a section and land on a long reference that opens by naming the same three components as the page below it.

Three things were genuinely duplicated, measured rather than eyeballed:

1. **One shared image, and only one.** `electronics-component-layout-topdown` was section 2 of the wire harness page and the bolting figure on the installation overview.
2. **Section 4.1, PSU assembly**, restated `psu-box.md`: the inlet's three leads to screws 1, 2 and 3 and the three output jacks on 7/4, 8/5, 9/6 with fork terminals.
3. **Section 5, Parts**, repeats the installation pages' own required components (the 40 mm fan, the buck, the USB hub). Left alone: a reference page listing what the section buys is defensible, and cutting it would take the fan and the buck out of the only place they are specced.

**What this does**

- **`/hardware/electronics/` is a landing page** with three entries: Wire harness, Installing the electronics, Connecting the components. It also routes to where the cables are bought (`/hardware/parts/harness-order/`) and to the four made by hand under Helpers, which nothing signposted before.
- **The harness reference moves to `/hardware/electronics/wire-harness/`** and keeps its content. No redirect needed: the old URL still resolves, it just belongs to the landing now.
- **Section 2 is gone** and the page links to the installation overview for the render instead. Sections renumber 3 to 7 into 2 to 6, and the two in-page references to "section 7" and the `#7--open-items` anchor follow.
- **Section 4.1 shrinks to a pointer** at the PSU box page and the pigtail helper, keeping the facts that are only here: 18 AWG, no AC cable to make, and the enclosure CAD link. It also said "6 DC output jacks" where the box takes three.
- **11 links repointed** to the new URL, every one whose link text says "wire harness". The 6 that say "Electronics" or "the electronics page" stay where they are and now land on a page that routes them, which is an improvement on landing in a spec document.
- **12 harness parts in the catalog** (the connectors, hook-up wire, sleeving, crimp kit) had `docs_page: /hardware/electronics/` and now point at the harness page. Regenerated with `generate.py --metadata-only`; the generated diff is those 12 strings and nothing else.

**Two things for you to decide**

- **No `## The finished result` on the landing.** Camera lamps and Bin frame have one because their children are assemblies; this one owns no build step, and a row of its children would repeat the installation overview's row directly below it. Say if you want one anyway.
- **`og_image` is the control board's catalog photo.** The layout render would have been the obvious pick, but it is the installation overview's preview already and I did not want the same thumbnail on both.

**Overlaps with #717.** Both branches touch `docs/src/content/hardware/electronics/installation/index.md`, this one only to repoint two links. Whichever merges second needs that hunk resolved; take the other side for everything except the two `/wire-harness/` URLs.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #724, opened 2026-09-20
docs: preparing LEGO for a sort run

basically-balloon:
**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1551039727053049999/1551132487478943845): "if there is nothing in the docs, then we will need a page for Mira to view and understand why items are being pulled".** The question that started it is [hers too](https://discord.com/channels/1430279849171222682/1536088194557419660/1551039727053049999): "is there anything in the docs, stating how to prep the lego prior to dumping it in the bulk bucket".

Nothing did. There is no intake or preparation page anywhere on the site, and no page says what the machine cannot take. This adds `/sorter/preparing-lego/` to Sorter → Operate, between setting the machine up and feeding it.

## Where each rule on the page comes from

- **80 mm, written as "ten studs".** [Marc, 2026-04-10](https://discord.com/channels/1430279849171222682/1460692585881534579/1492219678339371132): "design goal is to support a 80mm 'ball'". [flewber states the same limit as an 8 cm sphere](https://discord.com/channels/1430279849171222682/1493667046381326508/1493681494491791481) and [reports clogs from going over it](https://discord.com/channels/1430279849171222682/1460692585881534579/1488270702015549501). A LEGO stud is 8 mm, so ten studs is that number in a form an operator can count without a caliper.
- **Tyres and rubber.** The best-evidenced stoppage in the server, from four people independently: [cyborg_ar, "the first tire i threw at it stalled it"](https://discord.com/channels/1430279849171222682/1460692585881534579/1506289169247502347), [Marc, they do not drop out without wiggling](https://discord.com/channels/1430279849171222682/1460692585881534579/1506291102913659021), [Spencer, "I dont think there's a rotor steep enough for tires to fall on their own"](https://discord.com/channels/1430279849171222682/1460692585881534579/1506375004944732322), and chriskiepfer0591 [testing a batch of wheels and tyres](https://discord.com/channels/1430279849171222682/1460692585881534579/1521537188272279562) and [jamming the classification rotor with the bigger ones](https://discord.com/channels/1430279849171222682/1460692585881534579/1521537750904471744).
- **Cloth, string and tangles.** [Spencer, 2026-03-18](https://discord.com/channels/1430279849171222682/1460692585881534579/1483958747276640438): the vision system "would have quite a hard time detecting a clump of entangled pieces, they instead appear like one clump". Non-rigid elements have never been addressed as a category at all.
- **Pieces still joined together.** Same argument, plus [Marc's secondary model classifying single / multiple / "compound (for example wheel with rim)"](https://discord.com/channels/1430279849171222682/1440488123279736912/1497300149406400552), which is the tell that a joined pair reads as its own object.
- **Minifigures are pulled for value, not because they jam.** Recognition is not the problem: [Brickognize's author, in this server](https://discord.com/channels/1430279849171222682/1437436256652103731/1438212298929078302), says the model "supports minifigs, sets, and sticker sheets", and nothing in the code filters minifig or printed parts. What is unknown is how a whole figure travels through the channels, which nobody has reported. So the page pulls whole figures and soft accessories and explicitly leaves loose heads, torsos, legs, hats and hair in.
- **Duplo and Primo are a hard exclusion.** `FILTER_CATEGORIES = ["primo", "duplo"]` in `software/sorter/backend/classification/brickognize.py` drops any identification in those two categories, so such a piece can never be named.
- **Hand preparation is the assumed workflow, not a gap.** [d_bass asked on 2026-09-15](https://discord.com/channels/1430279849171222682/1430279852015091804/1549230927362916433) whether oversized pieces still have to come out of the bulk by hand; the answer in the thread was yes, and he runs a LEGO library that already does it.
- **Small parts are deliberately NOT on the take-out list.** [MattNZ argued for a minimum size in 2025](https://discord.com/channels/1430279849171222682/1430279852015091804/1437203427045150750) and was pushed back on, and no decision ever came. A small flat piece often cannot be named because the classification channel photographs from above only, but it goes to the misc bin rather than stopping anything, so pre-picking them is wasted work. The page says so.

## Not a specification, and the page says so

The project has never published a list of parts the machine will and will not take. The nearest thing on record is [flewber's definition of "excluded"](https://discord.com/channels/1430279849171222682/1430279852015091804/1466226254964592660): "if you put them through the machine, we do not guarantee any specific outcome". So the page carries a `warning:` calling itself a working house standard assembled from reported failures, to be corrected against what a real machine stops on. If anyone wants a rule on it changed, that is a better outcome than the page not existing.

## The sheet

The one-page version for a wall is drawn as a visual standard (numbered zones, icons, two words per rule) and lives on the asset service, embedded on the page with a full-size link for printing at A4. Nothing binary is committed.

**The 2x10 brick on it is drawn life size** ([asked for here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551134488065744957)), so the sheet is its own gauge: 79.8 mm long, 9.6 mm tall, studs on an 8 mm pitch, measured back off the rendered page at 79.79 mm. The 80 mm dimension bar under it is how an operator checks the printer did not scale the page.

**Tile 3's icon is a flag on a pole** after Alice rejected the first one ([here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551134726863982644)). Four candidates were drawn at print size (tangled string, cape, flag, chain links); the flag has the clearest silhouette at arm's length, and the tile's own text carries string, rope, chain, paper and card. **Superseded**: Alice asked for coiled string with a bit of chain instead ([here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551139637727199273)), so the icon is now a loose coil of string running into two chain links. Drawing the links as thin ellipses made them read as more string; they are rounded rectangles with a white fill so the overlap occludes, which is what makes a chain read as a chain. **Final arrangement** ([here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551144869492297779)): the tangle sits across the top and three links run from bottom left up to middle right, so the two things read as two things.

**A second sheet of cut-out box labels** was asked for after that ([here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551135446451495006)): the same seven icons plus the ask-someone one, eight to an A4 page with dashed cut lines, about 96 by 63 mm each, carrying the same numbers as the wall sheet so a piece matching tile 3 goes in box 3. It is a second figure on the page, hosted the same way.

**Each rule heading on the page now carries its icon** ([asked for here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551136524215779339)), so a reader working down the page gets the same visual cue as the sheet and the box in front of them. They are inline SVG rather than images: no extra asset round trip, they take the text colour through `currentColor`, and `rehype-slug` still generates the same heading ids (`#tyres-wheels-and-rubber` and the rest are unchanged, checked in the DOM).

**Tile 6 is "Not LEGO" rather than "Not our LEGO", and the docs section is "Rubbish, or Primo, Duplo and other brands"** ([asked for here](https://discord.com/channels/1430279849171222682/1551039727053049999/1551140086194896897)), with the rule's first line reordered to lead with rubbish so it follows the heading. Nothing in `src/content` linked the old `#duplo-primo-and-other-brands` anchor, checked before renaming.

Also in this PR: the nav entry under Sorter → Operate, and a line on the Sorter landing page.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:
**Merged.** Approved in Discord by BrickCycleAlice, ReveryX, and it only changes files this repo allows to merge that way, so the board merged it.

PR #728, opened 2026-09-20
docs(hex-frame): use M5 × 12 with the T-nut option in step 1

basically-balloon:
Step 1 of the hex frame offers two ways to fasten piece A into the External bracket — side, and the T-nut option silently inherits the M5 × 16 from the standard route, which is too long for it.

Extremetaz (@extremetaz) hit it on a build and [reported it](https://discord.com/channels/1430279849171222682/1525928562971115601/1551219137659797555): *"having the screw bottom out in the extrusion is not ideal as it can do so without clamping the bracket (though it does pin it into place)."* barthel (@uwe_barthel) [asked for the note](https://discord.com/channels/1430279849171222682/1551219137659797555/1551220675052576900): *"There should a note on the side, that the screw should be shorter (4mm) for use with the t-nut."*

## The measurement

Off the pinned `ext-bracket-left` STL (`da4b3abd3db7`), the two adjacent holes at each extrusion end:

| | diameter | bore | wall crossed |
|---|---|---|---|
| outer (self-tap) | ⌀4.40 mm | 7.50 mm | 8.50 mm |
| inner (clearance) | ⌀5.50 mm | 8.50 mm | 8.50 mm |

Both are perpendicular to the extrusion face, 10.0 mm apart along the extrusion, and both sit dead centre of the bracket's 20 mm seat face, so both point into the middle of the slot. The seat face is flat and uncounterbored out to at least r 6, so the grip is the full **8.50 mm** for either head style.

- **M5 × 16 into a T-nut**: 7.5 mm standing past the plastic. A 2020 roll-in nut is about 4.3 mm thick in a slot about 5.9 mm deep, so the screw runs out of slot roughly 1.6 mm before the head reaches the bracket. Pinned, no preload, which is the report.
- **M5 × 12 into a T-nut**: 3.5 mm past the plastic, thread starting about 1.6 mm down, so a little under 2 mm of engagement and the head lands. Light but real, and 12 is the next length down in the catalogue (12, 16, 20), so it is 12 or nothing. Same as barthel's 4 mm.
- **M5 × 16 in the outer hole is right and is untouched.** Nothing is clamped there: the screw threads into the plastic and the length past the wall is what drives the tip onto the slot floor, which is the bracing the step already describes.

Same shape as the bin retainers in #544, from the other end: there M5 × 12 stopped level with the extrusion face and needed to be 16.

## The change

One neutral `callout` under the **Option:** paragraph in `hex-frame.md`, saying to use M5 × 12 for that route and that those 12 are not in `parts_needed`, the same way the paragraph above already handles the 12 extra T-nuts.

It carried a second sentence on what the 16 does instead. barthel [asked for it out](https://discord.com/channels/1430279849171222682/1551219137659797555/1551224389603954810): *"Remove … and add a Note to the sentence before, that the M5x12mm are Not in the needed parts, Like the t-nut Note."* The reasoning lives in this body instead, per the house style.

`npm run check` clean, `validate_images.py` clean. No catalogue change: the option's fasteners are described in prose on the page rather than in `parts_needed`, which is how the 12 extra T-nuts are already handled.

## Not in here

The page intro says *"This frame's fastener total is fixed: 12 M5x16 screws … and the frame needs no T-nuts of its own."* That was written for the standard route and already glossed over the option before this change. Worth a separate look, left alone here.


basically-balloon:

basically-balloon:

basically-balloon:
**Extremetaz** approved this in Discord.

That is 2 of 2 approvals.

basically-balloon:
**Merged.** Approved in Discord by barthel, Extremetaz, and it only changes files this repo allows to merge that way, so the board merged it.

PR #731, opened 2026-09-21
docs(pico-headers): fit both rows before soldering, and a wood-and-vice support

basically-balloon:
Asked for by **BrickCycleAlice**, who soldered a Pico from this page and found both of its instructions wrong for her: [both sides need to be inserted before soldering to ensure the panel sits flat](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456428525486090), and [we didn't have a breadboard so made slots in a soft piece of wood to support and mounted in a vice](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456514538217476). Eleven photos came with it, and a set of [soldering tips](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456762643619871) after them.

### `/hardware/helpers/pico-headers/`

The old step 2 was one list that did a row at a time: stand a row in a breadboard, lower the Pico onto it, solder it, "do the other long edge the same way". That is the sequence she is correcting. It is now three steps.

- **Step 2, fit both rows before you solder anything.** With the reason, because the reason is the whole instruction: solder one row on its own and the second has to follow whatever angle the first one set, and the Pico will not sit flat in the control board. The which-face-do-the-pins-go-on callout is unchanged and still sits above it.
- **Step 3, stand the pins in something that holds them square.** A breadboard if you have one, otherwise two saw cuts in a piece of soft wood held in a vice, which is what she did. The cuts are dimensioned at **17.8 mm (0.7 in)**, the spacing of the Pico's two pin rows, so the next builder can cut the jig without a Pico in hand. That figure is the Pico datasheet's mechanical spec (section 2, figure 3), not a measurement off her photo.
- **Step 4, solder the four corners, then the rest.** Same as before but four joints rather than two, since both rows are in from the start, and it says to press a proud corner down after reheating it. Her three tips go in above it: as little flux as the joint needs, hot enough that the solder runs into the hole but not so hot the header plastic softens, and parts clean and straight from the packet because dust and finger oil make it harder.
- **Seven photos replace one placeholder and one stock photo.** Both rows dry-fitted, the block in the vice with the pins in the cuts, the first corner joint being made, and the finished board (`IMG_4758`, the angled shot on white, which is the same framing as the manufacturer photo it replaces). `step.html` count on this page goes from 3 to 4.

### The banner and the stamp

`warning:` is gone and `last_verified: 2026-09-21` is set, which is the pair #557 established for a page a builder has gone through: drop the banner, stamp the day it was approved. She has now built the thing this page describes and reported back on it, and the steps say what she actually did. She is added to `contributors:`.

This is the last of the three unverified steps on Soldering Pico headers from the 2026-09-17 count, so that page leaves the flagged list.

- The `prep-item` photo swaps from the manufacturer shot to hers, so the prepared component looks the same on both pages and the page no longer shows a board nobody here soldered. **Easy to revert if barthel wants his own photo back**, it is one line and both images are hosted.

Photos uploaded to `sorter-docs` as `pico-headers-rows-fitted`, `pico-headers-wood-jig-vice`, `pico-headers-soldering-first-pin` and `pico-headers-soldered-build-angled`, credited `Photo: BrickCycleAlice` in each caption.

### Step 2's photo was the wrong one, fixed

She checked the photo list and caught it: **IMG_4746, captioned "Both rows in, nothing soldered yet", has one row in and the far edge still bare holes.** Zoomed on the original, the far long edge is empty plated through-holes. It was the single photo the whole correction rests on, and it showed the thing the step tells you not to do.

Step 2 now uses **IMG_4747**, the side-on shot with a row in each long edge, both plastics flat against the board and the pads still bare gold with no solder on them. Uploaded as `pico-headers-both-rows-unsoldered`. IMG_4746 comes off the page; its asset stays hosted.

Photo-to-source mapping for the page as it now stands: step 1 IMG_4745, step 2 IMG_4747, step 3 IMG_4749 and IMG_4752, step 4 IMG_4751 and IMG_4756, finished result IMG_4758.

Step 4's caption said "The first corner joint" and she corrected that too: the iron is on the **second** corner in IMG_4751, with the pin at the far end of the same row already soldered. Caption and alt text now say so.

### Three drafts went up for this, this is the one

#732 and #733 were opened for the same four messages within twenty seconds of this PR. Both are closed. Everything they had that this did not is folded in here and credited in the commits that did it: the broken rows of 20 in step 1, the look-along-the-board check, the side-on photo of a finished joint, "desoldering forty pins", the jig as a pair of figures, and the **Getting clean joints** callout.

One thing differed between them rather than being additive. #733 had the Pico standing on a strip of wood *narrower* than the gap between the rows, with the pins hanging either side. That is not what she did: the slots message and her end-on photo both show two saw cuts with a row dropped into each. This PR says slots.

`npm run check`, `validate_images.py`, `validate_credits.py` and `validate_frontmatter.py` are all clean on both pages.

### This is the only open draft for these four messages

Three went up within twenty seconds of each other. #732 and #733 are closed against this one, and everything either of them had that this branch did not is on it. Review here; there is nothing to read in the other two.

Checked on the preview: `npm run check` is clean, `validate_images.py` passes all 329 immutable assets, and the rendered page shows the four steps, the tips callout, both jig photos and the finished shot in place. The `validate_credits.py` and `validate_frontmatter.py` failures in the tree are all on other pages and predate this branch.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #617, opened 2026-09-10
docs: arranging C-channels stands on the new support structure, not legs

basically-balloon:
**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1547408655928922293): "rewrite the arranging the c channels, with the new channel support structures instead of legs".**

The page still stood the feeder on the `Leg` / `Foot` / `Leg extension` stack, which was retired on 2026-09-02 by [#534](https://github.com/basicallysource/sorter-v2/pull/534) and is marked "Unused since 2026-09-02 ... do not re-add from old docs or photos" in the catalog. Its `parts_needed` asked for 9 of each of them.

**What the page says now**

- C1, C2 and C3 each stand on a **Layout guide**, three **support legs** and three **Support dovetail adapters** — `channel-one-support-structure`, `channel-two-support-structure`, `channel-three-support-structure` in `parts-calculator/catalog/parts.json`. `parts_needed` is those parts, at the catalog's quantities.
- **The drop between channels is 80 mm and it is built into the legs**, which step 3 previously said was "not written down anywhere today". C1's leg is `ext-2020-c1` at 228 mm (piece J), C2's is 148 mm and C3's is 68 mm, so with all three layout guides on one surface the seats land at 240 / 160 / 80 mm. Leg lengths measured off the published STLs; C1's is from the cut list.
- **No fasteners anywhere in the stand.** Every joint is `method: gravity` or friction in the catalog, and the Ø4.20 holes in the legs and Ø5.50 in the adapters take nothing.
- **Step 6 (the classification channel)** stays starred. It has no support structure in the tree, the top plate has no mount holes for a C-channel, and it has been asked in the server twice with no answer. The page gives the geometric argument for resting it on the top plate and labels it an inference.
- Step 5 said "light post and overhead camera mount", both retired by the same #534 and replaced by the camera lamp. Fixed to match [#604](https://github.com/basicallysource/sorter-v2/pull/604) and the feeder index.

[...]

PR #732, opened 2026-09-21
docs(helpers): soldering Pico headers from BrickCycleAlice's build

basically-balloon:
The [Soldering Pico headers](https://docs.basically.website/hardware/helpers/pico-headers/) helper page now follows a real build instead of the assembly tree. Seven photos from that build go on it, the step order changes to the one that keeps the Pico flat, and the AI first-draft banner comes off.

**Requested by BrickCycleAlice (@brickcyclealice) [from Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456428525486090):** *"when soldering the pico's both sides need to be inserted before soldering to ensure the panel sits flat. Here is some photos, do your thing."*

That is a correction to what the page said. It told the builder to stand one row in a breadboard, lower the Pico onto it, solder that row, and only then do the other edge, which leaves the board resting on one long edge while the first row sets. Step 2 is now "fit both rows before you solder anything", with the reason in one line.

**[From Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456514538217476):** *"we didn't have a breadboard so made slots in a soft piece of wood to support and mounted in a vice."*

New step 3. The breadboard stays as the first option, the block of wood with two saw slots, held in a vice, is the second, with her two photos of it.

**[From Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456762643619871):** *"tips for soldering, don't use too much flux, get it warm enough to run, but not so hot you melt the pico plastic. Ideally all parts should be clean and straight from the packet, dust and oil makes it harder."*

A callout under the soldering step: parts clean out of the packet, flux-cored solder carries enough on its own, heat the pin and pad until the solder runs in by itself and then take the iron off. The iron temperature already on the page (350C, 320C leaded) is unchanged.

**[From Discord](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456907519074394):** the finished Pico, which becomes the `## The finished result` photo and the page preview image.

### The banner and the stamp

`warning:` said the page was written from the [parts calculator](https://parts-calculator.basically.website/assembly?focus=pico-headers) and that the soldering was not photographed on a build. Both halves are now false, so it goes entirely rather than being softened, and `last_verified: 2026-09-21` goes on. Same shape as the funnel page (#557) and `arranging-c-channels.md` (#617). BrickCycleAlice joins `contributors:`.

### Photos

All seven are hers, uploaded to `sorter-docs`: the two broken-off rows, both rows fitted dry, the wooden jig, the jig in the vice, soldering in progress, the 40 finished joints on the chip side, and the finished board.

### The control board page's thumbnail

The manufacturer photo that was the result image (#709) was also the `prep-item` thumbnail for the prepared Pico on [Preparing the control board](https://docs.basically.website/hardware/electronics/installation/control-board-prep/). It is now her build photo, so the two pages show the same Pico, and she is added to that page's `contributors:`. Nothing else on it changes.

This is the only other page touched. It duplicates the thumbnail swap in #733, which was a second draft opened for the same request and is closed in favour of this one.


basically-balloon:
Duplicate. Three drafts went up for this request within twenty seconds of each other (#731, this one, #733). **#731 is the earliest at 05:19:31**, so that is the one to keep, and #733 is already closed against the same rule.

Everything on this branch that #731 did not have is now on it:

- the slots and the vice as two figures in an `img-row` rather than one photo
- the **Getting clean joints** callout, which reads better than the paragraph #731 had
- the broken rows of 20 next to the bare Pico in step 1

#731 additionally carries the slot spacing (17.8 mm / 0.7 in, the Pico's row spacing from the datasheet's mechanical section, so the jig can be cut before you own a Pico), the four-corners-first order, the side-on photo of what a finished joint should look like, and "desoldering forty pins" in the orientation callout, which stopped being twenty once both rows go in first.

Closing this one. Review #731.

basically-balloon:
Closing this one in favour of #731. Three drafts got opened for the same four messages within a minute of each other, which is my fault, not anyone's to untangle by reading all three.

#731 is the one to review. Where the two differed, it is the more accurate of the pair:

- The warning callout says undoing the orientation means desoldering **forty** pins, not twenty. Both rows go in before any solder now, so twenty is no longer the number.
- Step 4 tacks **four corners**, one at each end of both rows, then checks the board and finishes the remaining 36. This version still soldered one row through and then the other, which is the sequence BrickCycleAlice was correcting.
- The saw slots are dimensioned at **17.8 mm (0.7 in)**, the Pico's row spacing from the datasheet, rather than "a Pico's width apart" (the board is 21 mm wide, so that would cut them too far apart).

Everything from here that #731 did not already have is on it, including the tips callout and the "all 40 joints" check. The `control-board-prep.md` thumbnail swap is on #731 as well, so nothing here is lost.

PR #733, opened 2026-09-21
docs(helpers): both rows go into the Pico before any solder, photographed off a build

basically-balloon:
Asked for by [BrickCycleAlice in `#balloon-does-docs`](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456428525486090), with ten photos off her build: *"when soldering the pico's both sides need to be inserted before soldering to ensure the panel sits flat."*

### The correction

The page had her solder one row of 20 end to end, then do the other long edge. Nothing holds the second row in the same plane as the first while that is happening, so the Pico can come out sitting on a twist and will not go flat into the control board's sockets. **Both rows go in dry first**, held by friction, and then the four corners get tacked. The two rows hold each other square while she works.

Step 2 is retitled **Fit both rows, then solder**, and carries a callout saying why. The numbered list is now: push a row into each long edge, stand the Pico on something narrower than the gap between the rows so the pins hang clear, tack the four corners, look along the board and reheat a corner if a row leans or a plastic body has lifted, then solder the remaining 36.

The old list stood one row in a breadboard and lowered the Pico onto it, which is the one-row-at-a-time method this is replacing.

### Photos, which the page had none of

Her build is the first time this job has been photographed, so the placeholder and the stock shot both go:

- Step 1: the two rows of 20 broken off the strip, next to the bare Pico.
- Step 2: both rows pushed in with nothing soldered; the Pico standing on a strip of wood narrower than the two rows, so both sets of pins hang clear and the board sits flat; the iron on a corner pin; and a close side view of the finished joints with the plastic flat against the board, which is what she checks for.
- **The finished result** is now a Pico off a real build rather than the manufacturer photo, and it is the `og_image`.

All credited `Photo: BrickCycleAlice.` and uploaded to `sorter-docs` on the asset service.

### The draft banner comes off

Both steps on this page now come off Alice's build, and the banner's own sentence ("The soldering is not photographed on a build") stopped being true. So `warning:` goes entirely rather than getting softened, and `last_verified: 2026-09-21` goes on, which is the shape the PSU box and funnel pages use. **If that stamp is premature, say so and I will put the banner back** — she reported a fix rather than signing the page off in those words, and that reading is yours to make, not mine.

`tools_needed` gains the thing to stand the Pico on, which the old page needed and did not name.

The `prep-item` thumbnail for the prepared Pico was the same manufacturer photo, so it now uses the same build photo and the pages stay in step. Alice added to `contributors:`. Nothing else on that page changes.

`npm run check` passes, and `validate_credits.py`, `validate_frontmatter.py` and the immutable-URL check in `validate_images.py` are clean on both pages (the pre-existing failures elsewhere in the tree are untouched).


basically-balloon:
Closing as a duplicate. Two drafts got opened for the same request a few minutes apart; #732 is the one to review. It carries the same correction and the same photo set, plus BrickCycleAlice's two follow-up messages that this one predates: the jig is two saw slots cut in a block of soft wood held in a vice (not a strip narrower than the rows), and her soldering tips are a callout on the soldering step. The `control-board-prep.md` thumbnail swap from here is in #732 as well.

basically-balloon:
Duplicate of #731, which was opened 20 seconds earlier for the same four messages. Everything here that #731 did not already have is now on it: the two broken rows next to the bare Pico in step 1, the look-along-the-board check, the side-on photo of a finished joint, and "desoldering forty pins".

The one thing that changed in the merge is the support. This branch has the Pico standing on a strip of wood *narrower* than the gap between the rows, with the pins hanging either side. What she actually did is in [her second message](https://discord.com/channels/1430279849171222682/1536088194557419660/1551456514538217476): slots cut in a soft piece of wood, pins dropped into the slots, wood held in a vice. Her end-on photo shows the two saw cuts. #731 says that, and dimensions the cuts at 17.8 mm apart.

Closing this one. Review #731.

PR #735, opened 2026-09-21
catalog: the BOM buys a Pico with its headers already fitted

basically-balloon:
The catalog has specified a bare Raspberry Pi Pico plus 40 loose header pins since 2026-07-18, and a builder's first job has been soldering them on. barthel asked for that to stop being the default.

It has been a sanctioned alternative since June: he asked in `#electronics` [2026-06-14](https://discord.com/channels/1430279849171222682/1437436041685631129/1515624967981629440) whether a Pico with pre-soldered headers could be used, and Jon (**@effreek**) answered [2026-06-15](https://discord.com/channels/1430279849171222682/1437436041685631129/1515896319212912712): *"The pre soldered header version can be used as an alternate. Just saves you the trouble."* Nobody had changed the catalog since, so "may I buy one" was settled and "what does the BOM say" was not.

## The new part

`mcu-rpi-pico-presoldered`, uid `dq0q`, 1 per machine. The same RP2040 board with both rows of 20 2.54 mm pins factory-fitted and the three DEBUG pads left bare, which is mechanically identical to what the old route produced by hand: same board, same rows, same seating into the control board's sockets, nothing downstream changes.

Sourced from the [waveshare listing](https://www.amazon.com/waveshare-Pico-Microcontroller-Development-Pre-soldered/dp/B0912M3T1H) on Amazon US, **$11.39**, read off a screenshot of the live page on 2026-09-21 rather than a cached copy.

**The catalog photo is not the seller's gallery shot.** That one has a loose 1x3 debug header and a micro USB cable in frame beside the board. barthel asked for a picture of the Pico itself, so the row carries the plain manufacturer shot he picked on 2026-09-19 for the helper page: a Pico on white with both rows of 20 soldered on, the joints visible on the chip side and the three DEBUG pads bare. Same bytes as `sorter-docs/pico-headers-soldered-plain-full-fc3efcd791b5.jpg`, republished to the parts bucket, which is why the content hash matches.

**The official Pico H was deliberately not chosen.** It carries a 3-pin JST-SH debug socket standing off the top face, and nobody has checked that it clears the control board housing cover. The waveshare board has bare debug pads and raises no such question. It is also the board in the photo barthel picked for the helper page on 2026-09-19 (*"without the debug port"*).

## What is retired

`mcu-rpi-pico` (uid `hx8l`) and `header-pins-254` (uid `10o3`). Both lose their `sheet_qty` (1 and 40 per machine) so they reach no bill of materials, both carry a retired `description` and `note`, and `delete-bare-pico-and-header-pins` marks them at P4 with `condition: "retired"`. The objects themselves stay, because `check_generated_pins.py` refuses any change that makes a minted uid disappear; **sorter-v2#734** is the `bom` issue tracking the actual deletion.

**The header pins were not named in the request and are included anyway.** Their only use anywhere in the catalog is fitting them to that Pico, so leaving `sheet_qty: 40` behind would go on telling every builder to buy 40 pins nothing needs. Easy to back out if that is not wanted.

## Docs

- **Preparing the control board** takes `mcu-rpi-pico-presoldered` as a `parts_needed` line. Its prep item now leads with the fitted Pico as the standard ("The Pico comes with its two rows of 20 pins already fitted"), and soldering them yourself is one closing sentence carrying the link to the helper page. Neither the bare Pico nor the pins is listed as a required component on that page. The item's figure is untouched, so #731's swap of that photo still applies cleanly.
- **Electronics installation** no longer opens the list with the soldering step.
- **Soldering Pico headers** opens with a callout saying the step is optional, that a standard build skips the page, and why its own parts list shows two parts badged "to be removed": they are not counted into a machine and the catalog has both flagged for deletion.

`npm run check` clean on both apps, `validate_images.py` clean, `generate.py --metadata-only` clean (115 pinned parts, versioning, 60 connections).

## One extra commit: the `/changes` link had to move off an anchor

`delete-bare-pico-and-header-pins` is the first entry to target a part that lives inside an all-hardware rollup. `/changes` linked hardware targets to `/hardware#hardware-<id>`, and the hardware page only renders a rollup member's own row once the rollup is expanded, so the id is not in the prerendered HTML and SvelteKit's prerender failed the parts-calculator build on the first push.

Both links now go to `/hardware?hw=<id>`, which opens that part's detail modal and is the same query-not-anchor reasoning the `assemblies` branch two lines above already carries. `npm run build` is green.

## The alternative had to become a link, and the assembly joins the retirement

barthel, on the card: the helper page was named in the Pico's `alternative` text as a bare path, which is a path the reader retypes. Three changes, one place each:

- **`[[docs:<path>|<label>]]`** joins `[[hw:]]` and `[[part:]]` as a catalog prose reference, resolved against `DOCS_BASE` exactly the way an assembly's `docs` is. A path only, so a full URL cannot outlive the host it names; anything else stays a raw token. `plainDescription()` flattens it to its label for search and CSV.
- **`AlternativeBadge`** renders its detail through `CatalogText` instead of as a string, so any alternative naming another part or a docs page links to it.
- **The hardware card** shows the alternative at all. It was on the list row and inside its popover only, so the detail view — the thing you read before buying — never said a part had one. The card also renders `docs_page`, a field that has sat on four parts (`dupont-lead-2p-1m`, `jst-vhr-2`, `jst-phr-4`, `dc-jack-5521-inline`) and been drawn nowhere.

Catalog: the pre-soldered Pico's `alternative` carries the token, `mcu-rpi-pico` and `header-pins-254` gain `docs_page`, and **the `pico-headers` assembly joins `delete-bare-pico-and-header-pins`** under `targets.assemblies`. Hand-soldering the pins is an optional route now, so the assembly goes with the parts.

`npm run build` green; both links verified in the prerendered `/part/mcu-rpi-pico-presoldered` and `/part/mcu-rpi-pico`.

## The Pico is in the assembly tree, and two notes came off

barthel asked for the Pico's `note` about its bare DEBUG pads and the loose 1x3 header to go, and for the card's *"This count is a hand count carried over from the BOM spreadsheet"* caveat with it. The second one is not text: it is what the card says whenever a part's quantity comes from `sheet_qty` rather than from the tree. So the fix is to place the part.

**`ctrl-board-mount` v3** (uid `p5yl`, superseding `rns2`) takes three lines: **5 × `drv-tmc2209`**, **1 × `mcu-rpi-pico-presoldered`**, **10 × `jumper-cap-254`**. All three are fitted to `ctrl-board-basically` before its cover goes on, all three were carried as `sheet_qty` hand counts off the BOM spreadsheet, and all three are exactly what Preparing the control board has always listed alongside the board. `breaking: false` — nothing about the housing changes, an existing one is still an existing one.

**Only the Pico was named.** The drivers and the jumper caps had the identical defect on the identical board and showed the identical caveat, so they went in together rather than leaving two thirds of one board's parts outside the tree. Easy to pull back to just the Pico.

Version ceremony per `VERSIONING.md`: uid minted, `versions[]` entry authored with `"commit": null`, `stamp_versions.py` run (it backfilled `rns2` and the v2 line snapshot), the thirteen unrelated nodes its repo-wide sweep also stamped restored from `HEAD`, and v3 left pending because the board squashes. `check_versioning.py --base origin/main`, `check_generated_pins.py` and `npm run build` all clean; the card now reads "Needed 1 — summed from the assembly tree".

## Open, and not decided here

`/hardware/helpers/pico-headers/` stays. It is the documented route for anybody who already owns a bare Pico, it is linked from the pre-soldered Pico as its alternative, and it says on itself that the step is optional and that its two parts are not counted into a machine. Only the `pico-headers` assembly node is marked for deletion, not the page.

Nothing else is outstanding.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #621, opened 2026-09-10
docs(chute): chute core and door module rebuilt around a real build

basically-balloon:
Two chute pages rebuilt around BrickCycleAlice's photos of her own build, replacing renders where a real part says it better and adding the checks she hit on the way.

**Requested by BrickCycleAlice (@brickcyclealice)**, over a run of instructions in one thread, and by **barthel (@uwe_barthel)** [in review](https://discord.com/channels/1430279849171222682/1547548160262545469/1547573059819413525): *"Remove the first of the doubled images on chute core side and add a "the finished result" section (as we do in bin frame). Move the last photo in the finished result section and use it as preview image (as we do in bin frame roo)."* Each instruction is linked on the change it produced.

## Chute core

[Step 1](https://discord.com/channels/1430279849171222682/1547526128942583838/1547546044533833799): the three pocket renders replaced by photos of a printed core with the inserts in. Which photo is which face was settled by counting brass against the documented 8 + 6 + 4, not by shape, since the long sides are mirror images: **32** is the 8 side (its tell is the lone insert in the raised square pocket), **30** the 6 side, **33** the rear panel.

Its [built-core photo](https://discord.com/channels/1430279849171222682/1547526128942583838/1547569567545561088) of hers closes the page. It replaces the placeholder that stood after step 2, sits under a new `## The finished result` heading in the pattern the bin frame pages use, and is the page's `og_image`. That shape is [barthel's](https://discord.com/channels/1430279849171222682/1547548160262545469/1547573059819413525): the finished shot at the foot rather than duplicated at the top, and the same image doing the preview. The caption names the layer connectors, which [she pointed out](https://discord.com/channels/1430279849171222682/1547526128942583838/1547572315909390467) are in the shot and go on the top of each chute; the layer adapter board is the only one of the four not in it, since she does not have a board yet.

## Door module

- [Step 1](https://discord.com/channels/1430279849171222682/1547526128942583838/1547546693736603739): insert renders replaced by **35** (servo bracket housing), **36** (bearing race), **38** (bearing holder), and **37** added as her tip — with no bench vice, a holder stands in the housing's pocket while you press its inserts.
- [Step 2](https://discord.com/channels/1430279849171222682/1547526128942583838/1547548046467014676): **42** servo in the pocket, **43** cover screwed down.
[...]

PR #739, opened 2026-09-21
docs(door-module): step 3 had the two arms' photo pairs swapped

basically-balloon:
Extremetaz hit this on a real build: at step 3 the M3x8 would not go into the arm the upper photos show. It will not, and the page was at fault, but not in the way it first looked. **The screw lengths are right against the STLs; the two photo pairs are swapped.**

The upper pair (`...lower-arm-on...`, `...lower-arm-screws...`) shows the **side arm**, and the lower pair (`...side-arm-on...`, `...both-arms-underside...`) shows the **lower arm**. I placed those photos in #621 off the parts' bounding boxes, and read the side arm's big 62 x 46 mm face as "the wider of the two arms". It is the thinner one, 16.7 mm, and it just presents its large face to the camera.

**What settles which photo is which** is the screw heads, not the outline. Measured off `servo-bracket-lower-arm` and `servo-bracket-side-arm` against the housing's insert pockets, which land on the same coordinates (the parts are exported in their assembled pose):

| joint | plate | above the insert | M3x8 | M3x12 |
|---|---|---|---|---|
| lower arm to housing | 8.42 mm, with a O7.16 x 3.42 mm counterbore | 4.42 mm | 3.58 mm of thread | 7.58 mm into a 5.70 mm pocket, bottoms out |
| side arm to housing | 8.42 mm, plain countersink | 7.84 mm | 0.16 mm, never bites | 4.16 mm of thread |

So the lower arm's screws sit at the bottom of a deep round well and the side arm's finish flush. The second photo on the page shows two flush cross heads, so it is the side arm; the fourth photo shows two heads sunk in wells, so that face is the lower arm's.

**The change:**

- The two photo pairs swap, and the side arm now comes first, because the first photo shows one arm on and that arm is the side arm.
- Captions and alt text rewritten to match what is actually in each frame.
- Each arm gets a physical tell up front (slab on the side with flush screws, versus the one under the housing with screws down a well), which is what step 3 was missing.
- The "it is the wider of the two arms" line is gone, and the numbers are the measured ones: 7.84 mm above the insert on the side arm, 4.42 mm on the lower arm.

Fastener counts, the parts list and the step 8 ear screws are unchanged and were already correct.

## Mira pass, second commit

barthel asked in review not to forget who reads this page, and the first commit had put a fair amount of the measuring into it. The prose now says which screw to use and what goes wrong with the other one; the arithmetic behind it is the table above.

- **Side arm:** "Use the 12 mm on this arm. An 8 mm screw does not reach the insert at all." (was 7.84 mm of plate, 4.16 mm of reach)
- **Lower arm:** "Use the 8 mm on this arm. A 12 mm one bottoms out in the well before its head seats, so the arm never pulls down tight." (was the 3.42 mm counterbore, 4.42 mm above the insert, the 5.70 mm pocket)
- **The tells keep the outcome and lose the vocabulary:** flush with the face, versus down at the bottom of a well. "Plain countersinks" and "counterbored first" are our words, not hers.
- **Step 2's cover line got the same treatment**, one paragraph up and the same shape: "Use the 12 mm screws here, not the 8 mm: an 8 mm one barely enters the insert." It was not part of the photo fix, but it is the same page and the same reader.

The photo-to-arm assignment, the fastener calls, the counts and the alt text are unchanged.

**One gap left, not fixed here:** the page has no `tools_needed`, and neither does any other page under `assembly/distribution/chute/`. I cannot tell from these photos whether the M3 countersunk heads here are hex or cross, and the catalog does not record a drive type, so naming a driver would be a guess. Worth a section-wide pass.

## Tool list, third commit

`tools_needed: ["Hex key, 2 mm", Soldering iron or heat-set insert press]`. Every fastener on the page is an M3 countersunk socket screw, so the 2 mm hex key covers steps 2, 3, 5, 6, 7 and 8, and the iron or press is step 1's heat inserts. Nothing else on the page needs a tool: the bearings press in by hand and the vice tip already offers the housing as a jig.

**On the drive type**, which I said yesterday I could not settle from the photos: the catalog's M3 countersunk screws are hex socket. Every `scr-m3-*-cs` row carries `Head: Countersunk (flat) socket`, the vendor listing behind the 8 and the 12 is a socket-drive countersunk, and the catalog's own product photo shows hex sockets. The cross-recess heads visible on the cover and side-arm screws in the build photos are a builder's own substitute, so they do not set what the BOM buys.

No other page under `assembly/distribution/chute/` has a tool list. This one is not the fix for that.

## The rest of the Mira pass, same commit

Steps 4 to 8 carried the same measured prose the arm steps did, so it had the same treatment. Every trap stays, in the instruction rather than behind arithmetic.

- **Step 4:** the shaft's Ø19.8 mm and the spigot's 12.0 mm across the flats go; "one end finishes in a hexagonal spigot" is what she matches by eye.
- **Step 5:** the rim's 5.00 mm height goes, the "if a bearing has not gone home its cover will not pull flat" check stays.
- **Step 6:** "into 5.00 mm of wall" and the whole "8 mm is the right length because ... Measured off the STLs" paragraph go. The bullet already says to use the 8 mm.
- **Step 7:** the 15 mm radius and the 84°/96° hole spacing go from Alice's tip; "not spaced evenly, so it only drops on in two of the four quarter turns" is the part she acts on.
- **Step 8:** the ear thicknesses and reach figures go, leaving "an 8 mm screw would not reach the insert at all" and "a 12 mm one would bottom out in the pocket". The adapter socket's 12.4 mm and the 0.4 mm of slip-fit clearance become "it is a slip fit, so it pushes on without force".
- The door's **80.5° of travel** stays in the step 8 warning, which is the one measured number on the page a builder acts on.

The numbers themselves are in the notes behind this PR and in #621's body, not lost.

## Fourth commit: the intro and step 4 were still out of step with the rest

Asked in review whether Mira would be happy with the page as it stands. Read end to end, she would not have been, and two of the reasons were mine: the third commit trimmed steps 4 to 8 but missed step 4's numbered list, and the intro was still quoting the ear thicknesses step 8 had stopped quoting.

Taken out:

- **The intro's provenance paragraph** ("Every step from 2 to 7 now carries photographs of a real build ... Correct this page as you go"). It is provenance, it was already out of date (every step carries build photos now), and it asks the reader to fix the documentation, which is not something she does.
- **The ear thicknesses** in the fastener split, now that step 8 does not quote them. The fastener call stays.
- **"No heat inserts, the screws thread directly into the printed plastic"** from the servo adapter overview. Same sentence barthel had removed from a Preparation step on 2026-09-12.
- **The insert pocket's Ø4.2 x 5.7 mm** in step 1. What stays is which parts take none.
- **The parts-calculator reconciliation** ("the counts match ... 6 + 4 + 3 + 3"). The parts list is above her already.
- **Step 4's bore and clearance figures** (Ø20 bore, Ø19.8 shaft, Ø21.0 hole, 0.6 mm all round, 0.5 mm over the shaft). What stays is "neither should need forcing" and "if it does touch, something is not seated".
- **The 10.7° and 88.9° swing extremes.** The 80.5° of travel stays, because the step 8 warning rests on it.

## Two things left that are not mine to decide

- **The intro's "What each fastener is for" list** is the pre-loaded reference table barthel had removed from psu-box.md: four bullets splitting the module's fastener totals per sub-assembly, every one of which the step that uses it states again. Taking it out is a structural edit to a page section nobody has complained about, so it is not in this PR.
- **No `## The finished result` section and no `og_image`.** The page ends on the servo video. The house pattern wants a heading, the finished-module photo under it, and that same photo as the preview, which means moving one of step 8's two shots down rather than duplicating it.

## One thing that would still stop her, and it needs somebody who has done it

Step 8 says *"centre the servo (or let it settle at its power-on default)"*. Nothing on this page powers a servo, and centring is the only instruction here that needs something outside the page. The video may well show it; the page does not say. I have not invented a procedure for it.

## Fifth commit: barthel's call on the two open points

- **The intro's "What each fastener is for" list is out.** Four bullets of per-subassembly totals, every line of which the step that uses it states again. One clause in it lived nowhere else, that the bearing covers are thin at the rim and their screws want snugging down evenly rather than one tightened fully first, so that moved into step 5.
- **`## The finished result` is in**, at the foot, with a lead line, the `bracket-on-core-top` photo he named, and the pointer back to the chute core page. That photo moves out of step 8's pair rather than being copied, so the finished shot appears once, and it is now the page's `og_image`. Step 8 keeps the face-on shot as a single figure.
- **The servo position needs no change**: barthel says the video covers it, and step 8 already points at the video for how the coupling is set.

## The finished-result image stays as it is

Asked in review to find a better one. There is not a better one yet, and the search is worth recording so it is not repeated:

- **Of the 26 `assembly-door-module-*` assets, none shows the finished module.** The five wide, well-lit shots (flap closed, flap open, both sides, from the back) are all from step 6, before the servo bracket goes on. Only `bracket-on-core-coupled` and `bracket-on-core-top` are after step 8, and both are tight crops. The one in the section is the only image with the module on the core and the servo coupled.
- **A render cannot stand in.** `servo-mg995` and `mg995-servo-horn` are `cots` rows with no `stl_hash`, so the servo bracket renders as an open shell with nothing inside it. Rendered it to check, and that is exactly how it comes out.
- Doing that did settle how the module goes together, which is now in my notes: the parts are exported in three frames, the bracket group sits exactly 36.00 mm below the door's shaft axis, and the bearing covers are on a perpendicular axis and place off their two screw holes against the holder's insert pockets (24.588 mm apart on the holder, 24.57 mm on the cover).

**What would fix it:** one wide photo of a finished module, or a frame from the servo video if that is an acceptable source.

## Sixth commit: the finished result is a real photo now

BrickCycleAlice posted a finished module on a chute core in the review thread and barthel asked for it as the finished result and the preview image.

- **`IMG_4768` is the one used**, uploaded as `assembly-door-module-on-core-finished`. It shows the servo in its bracket, the adapter coupled to the door's shaft and the door hanging below the core, which is exactly what step 8 leaves behind.
- **Her other two shots (`IMG_4772`, `IMG_4773`) are not used.** They have the funnel fitted, which belongs to the funnel brackets page and which she says is coming back off until the chute is in her tower frame. A preview with a part this page never fits would send a builder looking for it.
- **`bracket-on-core-top` goes back into step 8** beside the face-on shot, restoring that pair, so the finished shot appears exactly once on the page.
- `og_image` is the new photo, and she is now in the page's `contributors`, which she was not despite having taken most of its photographs.

Image validator, credits validator, front matter validator and `npm run check` all pass.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #745, opened 2026-09-22
docs(channel stepper lead): re-housing the shipped lead is a third route

basically-balloon:
Asked for by **BrickCycleAlice** in `#balloon-electronics`, [here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552102983276306463): change the page to cover the route she is actually building, and put the `PHR-4` in Parts needed so the crimp kit is not the only way in.

### What changed

The page had two routes, both of which start by setting the shipped lead aside. She is doing a third thing: **keep the lead the motor came with, cut its Dupont end off, crimp a `PHR-4` on in the right coil order.** That answers both of the objections the page raises against the shipped lead (the Dupont goes in the bin, and the order is fixed while the new housing is loaded), and it keeps 1 m of cable instead of the 63 cm of the ready-made.

- **New section, first of the three routes**, with the meter-first coil check, the strip length, the crimp, and the load order.
- **`Why not the shipped lead` becomes `What is wrong with the end it comes with`**, because "not the one you use" is no longer what the page says. The two faults stay exactly as they were; what is added is that both of them live in the one housing, and that the motor end and the crossover are already correct inside the cable.
- **Parts needed gains `jst-phr-4` (4) and `jst-sph-002t` (16)** alongside the existing cable and `PHR-6`. Both cards already carry an alternative naming `connector-kit-crimp`, so a builder sees the one-box option and the buy-them-separately option without the page taking a side. Each route now states its own per-lead parts in its first line.
- **Tools needed** split into "only if you crimp" and "only if you move contacts".

### Facts this rests on

- The shipped lead is **26 AWG**, motor connector `PHR-6`, 1 m, and the motor's phase resistance is **2.3 Ω**: [17HE15-1504S datasheet](https://www.skysmotor.com/images/upload/File/17HE15-1504S.pdf). 26 AWG is inside the 24 to 28 AWG `jst-sph-002t` takes, which is what makes this route legal at all.
- StepperOnline's colour key, from their own listing: `A+` black, `A-` blue, `B+` green, `B-` red. The page still says to meter rather than trust it, because their listing also warns that the middle two are swapped on some motors.
- The as-shipped Dupont order was read off [her photo](https://discord.com/channels/1430279849171222682/1552083234249248842/1552100890331054221) (red, blue, green, black), which interleaves the coils. It is not quoted on the page, only the fix is.
- `connector-kit-crimp` carries PH housings 1 to 5 pin, so the `PHR-4` is in the box and the `PHR-6` it lacks is the end this route does not touch. From the catalog entry, [asked about here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552102082431950890).

### For the reviewer

**Route order is a call, not a fact.** I put the re-housing route first because it needs the least bought, but the page is Jon's and the 2026-09-14 decision was that the shipped lead is not used. If that decision was about the lead as it comes rather than about its wire, this is consistent with it; if it was meant more broadly, move the new section below the other two or drop it.

Nobody has built a lead by these steps yet, and the page's AI-generated-first-draft warning still says so.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #746, opened 2026-09-23
docs(board 24 V lead): say red is +24 V where the ends are listed

basically-balloon:
Flagged by **BrickCycleAlice** in `#balloon-electronics`, [here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552107490295160902): red goes with + on the barrel plug, and the page does not list it.

She is right that it is not where you would look for it. The page said it once, inside build step 2, as part of the instruction to meter the plug. Anyone reading "What the two ends are" to find out which conductor is positive found the plug's size, its centre-positive rating and the wire's two colours, but never the two joined up.

### The change, four lines

- **PSU end**: "centre-positive" now spells out that the tip is +24 V and the sleeve is ground.
- **Wire**: red is +24 V and black is ground, noted as the same convention as every other DC lead on the machine.
- **Step 2** keeps the meter instruction and now says why it is worth doing anyway: the moulded plug is somebody else's lead and a few are wired the other way. The colour says what to expect, the meter says what you have.

Nothing else moved, and the polarity itself is unchanged: `W1` is `RD`/`BK` in `electronics/wire_harness/board-power.yml`, and `J1` position 1 is +24 V through `F1`.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #747, opened 2026-09-23
docs(Pi 24 V lead): photos from a real build, and the splice becomes conditional

basically-balloon:
Asked for by **BrickCycleAlice** in `#balloon-electronics`, [here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552113634174439535), with both photographs: use the built lead as the finished result, the as-it-arrives shot as the prep, and make splicing conditional on the converter's leads being shorter than 100 mm.

### What changed

- **Splicing is now step 2, and only under 100 mm.** The page used to open by splicing a moulded plug's pigtail onto the converter's wires as the only route. The converter ships with about 120 mm of its own, so step 1 is now "measure them; 100 mm or more and nothing gets spliced", and the splice step says to skip it.
- **Step 3 and 4 cover both plug styles.** A screw-terminal plug (what she used, no solder) and a moulded one. The meter-the-centre-pin check stays for both, because that is what tells you which side is the tip.
- **`## The finished result`**, following the house pattern: lead line, single figure, then the pointer into connecting the components, with the two-identical-USB-C-sockets warning under it. The old `## Where it goes` heading is folded into it rather than duplicated. The photo is also the page's `og_image`, which it did not have.
- **The as-it-arrives photo** opens `Build it`, showing the converter, its short input pair and the loose plug.
- **`Wire` in "The two ends" now says "None, usually"**, with the 150 mm harness figure and the measured 120 mm beside it. `wire-22awg-2c` stays in Parts needed because the short-lead case still needs it.
- **Tools needed** splits: multimeter and a small screwdriver always, the iron and heat shrink only if you splice.
- **The AI-first-draft warning is retired.** One builder has now made this lead, so the warning says what is still unsettled instead: the finished length has never been measured on a mounted machine, and the 22 AWG is still a guess on the drawing.

### For the reviewer

**The catalog disagrees with the screw-terminal plug.** `dc-plug-5521-male`'s description says to buy a moulded plug and splice to it, and that a screw-terminal plug "works but is bulkier and comes loose". The page now presents both without ranking them, and says to tighten firmly and tug each wire. If that catalog line is a real preference rather than an aside, the page should say so and this is the place to tell me.

**The builder's own reasoning, [since asked](https://discord.com/channels/1430279849171222682/1552083234249248842/1552115810515034172):** the screw-terminal plug "was much easier to source, plus felt easier", and she is heat shrinking all of these connections. So the case against it is a tightness-over-time worry, not an availability or difficulty one, and heat shrink over the barrel answers part of it. Still worth a maintainer's view on whether the catalog line should soften or the page should rank the two.

Both photographs are hers, credited `Photo: BrickCycleAlice`, uploaded to the `sorter-docs` namespace. She is measuring the finished length once the converter is mounted, so the 150 mm figure may move again.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #748, opened 2026-09-23
docs(chute stepper lead): name the dial position, and the 24 AWG tail is not optional

basically-balloon:
Asked for by **BrickCycleAlice** in `#balloon-electronics`, [here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552115810515034172): put the dial symbol in step 1 of "Find the coils first", so the instruction matches what is printed on a meter.

### The change, three lines

- **Step 1** now reads "Set the multimeter to resistance, the **Ω** position on its dial." Her meter (a Digitor Q1585) has `Ω` and a separate continuity position, which is exactly the ambiguity the step had.
- **Step 2 gets the real number.** It said a coil pair "reads a few ohms". On this motor it does not: `23HS32-4004S` is **0.65 Ω per phase ±10%** ([datasheet](https://www.skysmotor.com/images/upload/File/23HS32-4004S.pdf)), so a builder reading 0.7 on a cheap meter and expecting "a few" has no way to tell a good pair from a bad probe. The step now says well under an ohm, names the 0.65, says the probe leads add to it, and says the other pair shows `OL` or a lone `1`.
- **The closing callout** followed the same correction, since it repeated "a few ohms".

The 4 A NEMA 23 is the low-resistance one. "A few ohms" is right for the channel steppers (2.3 Ω on the `17HE15-1504S`), which is where the phrase came from, and their page is not touched here.

### Second ask, same page: the motor's leads are 20 AWG

[Here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552126192453423155), with photographs of the motor mounted and its leads reaching right around the electronics level: can the JST go straight onto the motor's own cable? No, and length is not why.

- **The leads are `UL1007 AWG20`** on StepperOnline's own drawing for `23HS32-4004S`. `SPH-002T` takes 24 to 28 AWG. Four gauges over will not crimp with the strands captured.
- So **step 1 no longer offers "use them as they are"**. A 24 AWG tail is spliced on every time, at least 100 mm even when the motor's leads already reach, and the contacts crimp onto the thin ends. A callout under the intro says the same thing before a builder starts.
- **The soldering iron and heat shrink leave "only if you extend the leads"** in Tools needed and become unconditional.
- **Lead length:** the drawing says 300 mm, her motor shipped with 500 mm. The page now says 300 to 500 depending on the batch, and the front-matter warning says to check yours. The "whether the motor's leads are long enough to crimp directly is not recorded" clause is gone, because the answer is that it never matters.

### Not done, on purpose

Three other pages say "set the multimeter to continuity" without naming the symbol (`board-24v-lead`, `pi-24v-lead`, `psu-pigtail`). Continuity is a drawn symbol rather than a character, so it is a different edit, and nobody has asked for it. Say the word and it can follow.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #749, opened 2026-09-23
docs(limit switch lead): photograph the tabs, on the bench and in the machine

basically-balloon:
Photographs from **BrickCycleAlice** in `#balloon-electronics`, [here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552122537281064981): the switch's terminal positions, off the machine and on it.

### What changed

- **`Which position is which` gets the two shots as an `img-row`**: the receptacles on `COM` and `NC` with the bare `NO` blade between them, and the same two tabs once the switch is in its housing under the top plate. The section named three tabs and showed none of them.
- **A new line above them: the switch prints its own tab names.** The body carries a schematic with `NC` and `NO` against the side tabs and `COM` against the bottom one, which the page did not say and which is the fastest way for a builder to get this right.
- **The parts shot opens `Build it`**: the switch, two #187 receptacles and the stripped pair.
- **The front-matter warning is rewritten.** It was an AI-first-draft banner; the page now has build photographs, so it says instead that the build shown is wired the way the page says and that the two harness guesses (board pin order, tab choice) are still guesses. The "meter it" instruction stays in both places.
- **`og_image` set** to the terminals shot. The page had none.

### For the reviewer

- **This does not verify either guess.** The build in the photographs follows the page, which is not the same as a running machine confirming that `NC` is the tab the firmware wants. The callout saying so is unchanged.
- **No `## The finished result` section**, because the finished lead is not photographed: both shots are the switch end, and the keyed Dupont at the board end is not in frame. Worth adding when there is a picture of the whole lead.
- **The mounted shot may be wanted elsewhere too.** [Top interface](https://docs.basically.website/hardware/assembly/distribution/top-interface/) step 4 covers the limit switch housing and has no photographs at all. Same picture would serve it; not done here, one page per pull request.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #750, opened 2026-09-23
PSU box: name the mains lead, give it a card, and identify live by test not colour

basically-balloon:
Asked for by **BrickCycleAlice** in `#balloon-electronics`, [here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552128948123729921), after [asking what to use](https://discord.com/channels/1430279849171222682/1552083234249248842/1552128481221943327) to get power to the inlet and finding the answer nowhere in the docs.

### The gap

The page tells you twice not to put a mains cable in the inlet, and never says what that cable is. `psu-switch-fused` is in the catalog; a cordset is not, so a builder working through the BOM has nothing to tell them the last link to the wall is theirs to supply.

### The change

- **Step 2 gains a paragraph** naming it: an ordinary **IEC C13** lead, PC or monitor cord, with the local plug. Three core, because the machine earths through it to the supply's earth screw. 10 A or better, matching the module's own fuse. And explicitly that it is **in no parts list**, so nobody goes hunting for it in the BOM.
- **A closing callout**: once the machine is in service, pull the cord before opening the box, because the rocker may be breaking the neutral rather than the live. Nobody has recorded whether that module is single or double pole, so the wording says "may be".

### The lead colours are not dependable, and that changed the page

She photographed her module's leads on the block, then [metered it](https://discord.com/channels/1430279849171222682/1552083234249248842/1552133464403484703): **on her unit the blue lead is the live one.** That is the second of these modules found the opposite way round to the seller's photo, after Bill's in September.

Landed by colour on a unit like hers, the module's fuse and rocker end up in the **neutral**, and the supply stays live with the switch off. So:

- **Step 2 now lands live and neutral by name, not by colour**, and carries a warning that both orders have been found in the post.
- **It gives the test**: fuse out, meter each lead to each of the two flat pins in the C14. The pair still reading continuity is neutral; the one that only reads it with the fuse in and the rocker on is live, because the fuse and switch are in the live side.
- **The figure is her module after the swap**, [reshot once she had tested it](https://discord.com/channels/1430279849171222682/1552083234249248842/1552133788325380096): blue on `L`, red on `N`, yellow on earth. It now illustrates the test instead of contradicting it, and the caption says outright that another unit of the same part can be the other way round. The earlier shot, with red on `L`, is not used.
- **The step 1 drawing still shows red to `L`.** Its caption now says those colours are one module's and to check yours. Redrawing it to label the leads live and neutral is the proper fix and wants the diagram's source, which is not in the repo.
- **`psu-switch-fused` carries the warning too**, so it reaches anyone reading the BOM rather than the page.

### And a card for it, [asked for here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552130023815905411)

"So people gathering what they need for the page collect it." A card means a catalog entry, so there is one now: **`mains-cord-c13`**, category Power supply, `sheet_qty.per_machine: 1`, and it is in the page's `parts_needed` right after the inlet.

- Attributes carry the four things that decide whether the cord in your drawer will do: C13 at the machine end, local plug at the wall, **three core** (a 2-core figure-8 lead neither fits nor earths the machine), 10 A or better.
- The step 2 prose lost its "it is in no parts list" line, which this change makes false.
- `catalog.generated.json` regenerated with `generate.py --metadata-only`; its three checks pass (115 pinned parts, versioning, 60 connections).

**Sourced**, [asked for here](https://discord.com/channels/1430279849171222682/1552083234249248842/1552130158364987403) ("or be detailed in the bom, so people know to expect it"): Amazon Basics 6 ft NEMA 5-15P to C13, 18 AWG, UL listed, **$6.94**, `B072BYGKZZ`. Price read off a screenshot of the live listing today, not off a cached copy. The vendor note says outright that this is the US plug and that elsewhere you buy the same cord with your own, so the row is actionable without pretending the cord is universal.

**No product photo, on purpose.** The card shows the "no image" tile. The right cord differs by country and a picture of a US plug on the card would tell a UK or EU builder the wrong thing; the C13 end is the half that is universal. Say the word if you would rather have a photo of just that end.

**Not added to the assembly's `lines`.** That would be a version stamp on `meanwell-psu-box`, and whether a cord the builder already owns belongs in the machine tree is a call for whoever owns the BOM, not a side effect of a docs fix.

### Beyond the ask

The second item is mine, not hers. It is the safety wording Jose R asked for in `#v2-general` on 2026-09-18 when the reversible-plug question came up, and the page never got it. Drop it if you would rather keep the page to the build and put in-service safety somewhere central.

No part numbers, no catalog change, and no claim about the module's pole count.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

PR #755, opened 2026-09-23
docs(chute): tool lists on the four pages that had none, and a photo for funnel step 1

basically-balloon:
The two loose ends from [#739's review summary](https://discord.com/channels/1430279849171222682/1551696986174914572/1552239006115168328), which [barthel asked to have fixed](https://discord.com/channels/1430279849171222682/1551696986174914572/1552247662177292340).

## 1. No page under `assembly/distribution/chute/` had a tool list

A builder standing at the bench is told which fastener at every step and never what to drive it with. Each list comes from that page's own steps, not from the section:

- **`chute-core.md`** — `[Soldering iron or heat-set insert press]`. Step 1 presses its 18 inserts; step 2 bolts the four sub-assemblies on, and every one of those screws belongs to the sub-assembly's own page, which now carries its own list.
- **`funnel.md`** — `["Hex key, 2 mm"]`. Four M3 countersunk into the core's inserts, and the funnel itself snaps in with no fastener.
- **`layer-connectors.md`** — `["Hex key, 2 mm"]`. Two M3 countersunk per connector.
- **`pcb.md`** — `["Hex key, 2 mm for a button head or 2.5 mm for a socket head"]`. The board's M3 × 6 is `scr-m3-6-bhcs`, the one socket-or-button head on the chute, and the two heads take different keys. Every other fastener in the section is countersunk, which is 2 mm.

`door-module.md` is not touched here: its list is in #739.

## 2. Funnel step 1 had a placeholder asking for a photo that exists

`<div class="img-placeholder">Photo of both funnel brackets fitted to the chute core.</div>` is now BrickCycleAlice's `IMG_4769`, a bracket bolted to the core seen from the servo end with both screws home. Her shot of the far side is the page's finished result already, so this is the other frame rather than the same picture twice, and the caption says the second bracket is the mirror of it on the other long side.

## 3. One sentence beside it, flagged because nobody asked for it

Step 1 also said *"The bracket is 8.16 mm thick at the screw, so a 12 mm screw reaches 3.84 mm into the core's blind 5.70 mm insert and an 8 mm one would not reach it at all."* That is the shape that has come out of `door-module.md` all week, so it now reads *"Use the 12 mm, not the 8 mm: an 8 mm screw does not reach the core's insert at all."* Same trap, no arithmetic. Drop this hunk if you would rather keep the PR to the two things asked for.

Image, credits and front matter validators clean, `npm run check` clean, and all four front matter blocks parse.


basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

Issue #639, opened 2026-09-11
Machines can't recover on their own from a root filesystem error


spencerhhubert:
## Summary

On 2026-09-09 the root filesystem on Chris K's machine ("Batman's Brick Stash") hit an error while it was sorting. The machine could not recover on its own. It limped along read-only for about 11 hours, dropped off the network, and then every boot stopped at an interactive fsck prompt. It was down about 1.5 days and came back only after the owner plugged in a monitor and keyboard and ran fsck by hand.

The card failing is one problem; this machine's card has corrupted data before (#331). The bigger problem is that nothing in the image lets a machine get back up without someone at a keyboard, and nothing off the card holds its setup.

## What happened

- **2026-09-09, evening:** disk writes stop persisting. The last `chrony.drift` write and the last `fake-hwclock` save are minutes apart, and the ramlog sync due at midnight never lands. The backend stops reporting to Hive about 25 minutes later. Hive was healthy and taking data from other machines the whole time.
- **About 11 hours later:** the machine drops off the network entirely.
- **Every boot after that** stops in the initramfs:
  ```
  fsck.ext4 -a -C0 /dev/mmcblk1p1
  opi_root contains a file system with errors, check forced.
  Inodes that were part of a corrupted orphan linked list found.
  opi_root: UNEXPECTED INCONSISTENCY; RUN fsck MANUALLY.
  Failure: File system check of the root filesystem failed
  (initramfs)
  ```
  Two power cycles, a wired ethernet connection, and restarting the network changed nothing, because boot never got as far as networking.
- **2026-09-11:** the owner ran `fsck.ext4 -y /dev/mmcblk1p1` at the prompt and it booted clean. `lost+found` is empty, all tracked sorter source matches git, and there have been no mmc or ext4 errors since.

"Contains a file system with errors" means ext4 recorded an error at runtime, not only an unclean shutdown. This root was already `journal_data_ordered`, so #328's writeback mode is not the cause here. The kernel messages from the failure are gone: `/var/log` is zram (orangepi-ramlog), and its last sync to disk was that morning.

## Why a card error became a 1.5-day outage

Checked on the machine, which runs stock Orange Pi OS 1.2.2 Jammy. Nothing in `software/sorteros` changes any of these, so SorterOS machines behave the same way.

[...]

Issue #328, opened 2026-08-19
sorterOS images ship ext4 data=writeback, which silently corrupts SQLite on power loss


spencerhhubert:
## Summary

Machines in the field run their root filesystem with ext4's `journal_data_writeback`
data mode. Under that mode a power cut can leave **stale garbage inside files the
filesystem believes are intact**. SQLite's durability guarantees do not survive it,
and these machines lose power uncleanly on a regular basis.

This is not hypothetical: both SQLite databases on a deployed machine were found
corrupt during a routine disk-space investigation.

## Evidence

Root filesystem, as shipped:

```
/dev/mmcblkXp1 ext4 rw,noatime,errors=remount-ro,commit=600
Default mount options:  journal_data_writeback user_xattr acl
```

`journal_data_writeback` is set in the **filesystem superblock**, so it applies
regardless of what `/etc/fstab` says. It comes from the Orange Pi vendor base
image — there is nothing in this repo that sets it.

`PRAGMA integrity_check` on `local_state.sqlite`:

```
Tree 12     page 239535 cell 1: invalid page number 268675080
Tree 12     page 239535 cell 0: Rowid 3414 out of order
Tree 248723 page 1565932 cell 47: invalid page number 2681317773
Tree 248723 page 1565932 cell 50: Offset 11404 out of range 867..4092
Page 239624: never used
```

`local_metrics.sqlite` on the same machine was corrupt as well, and worse —
damage spread across many pages, including cross-linked ones:

```
Tree 5 page 31508 cell 321: 2nd reference to page 62410
Tree 5 page 31508 cell 317: invalid page number 2147512076
Tree 5 page 26953 cell 28:  Rowid 7260460 out of order
```

"Invalid page number", "offset out of range" and "2nd reference to page" are the
signature of blocks containing data that was never meant to be there — i.e.
unordered data writes, not failing media. Supporting that reading: the card
reported no I/O errors, `dumpe2fs` reported the filesystem clean, and there were
no ext4 errors in the kernel log. The wtmp record showed repeated unclean
shutdowns (`crash`, not `shutdown`).

## Impact

- Silent data corruption in `local_state.sqlite` — the DB holding piece records,
  bin state, sessions, and run history.
- Corruption is undetected until something happens to read the damaged page.
  Recovering the affected machine required a salvage-and-rebuild; it lost rows.
- **The corruption also disables the metrics retention policy** (see below), so
  the metrics DB grows without bound until the disk fills.

## Fix

[...]

PR #762, opened 2026-09-23
SorterOS 4.1: stable release on first boot, sorter.local, Wi-Fi setup that can't get stuck, models from Hive, tested before flashing

### LINKED ISSUE #583: remove model files from git lfs. repo clones hit bandwidth limits
[Spencer filed this from Discord](https://discord.com/channels/1430279849171222682/1546588142126235830)

really just need to check the things dependent on that. like install scripts.

---
_Opened from [a Discord post](https://discord.com/channels/1430279849171222682/1546588142126235830) by **Spencer** via the issue board. Replies in the Discord thread are mirrored here._

### LINKED ISSUE #675: Image customizer fails for everyone: "config too large: N bytes, capacity 2"

### LINKED ISSUE #677: SorterOS sets no hostname of its own, so sorter.local only works if you used the customizer
## Summary

Nothing in the SorterOS image sets a hostname. `sorter` comes only from the browser customizer, so a user who flashes the `.zip` directly gets a machine that answers to the base Orange Pi Ubuntu image's hostname instead, and every instruction that says `sorter.local` is wrong for them.

That is not a corner case at the moment: the customizer is broken for everyone (#675), so **the direct flash is currently the only working install path**, and it is the one where `sorter.local` never resolves.

## Detail

`sorter` is the initial value of the hostname field in the setup form:

```
software/sorteros/sorteros-setup/src/routes/+page.svelte
let hostname = $state('sorter');
```

It reaches the Pi only through `/etc/sorteros-config.toml`, and `sorteros-firstboot.py` applies it only if that file carries one:

```python
hostname = cfg.get("hostname")
if isinstance(hostname, str) and hostname.strip():
    log.info("setting hostname: %s", hostname)
    sh(["hostnamectl", "set-hostname", hostname])
```

`build.py`, `build/config.toml`, `chroot_apt.sh` and the overlay have no hostname handling at all, so with an unpatched config the base image's hostname stands. `backend/tests/test_server_security.py` treats that as `orangepi5`.

Related: `avahi-daemon` is not in the package list in `chroot_apt.sh`, so mDNS presumably comes from the base image. Worth confirming it is actually there, since everything user-facing depends on it.

## Suggested fix

Set the hostname to `sorter` in the image itself, as a build step or a firstboot default, and let the customizer's value override it. Then `sorter.local` is true for every install and the docs only have to say one thing.

## Reported by

Daddy-O's Bricks - Bill, 2026-09-17, after taking the #675 workaround: booted on Ethernet, `http://sorter.local:5173/` did not resolve, found the Pi through his router and reached the UI by IP. https://discord.com/channels/1430279849171222682/1550174361464606760/1550285942059696290

## Related

- #675 — the customizer bug that makes the direct flash the only route right now
- #676 — docs PR giving the correct address for each path, as an interim fix

I have not booted an unconfigured image to confirm the base hostname first-hand; that part is read from the code and the tests.

### LINKED ISSUE #705: WiFi settings: nmcli's terminal escape codes leak into the success message
The WiFi settings page shows a stray control character in front of nmcli's success message: `[2K Device 'wlan0' successfully activated ...`, with the `ESC` rendering as a box glyph in Chrome.

Seen on a real first setup by BrickCycleAlice, [screenshot in Discord](https://discord.com/channels/1430279849171222682/1550454875698237472/1550829376269062185), connecting an Orange Pi 5 to WiFi over the AP6275P module. The connection itself worked; this is cosmetic.

### Cause

`server/routers/wifi.py` hands `nmcli`'s stdout to the UI verbatim:

- `wifi_connect`, line 174: `return {"ok": True, "message": (proc.stdout or "").strip() or f"Connected to {ssid}"}`
- `wifi_disconnect`, line 195: the same shape
- the error paths at lines 172 and 193 pass `stderr`/`stdout` through the same way

`nmcli device wifi connect` draws a progress indicator while it associates and emits `ESC[2K` (erase line) before the final `Device 'wlan0' successfully activated with '<uuid>'.`. `_run` captures with `text=True` and no post-processing, so the escape survives into the JSON and then into the DOM. `.strip()` removes whitespace, not control sequences.

It is not visible over SSH because a terminal consumes the escape.

### Suggested fix

Strip C0 control characters and ANSI CSI sequences from `proc.stdout` / `proc.stderr` in `_run`, so every caller is covered rather than the two success paths only. Something like `re.sub(r"\x1b\[[0-9;?]*[ -/]*[@-~]|[\x00-\x08\x0b\x0c\x0e-\x1f]", "", s)` applied to both streams.

Passing `--colors no` to nmcli does not help here: `[2K` is cursor control, not colour.

### Not opened as a PR

`software/` is outside what I change on my own initiative. Filing this as a report rather than a change; happy to write the patch if someone wants it.

spencerhhubert:
Makes the documented SorterOS install path work end to end, removes the vision models from git, and adds a way to test an image before anyone flashes a card.

## SorterOS 4.1 (the image)

- **First boot checks out the stable release**, the newest `sorter/stable/v*` tag, the same channel the Versions page updates within, instead of a branch baked in at build time. (`main`'s build config had reverted to a stale `sorteros-v3` branch, so an image built from `main` would have broken at first boot.) `build.py --ref <branch>` bakes something else for a test image. The clone is blobless and made beside the target, so an interrupted clone can't look finished.
- **Every machine answers at `http://sorter.local`**: hostname `sorter`, avahi and nss-mdns installed (#677). A second machine becomes `sorter-2.local`.
- **The UI starts as soon as everything it needs is in place.** Tailscale installs and connects afterwards and gives up after 10 failures instead of holding port 80 forever. Any stage's error shows on the first-boot status page, which links to the install guide.
- **The root filesystem repairs itself.** ext4 is switched to `data=ordered` at build time (the vendor superblock ships `data=writeback`, which corrupted SQLite on power loss, #328), and `errors=panic` + `panic=10` + `fsck.repair=yes` make a filesystem error reboot and repair instead of leaving the machine read-only or waiting at an initramfs prompt (#639).
- **Getting online, decided at every boot** (`sorteros-network`, replacing the onboarding script, both AP scripts and `wifi-connect`): Ethernet if there's a cable; else the Wi-Fi given to the setup site (saved in NetworkManager once per network/password) or any saved network, given 90 s; else the `SorterOS-Setup-XXXXXX` network and its page, until the page joins a network, a cable appears, or a saved network comes back (retried every 3 minutes while nobody is on it, for a router slower to boot than the Pi). Nothing hangs off a flag written once. The old code marked Wi-Fi configured before checking it connected, which switched the setup network off for good and retried a wrong password every 60 s; and on Ethernet it could start the hotspot before DHCP landed and hold port 80 against the UI. While the setup network is up the Sorter UI steps aside so the page has port 80 (on any boot after the first, the page couldn't start at all).
- **The setup page**: a failed join brings the setup network back (joining had taken the radio); what the page collects is merged into the config instead of replacing it, so the setup site's Tailscale key is no longer dropped; hostname and SSH key from the page now apply (first boot re-applies the config whenever it changes); it uses whatever Wi-Fi device NetworkManager manages instead of assuming `wlan0` (a USB adapter is `wlx…`).
- **The setup site works on these images again**: the builder writes its placeholder back into `/etc/sorteros-config.toml`.
- **Overlay files are owned by root.** The builder's `rsync -a` copied the build user's uid onto `/etc`, `/etc/systemd/system` and the scripts root runs at boot; in the image that uid is the `orangepi` user's. Images already released were built the same way.
- **The first-boot service has a `HOME`.** Without one, `git config --global` fails under systemd; the old code hid that by skipping the rest of the clone stage on retry.
- **No git-lfs**, no LFS stage. Deleted: the frontend `.env` stage (nothing reads it), the overlay's copies of the four sorter units (first boot installs the repo's), the Mac-only build dashboard, Mac paths in the builder. The builder runs on any Linux box (arm64, or x86_64 with qemu-user-static) and verifies the base image's sha256.

## Test before flashing: `software/sorteros/test/`

- `check.py image` reads a built image and checks everything above (filesystem options, kernel args, hostname, enabled services, ownership, portal, no git-lfs).
- `boot.sh` boots the image in QEMU with a stock arm64 kernel, as a machine plugged into Ethernet with no hardware; `check.py boot` follows the first-boot status page stage by stage, then checks the backend, the checked-out ref, that Hive's default model landed on every channel, the hostname and avahi. On an Apple Silicon Mac it runs at native speed.
- `check.py wifi` runs `wifi-sim.sh` in the same VM: `mac80211_hwsim` radios for the Pi, a home router (hostapd + dnsmasq) and a phone (each in its own network namespace), and five scenarios against the real NetworkManager and setup page.
- `test_network.py` covers the same decisions in milliseconds against a fake NetworkManager and clock (17 cases).
- `sorteros-setup/scripts/patch_image.ts` patches an image with the site's own code, for testing a release.

These found the `HOME`, ownership and port-80 bugs above.

## Vision models come from Hive, not git

- `software/sorter/backend/bundled_models/` (34 files, ~119 MB) and every Git LFS rule and reference are gone, along with every code path for "bundled" models. A model directory put in the models folder by hand still works, as a **local** model.
- When a detection slot has no usable model (a new install, or an old `bundled:` id left in a machine's config), the backend asks Hive for the default detection model for its hardware (rknn on an Orange Pi 5, then ncnn/onnx), downloads it through the normal download queue, and assigns it to every such slot: C-Channel 2, C-Channel 3 and the classification C-channel. It never touches a slot with a working choice, and retries every 5 minutes when offline. The endpoint is public (#761), so this works before the machine is linked to Hive.
- Hive's default for rknn and onnx is **r5 C-Channel Full YOLO11s 320** (Ember).

## The setup site (#675)

setup.basically.website stopped at the marker text inside a compiled `.pyc`, about 1 GB before the real placeholder, and reported "capacity 2" for every config. It now matches the placeholder's comment lines and skips any region too small to be it. The test image reproduces the `.pyc` decoy; the old patcher fails on it, the new one passes. The builder also leaves compiled Python out of the image.

## Docs

The SorterOS install page describes the flow (optional setup site, then Ethernet, your Wi-Fi or the setup network; `sorter.local`; automatic model; updates via Versions), first-boot troubleshooting is new, the "choose a detection model" step is now "check the detection model", git-lfs is gone from the manual installs, and the SD card guidance says high-endurance.

## Tests

- Backend: the only failures are the 19 that fail on `main` too (one of the previous 20 was fixed here). `test_supervisor_crash_loop` flakes under full-suite load and passes alone. `test_wifi_output.py` is new (#705: nmcli's terminal codes no longer reach the UI).
- `test_network.py`: 17 passed.
- The image built from this branch passes `check.py image` (filesystem, kernel args, hostname, services, ownership, setup-site placeholder).
- QEMU on an M1 (HVF), first boot over Ethernet: the Sorter UI on port 80 about 2.5 minutes after the first-boot service started; Ember (rknn) from Hive on C-Channel 2, C-Channel 3 and the classification C-channel; hostname `sorter`, avahi running, no stage gave up.
- `check.py wifi`: 21 of 21. Setup-site Wi-Fi joined in 4 s; with a wrong password the setup network opened after 100 s and the phone fixed it (the Tailscale key survived); a wrong password typed on the phone brought the setup network back and the page said so; a router back after the Pi was rejoined 12 s after it came up; a cable closed the setup network.
- An image patched by `patch_image.ts` (hostname, Wi-Fi, SSH key, Tailscale key) booted with every one applied.

Closes #675, closes #677, closes #705, closes #583.


spencerhhubert:
Clean run on an image rebuilt from `b406f51b` (first boot pointed at this branch), booted in QEMU on an M1 with HVF, nothing patched by hand.

`check.py image`: all checks passed.

`check.py boot --expect-ref sorteros-setup-fixes --expect-default-model`:

```
  ssh-host-keys … write-machine-toml   done
  uv-sync              done     (15 s)
  pnpm-install         done
  pnpm-build           done     (30 s)
  install-services     done
  the Sorter UI has port 80
  ok   first boot hands port 80 to the Sorter UI
  ok   backend API answers
  ok   checked out sorteros-setup-fixes
  ok   a model was installed from Hive
  ok   C-Channel 2 uses a Hive model
  ok   C-Channel 3 uses a Hive model
  ok   Classification C-Channel (C4) uses a Hive model
  ok   hostname is sorter
  ok   avahi-daemon is running
  ok   no models committed in the checkout
  ok   git-lfs absent
  ok   no first-boot stage gave up
all checks passed
```

About 2.5 minutes from the first-boot service starting to the UI; Tailscale installed after the UI was up, then the first-boot service exited. The installed model is Ember's rknn build (`installed_as_default`, source hive.basically.website, sha256 matching Hive).

Not covered by QEMU, still needs a card: the WiFi setup network and portal, the NPU actually running the model, the kernel args taking effect through the Orange Pi boot loader, `sorter.local` from another computer.

basically-balloon:
**jtimmer312** approved this in Discord.

PR #764, opened 2026-09-23
docs(arranging c-channels): say which dovetail each camera lamp arm goes in

basically-balloon:
**Asked for by Daddy-O's Bricks - Bill** in `#balloon-does-docs`, https://discord.com/channels/1430279849171222682/1552428582029107221/1552432364314173461, with the three photos and the counts off his build.

**What changed**

- **Step 9 of Arranging C-channels now says which dovetail the arm goes in.** C2 and C3 take the 5th from the channel exit, the classification channel the 8th, counted anticlockwise seen from above with the first dovetail past the exit as number 1. Three of his photos go under it: C3 in its dovetail, the classification channel with its lamp on, and the seated close-up.
- **Step 9 now mentions the classification channel's lamp.** It said C2 and C3 only, and all three take one.
- **"A dovetail under the channel's NEMA bracket" is gone from four places**, replaced by the dovetails around the outside of the channel wall. The stator STL (`stator-323b2d1f94a5.stl`) carries 14 of them around the bottom of its outer wall, 25.71 degrees apart, undercut (about 11.5 mm across the mouth, 18 mm at depth, 20 mm tall); the discharge opening swallows two, leaving 12 to count. The C-channel arm mount's inner face is a 120.148 mm saddle against the stator's 120.0 mm outer wall, with a matching rail on it. The NEMA bracket has no feature that matches. Bill's photos show the mount down in a stator dovetail.
- **Both camera lamp pages carry the count in step 4**, and their top warnings no longer say the joint has never been photographed, because it now has been. The fit is still not dimensioned and they still say so.
- **Bill added to `contributors:` on the four pages**, and the classification channel page's warning now credits step 5 to his build.

- **The joint goes on from underneath.** Bill: it feeds into the dovetail from the bottom edge of the channel wall and clips over the top edge, so "slide it down" was wrong on all five pages. "Clips" over "latches" (no moving catch) and "clicks" (promises a sound nobody has confirmed, and the wrong cue on a part that snaps coming off). The STL agrees: the slot is open at the wall's bottom edge and closed at the top, and the mount's rail is about twice the slot's height.
- **A caution on all three pages**, asked for by ReveryX: the arm mount's dovetail breaks off easily when a lamp is pulled back out of the wrong slot, so count before pushing it on.

**Not done:** nobody has measured the dovetail clearance, so "do not force it and do not file it down" stays on both lamp pages.

`npm run check` clean, `validate_images.py` clean.

PR #760, opened 2026-09-23
docs(sorter): rework step 7 of first setup around the real camera cards

basically-balloon:
Step 7 was written from the wizard's source without a machine in front of it, and Bill has one. This rewrites it around what the cards actually say and do, and gives the zone editor a section of its own with a baseline to work from.

Asked for by @daddyosbricksbill in `#balloon-general`: https://discord.com/channels/1430279849171222682/1552315163607302256/1552386712561389669

**What changed**

- **Classification Top is now optional** in the camera table.
- The framing requirement is stated: a channel's drop zone **and** its exit have to be in frame, and if they are not, the camera lamp may need a different dovetail on its C-channel.
- **The two buttons are renamed to what the UI calls them.** `SetupCameraAreaCard.svelte` renders them as **Review zone** and **Picture tuning** (they flip to "Zone reviewed" and "Picture tuned" once done). The page said "Edit zone" and "Picture settings", which are the Settings page's labels, not the wizard's.
- **Review zone** comes first, matching both the wizard's workflow and the order the buttons sit in on the card, with Bill's walkthrough of the handles in place of the one-line placeholder, and five figures: the labelled zone editor, the shipped angles for C2, C3 and C4 as diagrams, and the three camera cards after setup with each zone labelled.
- **Picture tuning** follows it: color correction is not a manual setting, the camera-calibration link moved here from the end of the step (the "getting each camera sharp" line is gone), and the modes are listed — 1280x720 at 30 fps on C2 and C3, 3840x2160 at 30 fps on C4, and MJPG as the pixel format on all of them.
- The page's own AI-generated warning no longer claims step 7 has no screenshots and has not been checked against a machine. It has now.
- **Picture tuning ends with "Click Confirm"**, which is the primary button on that modal (`SetupPictureSettingsModal.svelte` passes `primaryActionLabel="Confirm"`, and `allowPrimaryActionWithoutChanges` means it can be clicked on a camera you did not touch). The zone editor has no Confirm, only **Save Zone**, so the zone walkthrough still ends there.
- A copy pass over the step at Bill's request: the camera-calibration sentence reads as one clause again, the handle walkthrough is punctuated as a list of things to set, and the line introducing the diagrams is no longer a fragment.

**One thing to check**

[...]

PR #767, opened 2026-09-23
docs(first setup): warn about the lamp arm dovetail before repositioning a lamp

basically-balloon:
**Asked for by Daddy-O's Bricks - Bill** in `#balloon-does-docs`, https://discord.com/channels/1430279849171222682/1552428582029107221/1552444826685935738: the same lamp dovetail warning he asked for in #764, next to the line in step 7 about moving a lamp when the drop zone and the exit are not both in frame.

**What changed**

One warning callout under that paragraph in step 7 of First setup: the dovetail on the arm can break when a lamp needs to be removed from it to be repositioned, so use caution if repositioning is needed. Reported by Bill and ReveryX from their own builds.

**Why this is a separate PR.** He asked for it on #760, which merged at 22:01 UTC while we were talking, so its branch is dead and this is the same change against `main`.

**Depends on #764** for one clause: the callout says step 9 of arranging C-channels is where the starting dovetail for each lamp is written down, which becomes true when #764 lands. The rest stands on its own.

`npm run check` clean.


basically-balloon:

basically-balloon:
**Daddy-O's Bricks - Bill** approved this in Discord.

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:
**Merged.** Approved in Discord by barthel, Daddy-O's Bricks - Bill, and it only changes files this repo allows to merge that way, so the board merged it.

PR #743, opened 2026-09-22
docs(hardware): state how tall the machine actually is

basically-balloon:
The Overview page told builders they choose how tall the machine is and then never said how tall that was. Nothing else on the site or in the catalog publishes a whole-machine height either.

**Added to "You choose how tall the machine is":** 1270 mm floor to top plate at 5 layers, 1772 mm to the top of the bulk bucket, 160 mm per layer either way, with the 3 layer figures worked out from that (950 mm and 1452 mm), plus a line that the bulk bucket is the highest point and is where the LEGO goes in.

**Every figure also carries a feet-and-inches conversion**, asked for on the board by Daddy-O's Bricks - Bill: 1270 mm (4 ft 2 in), 1772 mm (5 ft 10 in), 950 mm (3 ft 1 in), 1452 mm (4 ft 9 in), and 160 mm (6.3 in) per layer. The mm figure stays first and is the authoritative one; conversions are rounded to the nearest inch (1270 mm is exactly 50 in). Precedent for imperial in brackets on this site is the 1/8 inch cardboard line on Install the bins.

**Where the numbers come from:** measured on a finished 5 layer machine by BrickCycleAlice and reported in #balloon-does-docs with a photo of the tower, https://discord.com/channels/1430279849171222682/1536088194557419660/1551902838530113556.

The 160 mm per layer is the CAD length of piece C, the layer vertical (`ext-2020-c`, 160 mm CAD, cut to 154 mm), which is the gap between one hex ring and the next, so it is the layer pitch. The bulk bucket's 502 mm sits above the top plate and does not change with layer count.

`npm run check` clean.

PR #770, opened 2026-09-24
docs(hardware): what to do if the machine will live somewhere hot

basically-balloon:
Asked by [BrickCycleAlice](https://discord.com/channels/1430279849171222682/1552471841459671071/1552483758056411207), who is building in a tin shed in northern NSW coming into summer and asked whether a recommended ambient operating temperature range exists, then asked for the short version to live on a page for builders closer to the equator without air conditioning.

**There is no tested range**, and nothing on the site, in the catalog or in the meeting history sets one. So the new section says that first and then gives what the parts themselves limit, in the order they fail.

**New section on the Overview page, "If the machine will live somewhere hot"**, placed after the print figures because the actionable part of it is a filament decision:

- **PETG or ASA rather than PLA**, since PLA softens from around 55 C (131 F) and the parts at risk are the loaded ones bolted to a motor. PETG holds to around 80 C (176 F), and the page already says a full 5 layer set has been printed in PETG.
- **Cool the Orange Pi and keep detection on the NPU**, linking the existing [Orange Pi 5](https://docs.basically.website/hardware/orange-pi-5/) cooling section rather than repeating it. The 105 C (221 F) emergency shutdown figure is that page's.
- **Out of direct sun, free air around the supply.** The Mean Well LRS-350-24 is rated -25 to +70 C per its datasheet and the machine draws 30 to 90 W of its 350 W, so the supply is not the limit.
- **A working figure**: below about 40 C (104 F) in PLA, 45 to 50 C (113 to 122 F) in PETG with both Pi fans. That number is mine, worked from the limits above, not a spec anybody has tested, and the section is written so it reads that way.
- Condensation gets one clause, since a humid tropical climate is harder on the electronics than the heat is (the supply's spec is 20-90% RH non-condensing).

Every temperature carries a Fahrenheit conversion in brackets, as she asked, following the feet-and-inches convention #743 set on this same page.

`npm run check` clean, `validate_images.py` clean, rendered and read on the dev server.


basically-balloon:

basically-balloon:

basically-balloon:

basically-balloon:

That is 2 of 2 approvals.

basically-balloon:

